PR #9, opened 2022-09-04
Add Nix development environment for the Nix users out there

cody-quinn:
This PR adds a Nix development environment that will allow Nix users to easily build and work on valence without having to write their own. 

If using flakes you can get into the development shell using `nix develop`, otherwise you can use `nix-shell`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,94 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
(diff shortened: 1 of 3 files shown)


CatDevz:
If you could wait before merging, just realized the rustfmt config requires the unstable tool chain

CatDevz:
Should be good to merge at this point.

PR #14, opened 2022-09-04
Reduce boilerplate by making a default client event.

LeoDog896:
This adds an (encouraged) default client event from the examples to reduce confusion for examples.


rj00a (approved):
I had a reason for not doing this but I forgor 💀

PR #15, opened 2022-09-04
Add action bar support

LeoDog896:
This adds an action bar packet + function, as well as integration into conway.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -34,7 +34,7 @@ use crate::protocol::packets::s2c::play::{
     EntityStatus, EntityTrackerUpdate, EntityVelocityUpdate, GameJoin, GameMessage,
     GameStateChange, GameStateChangeReason, KeepAlive, MoveRelative, PlayerActionResponse,
     PlayerPositionLook, PlayerPositionLookFlags, PlayerRespawn, PlayerSpawnPosition, RegistryCodec,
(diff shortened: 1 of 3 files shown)


src/client.rs
```diff
@@ -485,6 +485,11 @@ impl<C: Config> Client<C> {
         }
     }
 
+    /// Sets the action bar for this client.
+    pub fn set_action_bar(&mut self, text: impl Into<Text>) {
+        self.send_packet(SendActionBar { text: text.into() });
```
rj00a, line 490: Might be better to defer sending this packet.

The reason is that the login/join game packet is also deferred, and a lot of packets don't work correctly if they are sent before the login packet.

The other reason is that if set_action_bar is called multiple times in a single tick, you'll only send one packet. But then you have to add a field to `Client`.

But whatever this works fine. Just something to think about.

src/protocol/packets/s2c.rs
```diff
@@ -629,6 +629,12 @@ pub mod play {
         }
     }
 
+    def_struct! {
+        SendActionBar {
```
rj00a, line 633: Should call this `OverlayMessage` for consistency with packets.json. I know it's dumb.

rj00a (changes requested):
I think there's a merge conflict now.

src/protocol/packets/s2c.rs
LeoDog896: It's not, though possibly for the def_struct macro for packets, I think using the names from JSON as a type hint / error would be good

Issue #16, opened 2022-09-04
GitHub workflow with more checks


rj00a:
I set up github workflows to run `cargo build` and `cargo test` on push to `main`.

Here are some other things it should check for:
- `cargo clippy` warnings
- `cargo fmt` changes
- broken doc links

CatDevz:
I'll work on this.

Issue #18, opened 2022-09-04
The `Text` struct is too large.


rj00a:
`cargo clippy` often complains that `std::mem::size_of::<Text>()` is too large. There are a few things that could be done.
- Stop using `Cow<'static, str>` everywhere and just use `String`. We can reintroduce it later if it turns out to be important.
- Get rid of the `Option<bool>`s by manually implementing `Serialize` and `Deserialize`. Use a bitfield struct with two bits per `Option<bool>`.

dyc3:
I'd like to take a stab at this.

PR #20, opened 2022-09-04
Add clippy linting and formatting validation to GitHub actions workflow

CatDevz:
Fixes issue #16 by failing the workflow if clippy lints dont pass or formatting is incorrect.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -24,3 +24,7 @@ jobs:
       run: cargo build --verbose
     - name: Run tests on ${{ matrix.os }}
       run: cargo test --verbose
+    - name: Run clippy lints
+      run: cargo clippy --verbose --no-deps
+    - name: Validate formatting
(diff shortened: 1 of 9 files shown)


rj00a:
I would like to avoid boxing the `Text` to fix the clippy issue. Fixing #18 should resolve that.

rj00a (changes requested):
I would like to avoid boxing the `Text` to fix the clippy issue. Fixing #18 should resolve that. In the meantime you can tell clippy to ignore it.

Also I don't believe `cargo clippy` will tell you about broken docs so this doesn't fully resolve #16.

dyc3:
Adding another step for `cargo test --doc` should give us signal about broken docs. It needs to be another step because the `--doc` arg makes cargo test _only_ the docs, and it skips the normal unit tests.

rj00a:
Closing in favor of #46

PR #25, opened 2022-09-05
Add support for Slot data decoding

guac420:
Adds the `Seek` trait to the `reader` in the `Decode` trait. This allows seeking functionality when decoding packets, which vanilla Minecraft utilizes. This functionality is necessary for decoding `Slot` data from packets. The PR also adds a basic `Slot` struct.

### DIFF
--- a/src/protocol/slot.rs
+++ b/src/protocol/slot.rs
@@ -1,4 +1,5 @@
-use std::io::{Read, Seek, SeekFrom, Write};
+use byteorder::ReadBytesExt;
+use std::io::Write;
 
 use crate::nbt::Compound;
 use crate::protocol::{Decode, Encode, VarInt};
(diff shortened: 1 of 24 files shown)


rj00a (changes requested):
Looks like there are some build errors. You can wrap a `&[u8]` in a `std::io::Cursor` to have it implement `Seek`.

rj00a (changes requested):
I realized `&mut &[u8]` would be a better interface than `Read + Seek` for a few reasons.
- You can't rewind/go backwards.
- Fewer fallible operations involved.

`&mut &[u8]` Actually implements `Read` so not much code would have to change.
---
`slot.rs` should be in the `protocol` module.

Sekky61:
I would opt for different type signature for `Slot`.
```rust
pub enum Slot {
    Empty,
    Full {
        item_id: VarInt,
        item_count: u8,
        nbt: Option<Compound>,
    },
}
```
Unless I am missing something, this seems more elegant and removes illegal states that would have to be checked.

guac420:
I took @Sekky61's advice and updated the type signature for `Slot`. I also looked into using `&mut &[u8]` instead of `Read + Seek` but found that `Slot::decode` needs the seek functionality in order to prevent parsing invalid nbt data and to prevent just assuming that an error in the `Compound::decode` function means there is no nbt data [like this](https://github.com/Jachdich/valence/commit/efae9d1ae90257fd1c835fd7e4c7e1591aae8b1b#diff-60a37e7c13d3cc380c03021332b2eb09af9f143c94e56b5cab27b57b15ff79ceR157-R168). The vanilla implementation of `Slot::decode` uses `Seek` functionality to determine if there is nbt data before trying to parse it.

src/protocol/slot.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// Gets the item's nbt data.
+    ///
+    /// If the slot is empty or there is no nbt data, then `None` is returned.
+    pub fn nbt(&self) -> Option<&Compound> {
+        match self {
+            Slot::Empty => None,
+            Slot::Present { nbt, .. } => nbt.as_ref(),
+        }
+    }
+}
```
rj00a, line 70: If the `Slot` type is private to the protocol module it's fine to leave all this getter/setter stuff off and just make the fields public. (In fact, it already is because it's an `enum`). If we need these we can add them later.

src/protocol/slot.rs
```diff
@@ -0,0 +1,186 @@
...
+        Ok(Slot::new(
+            VarInt::decode(r)?,
+            u8::decode(r)?,
+            match u8::decode(r)? {
+                0 => None,
+                _ => {
+                    r.seek(SeekFrom::Current(-1))?;
+                    Some(Compound::decode(r)?)
+                }
+            },
```
rj00a, line 108: We can do this better with `&mut &[u8]`. Consider:

```rust
let some_bytes = &mut &[1, 2, 3, 4, 5, 6, 7];

// Is there a zero byte at index zero?
if some_bytes.get(0) == Some(&0) {
    some_bytes.read_u8()?;
    // No NBT data.
} else {
    // There is NBT data.
}
```

rj00a (changes requested):

src/protocol/slot.rs
guac420: That makes sense, I'll remove these.

src/protocol/slot.rs
guac420: Oh, I see now. That is a lot simpler than the `Read + Seek`. Should the `Encode` trait signature also be updated to use `&mut &[u8]`?

rj00a: The analog for `Write` is actually `&mut [u8]` but you can leave it.

PR #27, opened 2022-09-06
Fix order of chunk creation in conway's example

SirJosh3917:
This would typically go unnoticed, but upon expanding the and making meaningful use of both directions, this being an issue arises.


rj00a:
Nice catch

PR #28, opened 2022-09-06
Create parkour.rs example

emortaldev:


emortaldev:
I forgot to remove the separate world when the player leaves

rj00a:
> I forgot to remove the separate world when the player leaves

You can still push changes. Don't forget to `cargo fmt`.

emortaldev:
I believe this is ready, worlds are now unloaded when players leave

examples/parkour.rs
```diff
@@ -0,0 +1,380 @@
...
+
+struct ServerState {
+    player_list: Option<PlayerListId>,
+}
+
+#[derive(Clone)]
+struct BlockAndTick {
+    pos: BlockPos,
+    //tick: i64
+}
```
rj00a, line 50: This could just be a `BlockPos`

rj00a (changes requested):
If you rebase with `main` you'll run into some issues.

Also `cargo clippy --example parkour`.

rj00a (approved):

PR #29, opened 2022-09-06
prevent typing in chat from kicking players

### LINKED ISSUE #24: Typing in chat causes client to disconnect
Using commit `cac348a4b6e788f3ecce24c7559c1789f4c8b402`, running example `combat`.

## Server log
```
[2022-09-05T13:03:17Z TRACE valence::server] entering accept loop
[2022-09-05T13:04:22Z ERROR valence::server] connection to 127.0.0.1:64121 ended: error during play: packet contents were not read completely
```

## How to reproduce
- Join the server
- Type a message
- Get disconnected

This bug occurs in all examples and non-examples.

dyc3:
Simply move the assertion for decompressed packet contents, because `P::decode_packet` consumes the packet it is given. It might be a good idea to consider just passing ownership of `decompressed` rather than `&mut decompressed` to make this behavior more clear. Not allowing `P::decode_packet` to modify the packet would also be a good solution.

fixes #24

# Test plan

run
```
cargo r -r --example conway
```

- connect to server on localhost
- type in chat
- see that you don't get kicked
- the chat message does not appear in chat (out of scope for this PR)

### DIFF
--- a/src/protocol/codec.rs
+++ b/src/protocol/codec.rs
@@ -196,12 +196,12 @@ impl<R: AsyncRead + Unpin> Decoder<R> {
                     .context("decompressing packet body")?;
 
                 let mut decompressed = self.decompress_buf.as_slice();
-                let packet = P::decode_packet(&mut decompressed)
-                    .context("decoding packet after decompressing")?;
                 ensure!(
                     decompressed.is_empty(),
                     "packet contents were not read completely"
                 );
+                let packet = P::decode_packet(&mut decompressed)
+                    .context("decoding packet after decompressing")?;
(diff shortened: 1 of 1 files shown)


XBagon (commented):
I'm assuming you tried this out? I just tried it with the `terrain` example and get kicked, because of ` packet contents were not read completely` immediately.

dyc3:
Yes, I tried it with `conway` but not with `terrain`.

dyc3:
I misunderstood the purpose of this assertion, whoops. I'll do some more digging and update #24, closing this for now.

PR #30, opened 2022-09-07
Update definition for ChatMessage and CommandExecution

### LINKED ISSUE #24: Typing in chat causes client to disconnect

dyc3:
fixes #24

I went ahead and updated `CommandExecution` too since these 2 packets are similar.

# Test plan

1. run `cargo r -r --example terrain`
2. run mc and join localhost
3. type any non command in chat
4. type `/help` in chat
5. see that you don't get kicked

### DIFF
--- a/src/protocol/codec.rs
+++ b/src/protocol/codec.rs
@@ -200,7 +200,10 @@ impl<R: AsyncRead + Unpin> Decoder<R> {
                     .context("decoding packet after decompressing")?;
                 ensure!(
                     decompressed.is_empty(),
-                    "packet contents were not read completely"
+                    format!(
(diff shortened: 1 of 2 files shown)


rj00a:
Nice work

Issue #19, opened 2022-09-04
Some text content types are not yet implemented.


rj00a:
https://minecraft.fandom.com/wiki/Raw_JSON_text_format

In the `text` module, plain text is implemented but the other content types are missing/need work.
- translated text
- scoreboard values
- entity names
- keybind
- NBT values

gabrielztk:
Can I try my hand at this issue?

dyc3:
Absolutely! Looking forward to your PR

PR #31, opened 2022-09-07
Decrease size of Text struct

### LINKED ISSUE #18: The `Text` struct is too large.

dyc3:
This implements the changes outlined in #18. However, I'm not really sure merging this in is a good idea because of the amount of code required to implement `Serialize` and `Deserialize` manually. I honestly don't think that its worth the effort, especially seeing that the `TextContent` enum is incomplete (see #19), and changes to that affects the `Serialize`/`Deserialize` impls.

closes #18

### Test plan

```
cargo test
```

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -5,6 +5,7 @@ use std::fmt;
 use std::io::{Read, Write};
 
 use serde::de::Visitor;
+use serde::ser::SerializeMap;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::ident::Ident;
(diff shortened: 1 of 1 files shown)


rj00a:
In retrospect my suggestions probably weren't so great. Sorry about that. I'll just tell clippy to shut up.

dyc3:
No worries, it was a good exercise for me.

PR #32, opened 2022-09-07
Add block entities

### LINKED ISSUE #5: Block entities
Support for [block entities](https://minecraft.fandom.com/wiki/Block_entity) should be added to the chunk API.

It looks like there are only a few things that need explicit support from Valence.
* The text on signs
* Banner patterns
* The entity inside of spawners
* Player data in player heads
* Maybe some other blocks?

My current idea is to put a `HashMap<u32, BlockEntity>` in each chunk. The key is the block index and the value is an `enum` of the possible block states.

MrlnHi:
This PR aims to add block entities.
Fixes #5


build/block.rs
```diff
@@ -522,9 +539,111 @@ pub fn build() -> anyhow::Result<TokenStream> {
...
+
+        #[derive(Clone, PartialEq, Debug)]
+        pub struct SignData {
+            pub glowing_text: bool,
+            pub color: TextColor,
+            pub texts: [Text; 4],
+        }
+
+        #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+        pub enum TextColor {
```
dyc3, line 604: We already have a `Color` type in `text.rs` -- would it be possible to reuse that here?

examples/conway.rs
```diff
@@ -296,6 +296,23 @@ impl Config for Game {
                         }
                     }
                 }
+
+                chunk.set_block_state(0, (51 - min_y) as usize, 0, BlockState::OAK_SIGN);
+                chunk.set_block_entity(
```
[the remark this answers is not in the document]
Marlonzockt: Yes, this is currently just for debugging purposes. I don't actually want to have a sign on every chunk of the conway example lol

build/block.rs
Marlonzockt: While they might have the same values, in vanilla minecraft they're from two different sources (doesn't mean we can't use it obviously, just one thing to consider)

Marlonzockt: Also - the `Color` type in `text.rs` consists of rgb values with constants for each color. This is not the behavior we want here, as signs only have that enum. Thus I will keep my enum

Marlonzockt:
I have decided to add all block entities that affect the client's visuals. This is the full list of block entities in vanilla minecraft.
Those ticked off will be implemented, those ticked off but enclosed in parentheses should be implemented but can't currently (for example lectern due to item api)

Feel free to correct me if I missed any

- [ ] Furnace
- [x] Chest
- [x] Trapped Chest
- [x] Ender Chest
- [ ] Jukebox
- [ ] Dispenser
- [ ] Dropper
- [x] Sign
- [x] Mob Spawner
- [x] Piston
- [x] (Brewing Stand)
- [x] Enchanting Table
- [ ] End Portal
- [ ] Beacon
- [ ] Skull
- [ ] Daylight Detector
- [ ] Hopper
- [ ] Comparator
- [x] Banner
- [ ] Structure Block
- [ ] End Gateway
- [ ] Command Block
- [ ] Shulker Box
- [ ] Bed
- [ ] Conduit
- [ ] Barrel
- [ ] Smoker
- [ ] Blast Furnace
- [x] (Lectern)
- [x] Bell
- [ ] Jigsaw
- [x] (Campfire)
- [ ] Beehive
- [ ] Sculk Sensor
- [ ] Sculk Catalyst
- [ ] Sculk Shrieker

rj00a:
Btw you can use `valence::prelude::*` in the example.

rj00a (changes requested):
- Doesn't look like we need both `BlockEntity` and `BlockEntityData` types. We should be able to determine the kind of the block entity from the discriminant of the enum. We don't want to expose `VarInt`s directly like that anyway.
- Having to manually implement every block entity type by hand is not ideal. It's not great because   we would need to remember to add more types as Minecraft is updated. It's also easy to get wrong and forget stuff.

  There's got to be a better way. Can we extract this data? How does Minestom handle the problem?
- We need a way to mark block entities as dirty. By this I mean we need to know when to send the  [Block Entity Data](https://wiki.vg/Protocol#Block_Entity_Data) packet when the block entity is modified. Usually this is done by setting a flag when the entity is modified and clearing all the flags at the end of the tick. With this consideration, the `block_entity_mut` function becomes problematic for reasons that I will avoid explaining because what I'm about to suggest will make this impossible to implement anyway.
- We need to take a step back and rethink the way block entity data is stored in chunks. Consider the case of `BlockEntityData::Sign`. When the block entity data needs to be sent to a client, here are the steps taken:
  1. For each of the four `Text` objects in the sign, a string buffer is allocated and serialization to JSON is performed.
  2. A new `Compound` and all of its fields are allocated to hold the various strings and other data.
  3. The `Compound` is encoded to NBT.
  4. The `Compound` is thrown out.
  
  This kind of work is happening to _every_ block entity for _every_ client on _every_ chunk load. Lots of duplicated work. Not only that, but the block entities are using a lot of memory too. I'm sure `std::mem::size_of::<SignData>()` is quite large, not counting the data it holds on the heap.

Putting this all together, here is what I'm proposing:

```rust
struct LoadedChunk {
    // ...
    block_entities: BTreeMap<u32, BlockEntityData>,
}

struct BlockEntityData {
    /// `true` when this block entity is dirty.
    modified: bool,
    kind: BlockEntityKind,
    /// Contains serialized NBT data for the block entity.
    data: Vec<u8>,
}

enum BlockEntity {
    Sign(SignData),
    Chest,
    Piston(PistonData),
    // ...
}
```

Then the methods on `Chunk` could look like:
```rust
fn block_entity(&self, x: usize, y: usize, z: usize) -> Option<BlockEntity>;

fn set_block_entity(&mut self, x: usize, y: usize, z: usize, block: Option<&BlockEntity>);

#[must_use = "If the old block entity is not needed, use `set_block_entity` instead."]
fn replace_block_entity(&mut self, x: usize, y: usize, z: usize, block: Option<&BlockEntity>) -> Option<BlockEntity>;
```

`block_entity` and `replace_block_entity` decode NBT from `data` while `set_block_entity` and `replace_block_entity` encode NBT to `data`.

My rationale here is that we're going to be sending block entities a lot more than we're reading their contents. How often in code do you need to read the text on a sign or the content of a spawner? Probably not as often as you need to send that data to clients.

# Problems I can think of right now
- `valence_nbt` does not provide any way to encode directly to a buffer. You have to make a `Compound` first. `serde` would let you do this, but `serde` is not a good fit for NBT.
- There is still some extra cloning that's necessary, but that will be fixed in #83

MrlnHi:
Thanks so much for taking the time to write that all down :)
just making sure: `BlockEntityData.kind` should be of type `BlockEntity` instead of `BlockEntityKind`, right?

rj00a:
> `BlockEntityData.kind` should be of type `BlockEntity` instead of `BlockEntityKind`, right?

No, `kind` would just be a fieldless enum/integer so we don't have to store both the encoded data and the decoded data at the same time.

MrlnHi:
another thing: Minestom doesn't handle block entities itself. The user can register a `BlockHandler` (iirc) and then decide what block entity data is sent to the client

rj00a:
If extracting the block entity stuff ends up being too much of a pain we could just throw in the towel and use a `Compound` as a compromise between read/write performance.

rj00a (commented):
Currently working on a fix for the manual packet implementation issue.

rj00a:
The block entity example is disconnecting the client.

rj00a:
Actually we should only do the map lookup when the old block state was a block entity. Too tired to fix that rn.

extracted/entities.json
```diff
@@ -1281,7 +1281,7 @@
         "name": "health",
         "index": 9,
         "type": "float",
-        "default_value": 20.0,
+        "default_value": 80.0,
```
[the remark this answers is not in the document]
MrlnHi: idk tbh. just started the generator and copied all over. probably only should've copied the blocks one over

MrlnHi:
> The block entity example is disconnecting the client.

Fixed

MrlnHi:
> Currently working on a fix for the manual packet implementation issue.

Yup, so I'm guessing we're gonna wait for that before merging?

src/chunk.rs
```diff
@@ -431,6 +456,30 @@ impl Chunk for UnloadedChunk {
...
+        assert!(
+            x < 16 && y < self.height() && z < 16,
+            "chunk block offsets must be within bounds"
+        );
+
+        let idx = (x + z * 16 + y * 16 * 16) as _;
+
+        self.block_entities
+            .get(&idx)
+            .map(|block| block.parse().unwrap())
```
MrlnHi, line 469: note to self: add meaningful message to unwrap

MrlnHi:
Ready for review

rj00a (commented):
- Yeah, the idea of using a `dirty` boolean in each block entity was not great. I think it would be better to have something like `modified_block_entities: BTreeSet<u32>` to track this information.
- Because `cached_init_packets` is a thing now, we don't need to store raw nbt data in each block entity. Better to just get rid of `BlockEntityData` and use `BlockEntity`. The `block_entity` getter could probably just return an `Option<&BlockEntity>`
- You should add a test or two in `chunk.rs` for block entities if you can.
- Just brainstorming, but it might be beneficial to unify `BlockState` and `BlockEntity` behind a common trait somehow? My current idea:
  ```rust
  pub trait Block: Clone + PartialEq + Debug {
      fn state(&self) -> BlockState;
      fn into_block_entity(self) -> Option<BlockEntity>;
  }
  ```
  I think this would be better for higher level methods on `Instance` and not `Chunk`. So probably not relevant for this PR.

crates/valence_protocol/build/block.rs
```diff
@@ -786,5 +856,25 @@ pub fn build() -> anyhow::Result<TokenStream> {
                 Self::from_bool(b)
             }
         }
+
+        #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+        pub enum BlockEntityKind {
+            #block_entity_type_variants
+        }
+
+        impl BlockEntityKind {
```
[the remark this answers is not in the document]
MrlnHi: Multiple different blocks can use the same Block Entity (for example signs with oak_sign, birch_sign etc all using block entity sign). Yea, we could make it return an iterator or something like that of all blocks that support the block entity, but I don't really see a use case here

crates/valence/examples/block_entities.rs
```diff
@@ -0,0 +1,147 @@
...
+    for UseItemOnBlock {
+        client, position, ..
+    } in block_interacts.iter()
+    {
+        if *position == SKULL_POS {
+            let Ok(client) = clients.get(*client) else {
+                continue
+            };
+
+            let uuid: [i32; 4] = unsafe { std::mem::transmute(client.uuid().as_u128()) };
```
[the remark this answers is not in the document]
MrlnHi: In this case maybe having a `From<Uuid> for Value` would be neat, right?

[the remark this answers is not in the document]
rj00a: Yeah that would be good. The impl would have to be in `valence_nbt` behind a `uuid` feature.

crates/valence_protocol/build/block.rs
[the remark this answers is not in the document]
rj00a: Ah right. That makes my idea about the `Block` trait not really work.

crates/valence_protocol/src/text.rs
```diff
@@ -829,6 +830,15 @@ impl<'a> From<&'a Text> for Cow<'a, Text> {
     }
 }
 
+impl From<Text> for Value {
+    fn from(value: Text) -> Self {
+        Value::String(
+            serde_json::to_string(&value)
+                .unwrap_or_else(|err| panic!("failed to jsonify text {value:?}\n{err}")),
```
rj00a, line 837: Hmmm I forgot that Text -> Json is a fallible operation... this isn't so great then. Maybe `Text` could be redesigned in a way to avoid this problem.

rj00a: It might not actually be possible for `Text` to panic when writing to a `String` buffer so it's fine I guess.

MrlnHi: It really shouldn't be able to happen I think

dyc3 (changes requested):
As a part of this PR, I would like to see the `chest` example updated so the chest is no longer invisible when you first join.

MrlnHi:
> As a part of this PR, I would like to see the `chest` example updated so the chest is no longer invisible when you first join.

that is actually working already as now the `set_block_state` function also places the corresponding block entity with no data at the same position :smiley:

crates/valence_protocol/build/block.rs
```diff
@@ -373,6 +398,69 @@ pub fn build() -> anyhow::Result<TokenStream> {
...
+        .collect::<TokenStream>();
+
+    let block_entity_kind_to_ident_arms = block_entity_types
+        .iter()
+        .map(|block_entity| {
+            let name = ident(block_entity.name.to_pascal_case());
+            let ident = &block_entity.ident;
+
+            quote! {
+                Self::#name => Ident::new(#ident).unwrap(),
```
MrlnHi, line 459: @rj00a idea I'd like to have your opinion on:
- We create an unsafe function `Ident::new_unchecked` that takes in both the string and the index (second field) and call that here
Because right now everytime you call `BlockEntityKind::ident` it needs to parse the entire string.
Instead we parse it here during compile time. Not sure if it's worth it, just an idea I had

rj00a (commented):
I've been mulling over this today and I think there are some changes we should make. Let me know what you think of my idea.

I would like to see the `(set_)block_state` and `(set_)block_entity` functions unified to make things simpler. Here is what `set_block` could look like.

```rust
fn set_block(&mut self, x: usize, y: usize, z: usize, block: impl Into<Block>) -> Block { ... }
```

Then `Block` looks like

```rust
pub struct Block {
    state: BlockState,
    entity: Option<BlockEntity>,
}
```

`Block` has the invariant that `entity` is correct for the current `state`. For instance, if `state` was `BlockState::AIR` then `entity` _must_ be `None`, because air is not a block entity.

`Block` would have a few different methods. We could add some convenient constructors to `Block` for making certain block entities.

There would also be `From<BlockState> for Block` so you could call `set_block` with a block state for convenience.

Things get a bit trickier when considering what `block` and `block_mut` functions would look like because we can't just return `&Block` and `&mut Block` respectively. I suggest we make `BlockRef` and `BlockMut` types.

```rust
struct BlockRef<'a> {
    state: BlockState,
    entity: Option<&'a BlockEntity>,
}

struct BlockMut<'a> {
    state: BlockState,
    /// Entry into the block entity map.
    entry: Entry<'a, u32, BlockEntity>,
    modified: &'a mut BTreeSet<u32>,
}
```

MrlnHi:
I think your idea is good, just I think we shouldn't remove the existing `set_block_state` and `set_block_entity` functions (if anything remove the block entity placement part from the `set_block_state` one), but just have the block ones additionally. It wasn't clear from what you said if you wanted to scrap the other one's or do what I just proposed. What do you think?

rj00a:
We could keep the old functions for the rare performance sensitive situations, but I think we should highly discourage their use and point towards `set_block`/`block` as the alternative. I also think we should remove `set_block_state` and `set_block_entity` functions from `Instance` and only make `set_block` available there. And as you said, the block entity placement behavior in `set_block_state` should be removed.

crates/valence_protocol/src/ident.rs
```diff
@@ -40,6 +40,17 @@ pub struct Ident<S> {
     path_start: usize,
 }
 
+impl<S> Ident<S> {
+    /// Returns an Ident with the given fields
+    ///
+    /// # Safety
+    /// This function does not check for the validity of the Ident.
+    /// For a safe version use [`Ident::new`]
+    pub const unsafe fn new_unchecked(string: S, path_start: usize) -> Self {
```
rj00a, line 49: I don't think this function should be marked as unsafe because unsafe code cannot rely on the invariants of `Ident` to be upheld to begin with. By using an evil implementation of `AsRef` or `Borrow` combined with `as_str_ident`, you can create invalid `Ident`s within safe code.

That said, this function really should not be used under normal circumstances. It would be good to add `#[doc(hidden)]`.

crates/valence_protocol_macros/src/ident_str.rs
```diff
@@ -36,6 +36,6 @@ pub fn ident_str(item: TokenStream) -> Result<TokenStream> {
     };
 
     Ok(quote! {
-        unsafe{Ident::new_unchecked(#ident, #path_start)}
+        Ident::new_unchecked(#ident, #path_start)
```
rj00a, line 39: `Ident` needs to be fully qualified because it might not be in scope where the macro is used.

```rust
::valence_protocol::ident::Ident::new_unchecked(#ident, #path_start)
```

JoelEllis:
> I would like to see the (set_)block_state and (set_)block_entity functions unified to make things simpler. Here is what set_block could look like.
> 
> ```rust
> fn set_block(&mut self, x: usize, y: usize, z: usize, block: impl Into<Block>) -> Block { ... }
> ```

Whilst you're thinking about this, I think the current API has a bit of a foot-gun here,

> If the position is not within a loaded chunk or otherwise out of bounds, then `BlockState::AIR` is returned with no effect.

instead it would be better to return a result, rather than effectively returning 0. :smile:

MrlnHi:
> > I would like to see the (set_)block_state and (set_)block_entity functions unified to make things simpler. Here is what set_block could look like.
> > 
> > ```rust
> > fn set_block(&mut self, x: usize, y: usize, z: usize, block: impl Into<Block>) -> Block { ... }
> > ```
> 
> Whilst you're thinking about this, I think the current API has a bit of a foot-gun here,
> 
> > If the position is not within a loaded chunk or otherwise out of bounds, then `BlockState::AIR` is returned with no effect.
> 
> instead it would be better to return a result, rather than effectively returning 0. :smile: 

Have a look at the PR. It now returns an `Option` :smile:

crates/valence/src/instance/chunk.rs
```diff
@@ -46,6 +53,97 @@ struct Section {
...
+/// Mutable reference to a block in a chunk
+#[derive(Debug)]
+pub struct BlockMut<'a> {
+    state: BlockState,
+    /// Entry into the block entity map.
+    entry: Entry<'a, u32, BlockEntity>,
+    modified: &'a mut BTreeSet<u32>,
+}
+
+impl<'a> BlockMut<'a> {
```
rj00a, line 124: There should be a `fn nbt(&self) -> Option<&Compound>` for completeness.

crates/valence/src/instance/chunk.rs
```diff
@@ -46,6 +53,97 @@ struct Section {
...
+        match &mut self.entry {
+            Entry::Vacant(_) => None,
+            Entry::Occupied(oe) => {
+                self.modified.insert(*oe.key());
+                Some(&mut oe.get_mut().nbt)
+            }
+        }
+    }
+
+    pub fn set_nbt(&mut self, nbt: Compound) {
```
rj00a, line 139: This can return the previous NBT compound?

crates/valence/src/instance/chunk.rs
```diff
@@ -46,6 +53,97 @@ struct Section {
...
+        match &mut self.entry {
+            Entry::Vacant(_) => todo!(),
```
rj00a, line 141: Unfinished.

crates/valence_nbt/src/compound.rs
```diff
@@ -41,11 +41,11 @@ impl fmt::Debug for Compound {
 }
 
 impl Compound {
```
rj00a, line 43: `new` and `with_capacity` cannot be `const` because `Map::new` and `Map::with_capacity` are not always `const`.

rj00a (commented):
I fixed CI. There are a few errors.

crates/valence/src/instance/chunk.rs
```diff
@@ -46,6 +53,97 @@ struct Section {
...
+            },
+        }
+    }
+
+    /// # Panics
+    /// This function panics if the specified [`BlockState`] does not support a
+    /// block entity yet `nbt` is [`Option::Some`]
+    pub const fn with_nbt(state: BlockState, nbt: Option<Compound>) -> Self {
+        if state.block_entity_kind().is_none() {
+            assert!(nbt.is_none(), "blockstate does not support nbt");
```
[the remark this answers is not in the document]
MrlnHi: Just doing that silently isn't that good of an idea I feel like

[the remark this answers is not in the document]
rj00a: I would normally agree with you in situations like these, but considering that we have no way to verify the actual contents of the compound are correct, panicking feels inconsistent. In other words, we're already silently ignoring errors by accepting nonsensical compounds, so setting `nbt` to `None` silently is consistent with this behavior.

In any case, if we must fail loudly it should probably return `Option<Self>` instead of panicking. But I would prefer to just set `nbt` to `None`.

Side note: Should this function accept a `Compound` instead of an `Option<Compound>`? Passing in `None` is equivalent to calling `new` so you might as well call `new` instead.

crates/valence/src/instance/chunk.rs
rj00a: Actually, I'm not sure this function should exist at all. If you need to mutate the NBT, you can use `nbt_mut` instead and avoid the problems associated with `Block::with_nbt`.

crates/valence/src/instance/chunk.rs
```diff
@@ -46,6 +53,97 @@ struct Section {
...
 
+/// Represents a block with an optional block entity
+#[derive(Clone, Debug)]
+pub struct Block {
+    state: BlockState,
+    /// Nbt of the block entity
+    nbt: Option<Compound>,
+}
+
+impl Block {
```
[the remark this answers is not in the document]
JadedBlueEyes: Perhaps just make Block copy?

PR #34, opened 2022-09-07
Added SetExperience and SetHealth packets

Sekky61:
[edit]
This PR implements [ExperienceBarUpdate](https://wiki.vg/Protocol#Set_Experience) and [HealthUpdate](https://wiki.vg/Protocol#Set_Health) packets.

I decided not to introduce a new field for Total Experience, health or hunger on `Client`, storing that data and conversion between Total Experience and level with xp bar would be handled downstream.

There are two new functions on `Client`, `set_level` and `set_health_and_food`.
Their docs describe ranges of the arguments.
It is ok to send these packets on the same tick player connected, that is why I chose not to defer sending them.

Setting total_xp has no effect on the score on death screen, so I have no idea what it does. There is a todo in `set_level` docs.

### `HealthUpdate`
* Send packet on the same tick player connected
    * Outcome: packet **can** be sent on the same tick player connected (right after `GameJoin` packet)
* Send packet to player in creative mode
    * Outcome: works as expected, after changing gamemode back to survival, health bar was changed
* Send packet with health `<= 0`
    * Outcome: player dies. (Player is stuck now because respawn button does not work).
* Send packet with food `<= 0`, or food_saturation `<= 0`
    * Outcome: client does not crash, we can test and document this later when eating works.

### `ExperienceBarUpdate`
* Send packet on the same tick player connected
    * Outcome: packet **can** be sent on the same tick player connected (right after `GameJoin` packet)
* Send packet to player in creative mode
    * Outcome: works as expected, after changing gamemode back to survival, xp bar was changed
* Send packet with negative `total_xp`
    * Outcome: client did not crash

### Docs
* Documentation for the new helpers `set_level` and `set_health_and_food` rendered correctly.
---
Old:

Packets were tested manually.

API-wise, should these packets have a method on `Client`? If so, I will append this PR.

Also, it seems [Set Experience](https://wiki.vg/Protocol#Set_Experience) allows to set fields inconsistently (for example as `Total Experience = 0` and `Level = 5`). I couldn't find, if that could cause issues.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -32,11 +32,11 @@ use crate::protocol::packets::s2c::play::{
     BiomeRegistry, ChatTypeRegistry, ChunkLoadDistance, ChunkRenderDistanceCenter, ClearTitles,
     DimensionTypeRegistry, DimensionTypeRegistryEntry, Disconnect, EntitiesDestroy,
     EntityAnimation, EntityAttributes, EntityAttributesProperty, EntityPosition, EntitySetHeadYaw,
-    EntityStatus, EntityTrackerUpdate, EntityVelocityUpdate, GameJoin, GameMessage,
-    GameStateChange, GameStateChangeReason, KeepAlive, MoveRelative, OverlayMessage, PlaySoundId,
-    PlayerActionResponse, PlayerPositionLook, PlayerPositionLookFlags, PlayerRespawn,
-    PlayerSpawnPosition, RegistryCodec, Rotate, RotateAndMoveRelative, S2cPlayPacket,
-    SoundCategory, UnloadChunk, UpdateSubtitle, UpdateTitle,
+    EntityStatus, EntityTrackerUpdate, EntityVelocityUpdate, ExperienceBarUpdate, GameJoin,
+    GameMessage, GameStateChange, GameStateChangeReason, HealthUpdate, KeepAlive, MoveRelative,
+    OverlayMessage, PlaySoundId, PlayerActionResponse, PlayerPositionLook, PlayerPositionLookFlags,
+    PlayerRespawn, PlayerSpawnPosition, RegistryCodec, Rotate, RotateAndMoveRelative,
+    S2cPlayPacket, SoundCategory, UnloadChunk, UpdateSubtitle, UpdateTitle,
 };
 use crate::protocol::{BoundedInt, ByteAngle, NbtBridge, RawBytes, VarInt};
 use crate::server::{C2sPacketChannels, NewClientData, S2cPlayMessage, SharedServer};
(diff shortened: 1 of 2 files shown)


XBagon (changes requested):
Looks good, maybe you wanna add a `set_experience` and `set_health` method to `Client` :)
And for the float between 0 and 1 a new-type could be introduced which only successfully parses in that range.. but I guess that would more optional :P

dyc3 (commented):
I personally would love to see a better test plan but I'm not going to block you for it. Are we sure that these are f32 and not ints? Seems weird to me.

Sekky61:
The packets correctly affect the health/hunger/xp bar. I couldn't figure out a better way of testing this, any ideas for future? Or did sou just mean I should be more detailed in description of the testing?

dyc3:
Both. Anybody should be able to pull down your code and verify that it works. In order to do that you need to include instructions for how to test your code, and any sample code that you used to test with.

rj00a (changes requested):
Regarding `BoundedInt` `BoundedString` etc. those are mostly for catching bad serverbound packets as early as possible/convenience/documentation. It's fine to leave them off because as you said the server might want to do some intentionally janky things. Implementing BoundedFloat at some point would be good, but that's out of scope.

`set_level` and `set_health` (maybe `set_health_and_food`?) functions on `Client` should be made. One question is whether or not to defer sending the packet. By that I mean sending the packet directly in `set_level` or sending it in `Client::update`. The advantage with deferring is that if `set_level` is called multiple times in a single tick then only one packet is sent. You can also check that the new value is equal to the old value, in which case you can skip sending a packet altogether as an optimization because it has no effect. The disadvantage with deferring is that it is more code and it makes the `Client` struct bigger.

Sometimes you _have_ to defer sending packets because the join game/login packet is itself deferred, and a lot of packets don't work correctly if you send them before join game/login. To test this, try sending the packet on the same tick the client joins the game. If it works fine then you're good. If not, defer.

If (and only if) you decide to defer, make the appropriate getters as well.

Something interesting I realized is that there is actually multiple ways to set client health. It is possible to set the client's health through tracked entity data (`Player`).

Finally, the names of the packets should match [packets.json](https://github.com/valence-rs/valence/blob/main/extracted/packets.json) where reasonable. I should make a note of this somewhere.

Thank you for coming to my TED talk.

Sekky61:
Thank you for great feedback. I updated the first message with the current state of the PR.

dyc3:
> Player is stuck now because respawn button does not work

Sounds like a good follow up task. We should have a simple example where the player dies and is able to respawn.

rj00a (approved):

PR #35, opened 2022-09-08
Add support for commands & completion

XBagon:
I've added the struct for the [Commands](https://wiki.vg/Protocol#Commands) packet and implemented de/encoding for the [Node Format](https://wiki.vg/Command_Data#Node_Format) needed for Commands.
Most variations of [Parser](https://wiki.vg/Command_Data#Parsers) are still missing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,3 +70,9 @@ num = "0.4"
 
 [workspace]
 members = ["packet_inspector"]
+
+[profile.dev.package."*"]
+opt-level = 3
+
+[profile.dev]
+opt-level = 1
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -15,6 +15,7 @@ use serde::Serialize;
 use uuid::Uuid;
 pub use var_int::VarInt;
 pub use var_long::VarLong;
+pub use node::Node;
 use vek::{Vec2, Vec3, Vec4};
 
 use crate::entity::EntityId;
@@ -25,6 +26,7 @@ pub mod codec;
 pub mod packets;
 mod var_int;
 mod var_long;
+mod node;
 
 /// Types that can be written to the Minecraft protocol.
 pub trait Encode {
--- a/src/protocol/node.rs
+++ b/src/protocol/node.rs
@@ -1,0 +1,235 @@
+use std::io::{Read, Write};
+
+use anyhow::{bail, Context};
+use byteorder::{ReadBytesExt, WriteBytesExt};
+
+use crate::ident::Ident;
+use crate::protocol::{BoundedString, Decode, Encode, VarInt};
+
+#[derive(Clone, Debug)]
+pub struct Node {
+    children: Vec<VarInt>,
+    data: NodeData,
+    is_executable: bool,
+    redirect_node: Option<VarInt>,
+}
+
+impl Encode for Node {
+    fn encode(&self, w: &mut impl Write) -> anyhow::Result<()> {
+        let enum_id = match self.data {
+            NodeData::Root => 0,
+            NodeData::Literal(_) => 1,
+            NodeData::Argument(_) => 2,
+        };
+
+        let flags = enum_id
+            & (self.is_executable as u8 * 0x04)
+            & (self.redirect_node.is_some() as u8 * 0x08)
+            & (if let NodeData::Argument(argument) = &self.data {
+                argument.suggestions_type.is_some()
+            } else { false } as u8 * 0x04);
+
+        w.write_u8(flags)?;
+        self.children.encode(w)?;
+
+        if let Some(redirect_node) = self.redirect_node {
+            redirect_node.encode(w)?;
+        }
+
+        match &self.data {
+            NodeData::Root => {}
+            NodeData::Literal(literal) => {
+                literal.name.encode(w)?
+            }
+            NodeData::Argument(argument) => {
+                argument.name.encode(w)?;
+                argument.parser.encode(w)?;
+                if let Some(suggestions_type) = &argument.suggestions_type {
+                    suggestions_type.encode(w)?;
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl Decode for Node {
+    fn decode(r: &mut impl Read) -> anyhow::Result<Self> {
+        let flags = r.read_u8()?;
+
+        let is_executable = flags & 0x04 != 0;
+        let redirect_node = if flags & 0x08 != 0 {
+            Decode::decode(r)?
+        } else {
+            None
+        };
+
+        let children = Decode::decode(r)?;
+
+        let enum_id = flags & 0x03;
+        let data = match enum_id {
+            0 => NodeData::Root,
+            1 => NodeData::Literal(Literal {
+                name: Decode::decode(r)?,
+            }),
+            2 => NodeData::Argument(Argument {
+                name: Decode::decode(r)?,
+                parser: Decode::decode(r)?,
+                suggestions_type: if flags & 0x10 != 0 {
+                    Decode::decode(r)?
+                } else {
+                    None
+                },
+            }),
+            _ => bail!("Invalid NodeData variant")
+        };
+
+        Ok(Node {
+            children,
+            data,
+            is_executable,
+            redirect_node,
+        })
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum NodeData {
+    Root,
+    Literal(Literal),
+    Argument(Argument),
+}
+
+#[derive(Clone, Debug)]
+pub struct Literal {
+    name: BoundedString<0, 32767>,
+}
+
+#[derive(Clone, Debug)]
+pub struct Argument {
+    name: BoundedString<0, 32767>,
+    parser: Parser,
+    suggestions_type: Option<Ident>,
+}
+
+def_enum! {
+    Parser: VarInt {
+        BrigadierBool: bool = 0,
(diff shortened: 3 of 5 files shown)


src/protocol/packets/s2c.rs
```diff
@@ -227,6 +227,14 @@ pub mod play {
         }
     }
 
+    def_struct! {
+        Commands {
+            count: VarInt,
```
Sekky61: I think you should omit `count` field, as `Encode for Vec<T>` already encodes length.
https://github.com/valence-rs/valence/blob/c955ab8799ef23f27c9c222a4fff8743326682a3/src/protocol.rs#L599

Sekky61: Oh, didn't notice. Good job

src/protocol/node.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub parser: Parser,
+    pub suggestions_type: Option<Ident>,
+}
+
+def_enum! {
+    Parser: VarInt {
+        BrigadierBool: bool = 0,
+        BrigadierFloat: BrigadierFloat = 1,
+        BrigadierInteger: BrigadierInteger = 2,
+        BrigadierLong: BrigadierLong = 3,
```
Sekky61, line 121: `BrigadierInteger` should have value `3` and `BrigadierLong` should have value `4`

src/protocol/node.rs
```diff
@@ -0,0 +1,235 @@
...
+
+#[derive(Clone, Debug)]
+pub struct BrigadierFloat {
+    pub min: Option<f32>,
+    pub max: Option<f32>,
+}
+
+impl Encode for BrigadierFloat {
+    fn encode(&self, w: &mut impl Write) -> anyhow::Result<()> {
+        let flags = (self.min.is_some() as u8) << 0 & (self.max.is_some() as u8) << 1;
```
Sekky61, line 134: Use bitwise `OR` here

src/protocol/node.rs
```diff
@@ -0,0 +1,235 @@
...
+
+#[derive(Clone, Debug)]
+pub struct BrigadierInteger {
+    pub min: Option<i32>,
+    pub max: Option<i32>,
+}
+
+impl Encode for BrigadierInteger {
+    fn encode(&self, w: &mut impl Write) -> anyhow::Result<()> {
+        let flags = (self.min.is_some() as u8) << 0 & (self.max.is_some() as u8) << 1;
```
Sekky61, line 171: Same issue

src/protocol/node.rs
```diff
@@ -0,0 +1,235 @@
...
+
+#[derive(Clone, Debug)]
+pub struct BrigadierLong {
+    pub min: Option<i64>,
+    pub max: Option<i64>,
+}
+
+impl Encode for BrigadierLong {
+    fn encode(&self, w: &mut impl Write) -> anyhow::Result<()> {
+        let flags = (self.min.is_some() as u8) << 0 & (self.max.is_some() as u8) << 1;
```
Sekky61, line 208: Same issue

src/client.rs
```diff
@@ -28,16 +28,7 @@ use crate::protocol::packets::c2s::play::{
...
-    EntityStatus, EntityTrackerUpdate, EntityVelocityUpdate, GameJoin, GameMessage,
-    GameStateChange, GameStateChangeReason, KeepAlive, MoveRelative, OverlayMessage, PlaySoundId,
-    PlayerActionResponse, PlayerPositionLook, PlayerPositionLookFlags, PlayerRespawn,
-    PlayerSpawnPosition, RegistryCodec, Rotate, RotateAndMoveRelative, S2cPlayPacket,
-    SoundCategory, UnloadChunk, UpdateSubtitle, UpdateTitle,
-};
+use crate::protocol::packets::s2c::play::{BiomeRegistry, ChatTypeRegistry, ChunkLoadDistance, ChunkRenderDistanceCenter, ClearTitles, DimensionTypeRegistry, DimensionTypeRegistryEntry, Disconnect, EntitiesDestroy, EntityAnimation, EntityAttributes, EntityAttributesProperty, EntityPosition, EntitySetHeadYaw, EntityStatus, EntityTrackerUpdate, EntityVelocityUpdate, GameJoin, GameMessage, GameStateChange, GameStateChangeReason, KeepAlive, MoveRelative, OverlayMessage, PlaySoundId, PlayerActionResponse, PlayerPositionLook, PlayerPositionLookFlags, PlayerRespawn, PlayerSpawnPosition, RegistryCodec, Rotate, RotateAndMoveRelative, S2cPlayPacket, SoundCategory, UnloadChunk, UpdateSubtitle, UpdateTitle, Commands};
```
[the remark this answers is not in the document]
XBagon: yep didn't use rustfmt/clippy yet 😰

src/protocol/node.rs
XBagon: Oops skipped the Double.. thanks

src/protocol.rs
```diff
@@ -10,23 +10,23 @@ use arrayvec::ArrayVec;
...
 use crate::entity::EntityId;
 use crate::nbt;
 
 mod byte_angle;
 pub mod codec;
+pub mod node;
 pub mod packets;
 mod var_int;
 mod var_long;
-pub mod node;
```
Sekky61: It seems, with macros, order of modules is important. Try putting `node` under `packets`.
```rust
mod byte_angle;
pub mod codec;
pub mod packets;
mod var_int;
mod var_long;

pub mod node;
```

XBagon: that helped, thanks!

src/config.rs
```diff
@@ -11,6 +11,9 @@ use crate::dimension::Dimension;
 use crate::server::{NewClientData, Server, SharedServer};
 use crate::text::Text;
 use crate::{Ticks, STANDARD_TPS};
+use crate::protocol::{Node, VarInt};
+use crate::protocol::node::{Literal, NodeData};
+use crate::protocol::packets::s2c::play::Commands;
```
rj00a, line 16: Stuff from the `protocol` module should not be in the public API unless it's being reexported. The protocol module is `#[doc(hidden)]` and might be made private in the future.

src/config.rs
```diff
@@ -167,6 +170,28 @@ pub trait Config: Sized + Send + Sync + UnwindSafe + RefUnwindSafe + 'static {
         vec![Biome::default()]
     }
 
+    fn commands(&self) -> Commands {
```
rj00a, line 173: What happens if you send multiple `Commands` packets? Do subsequent packets delete the set of commands and replace them with new ones? If that's the case then adding commands should be done with `Client` and not `Config`.

Is it possible to construct a bad `Commands` packet using this that would disconnect players? If it is then we need to be doing validation to ensure that doesn't happen.

src/protocol/node.rs
```diff
@@ -0,0 +1,248 @@
+use std::io::{Read, Write};
+
+use anyhow::{bail, Context};
+use byteorder::{ReadBytesExt, WriteBytesExt};
+
+use crate::ident::Ident;
+use crate::protocol::{BoundedString, Decode, Encode, VarInt};
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Node {
```
rj00a, line 10: Not sure if I like the name `Node`. Seems a bit vague. Might be better to call this `CommandNode` for clarity when used outside the module. Especially since this is only used in one place AFAICT.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ bitvec = "1"
 byteorder = "1"
 cesu8 = "1.1.0"
 cfb8 = "0.7"
+derive_more = "0.99.17"
```
rj00a, line 27: Would like to avoid adding this dependency if possible.

src/commands.rs
```diff
@@ -0,0 +1,455 @@
...
+    (, $t:expr, $f:expr) => {
+        $t
+    };
+    ($typ:ty, $t:expr, $f:expr) => {
+        $f
+    };
+}
+
+#[macro_export]
+macro_rules! def_command {
```
rj00a, line 194: We should avoid exposing a macro based solution. Remember that Valence is supposed to be minimal and should only contain what is absolutely necessary to work with the protocol. Extra convenience stuff should be in another crate.

Besides, what I think users would want instead is a [derive macro](https://doc.rust-lang.org/reference/procedural-macros.html#derive-macros), and proc macros need to be in a separate crate anyway.

rj00a (changes requested):
When the PR is further along I will have more to say.

src/protocol/node.rs
XBagon: True, I just straight copied all names from https://wiki.vg/Protocol#Commands and https://wiki.vg/Command_Data for the moment.

Cargo.toml
XBagon: That's what I thought.. I guess I need to write another macro then for this and wrap the enum? 😬

src/commands.rs
XBagon: I should have discussed this beforehand, true.. I though about making a proc macro, but thought maybe it would work as is. But you're absolutely right. Another re-write will happen then eventually :P

qualterz:
Is the [Radix Tree](https://en.wikipedia.org/wiki/Radix_tree) data structure able to be used here for the command validation and completion?

rj00a:
Closing since this is old and we have #351 and #446.

PR #44, opened 2022-09-09
upgrade aes and cfb8 crates from 0.7 to 0.8

dyc3:
Right now, the unit tests for this will fail. Do not merge this as is, it will break clients being able to connect.

There are some breaking changes moving from 0.7 to 0.8. Namely, the cfb8 crate now provides 2 types `Encryptor<T>` and `Decryptor<T>` and the `encrypt()` and `decrypt()` methods now **consume the `cipher` instead of just borrowing it.** (eg. `fn encrypt(self, ...)` rather than `fn encrypt(&self, ...)`)

I'm putting this up as a draft so somebody else can pick it up if I don't get back around to it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ build = "build/main.rs"
 authors = ["Ryan Johnson <ryanj00a@gmail.com>"]
 
 [dependencies]
-aes = "0.7"
+aes = "0.8"
 anyhow = "1"
 approx = "0.5.1"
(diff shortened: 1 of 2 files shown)


rj00a:
> the encrypt() and decrypt() methods now consume the cipher instead of just borrowing it

That's the problem. The state of the cipher is supposed to change over time as new data comes in. How are we supposed to do that if the cipher is consumed in one go?

Issue #17, opened 2022-09-04
Accurate entity hitboxes


rj00a:
`entity.rs` has a function called `hitbox` which, as you can guess, returns the Axis-aligned bounding box of an entity.

Extracting correct hitbox data automatically appears to be infeasible because it depends on the entity's state and varies wildly between different entity types. This info will have to be filled in manually.

Right now there is a bunch of placeholder data and some `todo!`s.

PR #45, opened 2022-09-10
Update entity hitboxes

guac420:
This PR addresses issue #17 and adds special cases for the creation of `EntitySpawnPacket`s for item frames, falling blocks, and fishing bobbers.


guac420:
I tried to implement the hitboxes for paintings but they require the direction the painting is facing and that isn't tracked by `TrackedData::Painting` so I'm not sure how to go about it.

rj00a (changes requested):
Thanks. That looked tedious.

Since the size of the baby hitbox is just the normal hitbox with all dimensions divided by 2 (correct me if I'm wrong) I think it would be more maintainable to have a local function like:
```rust
fn baby(is_baby: bool, adult_hitbox: [f64; 3]) -> [f64; 3] {
    if is_baby { adult_hitbox.map(|a| a / 2.0) } else { adult_hitbox }
}
```
For the out of bounds values, returning a sensible default hitbox as you've done is fine. In the future we might be able to make improvements to the data extractor so we can eliminate those invalid states.

For the paintings, I think you can just derive its direction from the yaw of the entity. Have it "snap" to the closest cardinal direction.

guac420:
> Since the size of the baby hitbox is just the normal hitbox with all dimensions divided by 2 (correct me if I'm wrong) I think it would be more maintainable to have a local function like:
> 
> ```rust
> fn baby(is_baby: bool, adult_hitbox: [f64; 3]) -> [f64; 3] {
>     if is_baby { adult_hitbox.map(|a| a / 2.0) } else { adult_hitbox }
> }
> ```

Yea there are only like two exceptions to this, so the function makes more sense. Good idea for the paintings :) I'll get working on the changes.

src/entity.rs
```diff
@@ -613,6 +825,39 @@ impl<C: Config> Entity<C> {
...
+            })),
+            TrackedData::FallingBlock(_e) => Some(EntitySpawnPacket::Entity(EntitySpawn {
+                entity_id: VarInt(this_id.to_network_id()),
+                object_uuid: self.uuid,
+                kind: VarInt(self.kind() as i32),
+                position: self.new_position,
+                pitch: ByteAngle::from_degrees(self.pitch),
+                yaw: ByteAngle::from_degrees(self.yaw),
+                head_yaw: ByteAngle::from_degrees(self.head_yaw),
+                data: VarInt(0), // TODO: set block state id
```
rj00a, line 847: Would be good if this was `VarInt(1)` for the time being so that the falling block is stone instead of air.

guac420:
> Thanks. That looked tedious.
> 
> Since the size of the baby hitbox is just the normal hitbox with all dimensions divided by 2 (correct me if I'm wrong) I think it would be more maintainable to have a local function like:
> 
> ```rust
> fn baby(is_baby: bool, adult_hitbox: [f64; 3]) -> [f64; 3] {
>     if is_baby { adult_hitbox.map(|a| a / 2.0) } else { adult_hitbox }
> }
> ```
> 
> For the out of bounds values, returning a sensible default hitbox as you've done is fine. In the future we might be able to make improvements to the data extractor so we can eliminate those invalid states.
> 
> For the paintings, I think you can just derive its direction from the yaw of the entity. Have it "snap" to the closest cardinal direction.

So far I've made all of the changes except for implementing paintings. I'm just testing my implementation now to make sure that it works as expected.

src/entity.rs
```diff
@@ -422,6 +422,14 @@ impl<C: Config> Entity<C> {
     ///
     /// [interact event]: crate::client::ClientEvent::InteractWithEntity
     pub fn hitbox(&self) -> Aabb<f64> {
+        fn baby(is_baby: bool, adult_hitbox: [f64; 3]) -> [f64; 3] {
```
rj00a, line 425: Might even be able to use a closure here. It's even shorter because of the type inference.
```rust
let baby = |is_baby, adult_hitbox| if is_baby { adult_hitbox.map(|a| a / 2.0) } else { adult_hitbox };
```
Although the Rust compiler might complain about missing types. If that's the case then the function is fine. Doesn't really matter.

rj00a:
To demonstrate that these changes are correct, I think it would be a good idea to create a new example to replace `raycast.rs`.
- Call it `entity_raycast.rs` instead (since there will be a `block_raycast.rs` eventually).
- Have entities lined up in a row like the sheep currently are.
- Show some exotic entities like item frames and paintings in different orientations, shulkers moving up and down, normal and baby variants, etc.
- Send a message telling players to press F3+B to turn on hitboxes.
- Use the action bar to diplay "Intersection" in green or "No Intersection" in red.

Feel free to deviate from my suggestions if you have a better idea.

rj00a:
About to push a bunch of changes

rj00a:
Sorry, I'm bad at git and made a new branch + PR instead of adding to this one. I think you'll still get credit though because your commits are included.

#70

PR #48, opened 2022-09-10
Proxy Support

### LINKED ISSUE #39: Proxy Support
Valence should have support for [bungeecord](https://github.com/SpigotMC/BungeeCord) and [velocity](https://velocitypowered.com/) proxies. I think something like this could work in `Config`.

```rust
#[non_exhaustive]
#[derive(Copy, Clone, Eq, PartialEq, Default)]
pub enum ConnectionMode {
    #[default]
    Online,
    Offline,
    Bungeecord,
    Velocity,
}
```

(Looks like velocity does not support 1.19 currently? What should we do if Valence is updated before the proxies are?)

Tert0:
This PR adds Proxy Support.
# TODO
- [x] Velocity Support
- [x] Bungeecord Support
- [x] Add Plugin Message API (Client Event)
- [x] VELOCITY_MODERN_FORWARDING_WITH_KEY_V2

Closes #39


src/client.rs
```diff
@@ -373,6 +375,10 @@ impl<C: Config> Client<C> {
         self.msgs_to_send.push(msg.into());
     }
 
+    pub fn send_plugin_message(&mut self, msg: PluginMessageToClient) {
+        self.plugin_messages_to_send.push(msg);
```
rj00a, line 379: Any reason you can't send the plugin message packet here directly? Does the plugin message packet work if sent before the login/join game packet? Also this function should take the channel and data as args to avoid exposing the `PluginMessageToClient` in the public API.

src/config.rs
```diff
@@ -75,21 +75,51 @@ pub trait Config: Sized + Send + Sync + UnwindSafe + RefUnwindSafe + 'static {
...
+    /// It fetches all player data (uuid, username and skin) from the bungeecord proxy but does not
+    /// block connections, which are not from the proxy.
+    /// If the server is publicly exposed anyone can connect with any name, uuid and skin similar to offline mode.
+    ///
+    /// # `ConnectionMode::Velocity`
+    /// This should be used if the server runs behind a Velocity Proxy, which is configured to run
+    /// with the player info forwarding mode `modern`.
+    /// It fetches all player data (uuid, username and skin) from the velocity proxy and blocks
+    /// all connections, which are not from the proxy.
+    /// To ensure this the return of `velocity_secret()` has to match the secret key of the velocity proxy.
```
rj00a, line 106: Perhaps it would be better to move these docs to the variants of the `ConnectionMode` enum.

src/config.rs
```diff
@@ -249,3 +279,13 @@ pub enum ServerListPing<'a> {
...
 }
+
+#[non_exhaustive]
+#[derive(Copy, Clone, Eq, PartialEq, Default)]
+pub enum ConnectionMode {
+    #[default]
+    Online,
+    Offline,
+    Bungeecord,
+    Velocity,
```
rj00a, line 290: Can we attach the secret to the `Velocity` variant of this enum? Also I've made it a heap allocated string.
```suggestion
    Velocity {
        // TODO: docs
        secret: String,
    }
```

src/protocol/packets/c2s.rs
```diff
@@ -8,7 +8,17 @@ pub mod handshake {
...
             protocol_version: VarInt,
-            server_adddress: BoundedString<0, 255>,
+            server_address: BoundedString<0, 255>,
+            server_port: u16,
+            next_state: HandshakeNextState,
+        }
+    }
+
+    def_struct! {
+        BungeecordHandshake {
```
rj00a, line 18: Feel free to remove the `BoundedString` from `Handshake` and check the bounds later. Then you can remove `BungeecordHandshake`.

src/proxy/velocity.rs
```diff
@@ -0,0 +1,5 @@
+pub static VELOCITY_PLAYER_INFO_CHANNEL: &str = "velocity:player_info";
+pub const VELOCITY_SUPPORTED_VERSION: i32 = 1;
```
rj00a, line 2: Considering that Valence is a library, I'm a bit concerned about what the implications of this is for semver compatibility. Is updating support for velocity/bungeecord/waterfall a breaking change? Are versions of velocity/bungeecord/waterfall backwards compatible?

rj00a (changes requested):
Looking good so far!

src/client.rs
Tert0: > Any reason you can't send the plugin message packet here directly? Does the plugin message packet work if sent before the login/join game packet?
While in login state there are `Login Plugin Request` and `Login Plugin Response` packets, which are limited to server requests and clients responses. Thats why i'm not sure if directly sending it would work.

src/server.rs
```diff
@@ -564,7 +564,10 @@ async fn handle_connection<C: Config>(
 
     let handshake: Handshake;
     handshake = c.dec.read_packet::<Handshake>().await?;
-    ensure!(handshake.server_address <= 255 || server.connection_mode() == ConnectionMode::Bungeecord, "server address too long");
+    ensure!(
+        handshake.server_address.len() <= 255 || server.connection_mode() == ConnectionMode::Bungeecord,
```
rj00a, line 568: `.len()` returns the byte length of the string, not a count of characters. `BoundedString` counts the characters.

`.chars().count()` should do the trick.

src/client.rs
```diff
@@ -375,11 +375,9 @@ impl<C: Config> Client<C> {
         self.msgs_to_send.push(msg.into());
     }
 
-    pub fn send_plugin_message(&mut self, channel: Ident, data: BoundedArray<u8, 0, 1048576>) {
-        self.plugin_messages_to_send.push(PluginMessageToClient{
-            channel,
-            data,
-        });
+    pub fn send_plugin_message(&mut self, channel: Ident, data: RawBytes) {
```
rj00a, line 378: `RawBytes` is part of the `protocol` module. You can accept a `Vec<u8>` instead.

rj00a (changes requested):

src/client.rs
rj00a: By the time you have a `Client` you're already past the login stage and into the play stage so I think it would work.

Tert0:
I've read the source code of Velocity and the PaperMC Implementation

rj00a:
Reviewing this now. Will be pushing changes tomorrow.

rj00a (commented):
# Changes I've Made
- Cleaned up documentation in `config.rs`.
- The `handle_login` function was quite large so I moved all that logic into a new submodule.
- Removed the `proxy` module since there wasn't much there.
- Removed incorrect uses of `unwrap` and `expect`. (If we panic it would bring down the whole server).
- Used existing types to parse where applicable.
- Set the IP address of the client in the BungeeCord login procedure.
- Removed Velocity version check. Let me know if it was wrong to remove this.
- More descriptive error messages.
- Many stylistic adjustments and optimizations.

# Issues to Resolve Before Merging
- [ ] The second element of the NUL-delimited list in the BungeeCord procedure appears to be the client's IP address, but I'm not certain. Can you confirm?
- [ ] `VELOCITY_MODERN_FORWARDING_WITH_KEY_V2` is unimplemented. Probably not a blocker for merging. Not sure how important it is.
- [ ] After logging in with `ConnectionMode::BungeeCord` or `ConnectionMode::Velocity`, the client fails to fully join and is disconnected with the following error. This problem exists before and after my changes.

last-screenshot

Tert0:
> - The second element of the NUL-delimited list in the BungeeCord procedure appears to be the client's IP address, but I'm not certain. Can you confirm?
I think it's the IP you have used to conntect to the server. (e.g. 127.0.0.1 or localhost)
> - `VELOCITY_MODERN_FORWARDING_WITH_KEY_V2` is unimplemented. Probably not a blocker for merging. Not sure how important it is.
I think i fergot to Push These changes, so i'll push them later.

rj00a:
> I think it's the IP you have used to conntect to the server. (e.g. 127.0.0.1 or localhost)

When I tried it, the first element was `localhost` while the second was `127.0.0.1`. So I think the first is the address the player used to connect (possibly requiring DNS lookup) while the second is the address of the connecting player. Correct me if I'm wrong.

rj00a (approved):

rj00a:
I'm going to go ahead and merge what we have here to avoid conflicts.

PR #49, opened 2022-09-11
WIP: building example

dyc3:
This adds a new example to demonstrate how to implement block breaking/placing.

In order for this to work, I also had to:
- update definition for UpdateCreativeModeSlot

I'm putting this up so people know it exists, I'll finish it up later.

# Test plan

1. run `cargo r -r --example building`
2. start minecraft and join localhost
3. See that breaking and placing blocks works

NOTE: We don't keep track of inventories yet, so no matter what block you place it's going to be dirt instead.


src/block_pos.rs
```diff
@@ -23,6 +24,28 @@ impl BlockPos {
     pub fn at(pos: impl Into<Vec3<f64>>) -> Self {
         pos.into().floor().as_::<i32>().into()
     }
+
+    /// Get a new [`BlockPos`] that is a adjacent to this position in `dir`
```
rj00a, line 28:
```suggestion
    /// Get a new [`BlockPos`] that is adjacent to this position in `dir`
```

src/block_pos.rs
```diff
@@ -23,6 +24,28 @@ impl BlockPos {
...
+    ///
+    /// ```rust
+    /// use valence::block::BlockPos;
+    /// use valence::client::BlockFace;
+    ///
+    /// let pos = BlockPos::new(0, 0, 0);
+    /// let adj = pos.get_in_direction(BlockFace::South);
+    /// assert_eq!(adj, BlockPos::new(0, 0, 1));
+    /// ```
+    pub fn get_in_direction(&self, dir: BlockFace) -> BlockPos {
```
rj00a, line 39: Should take `self` by value because `BlockPos` is `Copy`.
```suggestion
    pub fn get_in_direction(self, dir: BlockFace) -> BlockPos {
```

src/client.rs
```diff
@@ -47,7 +47,7 @@ use crate::world::{WorldId, Worlds};
 use crate::{ident, LIBRARY_NAMESPACE};
 
 /// Contains the [`ClientEvent`] enum and related data types.
-mod event;
+pub mod event;
```
rj00a, line 50: There is a `pub use event::*;` that exports everything so this doesn't need to be public.
```suggestion
mod event;
```

examples/building.rs
```diff
@@ -0,0 +1,233 @@
+use std::net::SocketAddr;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use log::LevelFilter;
+use num::Integer;
+use valence::biome::Biome;
+use valence::block::BlockState;
+use valence::chunk::{Chunk, UnloadedChunk};
+use valence::client::event::DiggingStatus;
```
rj00a, line 9:
```suggestion
use valence::client::DiggingStatus;
```

rj00a (changes requested):
I'll merge this but we should definitely fix this up when the inventory stuff is further along. Also we need to check that the player is not placing a block that intersects with their own hitbox.

examples/building.rs
dyc3: `valence::protocol::packets::c2s::play::DiggingStatus` conflicts with `valence::client::DiggingStatus`, so making this change makes it not compile. I'm going to go ahead and remove that import and just use the fully qualified name `valence::protocol::packets::c2s::play::DiggingStatus` in `client.rs`

rj00a (approved):

PR #51, opened 2022-09-12
Dying and respawning

Sekky61:
This PR implements dying and respawning of player and `death.rs` example.

## Unresolved

- [ ] `DeathMessage` packet has no effect when sent on the tick client was created.
- [ ] Investigate `NoRespawnBlockAvailable` message in `GameStateChange` packet.
- [ ] Investigate `PlayerRespawn`'s field `copy_metadata`.
- [ ] Interaction with compass. I think we should solve this in another PR (after implementing `Inventory` packet etc.).

### `Client::kill` method

Method for killing the client. Allows to set a message on the death screen.

### `Death.rs` example

To demonstrate (and to follow up on [this comment](https://github.com/valence-rs/valence/pull/34#issuecomment-1242363870)) I also created an example.

It shows how to enable death screen, kill player and respawn, including respawning in another `world`.
Once player inventory is working, interaction with compass would be added to the example.

## Tests

Following tests were conducted in the `death.rs` example.

* Behavior with and without death screen enabled
    * When Death screen is *enabled*, it will appear. Respawn button is clickable and works.
    * When Death screen is *disabled*, respawn is immediate.
    * In both instances, `ClientStatus` packet will be sent.
* `GameJoin.enable_respawn_screen` packet field
    * When set to `true` enables death screen exactly as `GameStateChange` packet with `EnableRespawnScreen` option.
* `DeathMessage` packet sent on the first tick a client joined
    * Has no effect on the client.
* Respawn after receiving `GameStateChange` with `WinGame` reason
    * Works with both rolling the credits and immediate respawn options.

### DIFF
--- a/examples/death.rs
+++ b/examples/death.rs
@@ -10,7 +10,6 @@ use valence::config::{Config, ServerListPing};
 use valence::dimension::Dimension;
 use valence::entity::{EntityId, EntityKind};
 use valence::player_list::PlayerListId;
-use valence::protocol::packets::s2c::play::{GameEvent, GameStateChangeReason};
 use valence::server::{Server, SharedServer, ShutdownResult};
 use valence::text::{Color, TextFormat};
 use valence::world::WorldId;
@@ -37,7 +36,6 @@ struct Game {
 #[derive(Default)]
 struct ClientState {
     entity_id: EntityId,
-    enable_death_screen_packet_sent: bool,
     // World and position to respawn at
     respawn_location: (WorldId, Vec3<f64>),
     // Anticheat measure
@@ -193,21 +191,13 @@ impl Config for Game {
                         None,
                     );
 
+                client.set_respawn_screen(true);
+
                 client.send_message("Welcome to the death example!".italic());
                 client.send_message("Step over the left line to die. :)");
                 client.send_message("Step over the right line to die and respawn in second world.");
                 client.send_message("Jumping down kills you and spawns you in another dimension.");
                 client.send_message("Sneaking triggers game credits after which you respawn.");
-            }
-
-            if !client.state.enable_death_screen_packet_sent && !client.created_this_tick() {
-                // This packet enables death screen
-                // TODO create helper function, this must be deferred
-                client.send_packet(GameEvent {
-                    reason: GameStateChangeReason::EnableRespawnScreen,
-                    value: 0.0,
-                });
-                client.state.enable_death_screen_packet_sent = true;
             }
 
             // TODO after inventory support is added, show interaction with compass.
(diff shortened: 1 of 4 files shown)


rj00a:
> Client has a new flag `can_respawn`. This prevents malicious clients from misusing `ClientStatus::PerformRespawn` packets.

Validation of client events should happen downstream, or at least not in `Client` directly. There are a few reasons for this.
- Anticheat developers are going to want to know when clients are trying to do something they're not supposed to be doing.
- Trying to validate client events is a bit of a slippery slope. This packet is easy to verify, but there are a lot of packets that aren't or don't really have a single solution. Movement is a good example. I believe users are going to want to tailor their checks for their own unique situations.

It's better to think about client events as _requests_. It is then the Valence user's job to decide if they want to honor the request or not.

I will clarify my stance in the docs.

src/block_pos.rs
```diff
@@ -88,6 +88,19 @@ impl From<BlockPos> for Vec3<i32> {
...
+    fn into(self) -> Vec3<f64> {
+        let BlockPos { x, y, z } = self;
+        Vec3 {
+            x: x as f64,
+            y: y as f64,
+            z: z as f64,
+        }
+    }
+}
+
```
rj00a, line 103: You should never `impl Into` in Rust. Use `From` instead. How about
```rust
impl From<BlockPos> for Vec3<f64> {
    fn from(pos: BlockPos) -> Self {
        Vec3::new(pos.x.into(), pos.y.into(), pos.z.into())
    }
}
```

rj00a (changes requested):

Sekky61:
I hope I finally understand what should and what shouldn't be in scope of Valence now. Took a while :sweat_smile:.

src/block_pos.rs
Sekky61: Yeah I misunderstood the blanket `impl`, I wanted to avoid lossy conversions from `f64` to `i32`. Anyway, I think it may not be necessary, so I will undo it.

rj00a:
Yeah I'm still figuring it out myself. Valence is very unusual in this way.

Sekky61:
Conflicts resolved

rj00a:
Is this ready for review?

Sekky61:
Yes

rj00a:
I made some changes to the example I think you will agree with.

I'm assuming you plan on addressing the bare `send_packet`s in the example before this is merged?

Sekky61:
Great changes, so much cleaner.

I created the methods, we can certainly discuss the API. I also created a discussion on Discord about solving packet deferring, I think we need a more general solution.

rj00a:
Posting this here as well for posterity.
> I've just thought of a ~~awful hack~~ genius solution to solve this. Very soon in #74 I plan on changing the way packets work by appending them to a buffer in `Client` instead of sending them though a channel. This buffer of packets would then be sent all at once at the end of the tick to the tokio task. Anyway, what if we just... insert the `LoginPlay` packet at the _beginning_ of this buffer before sending it? Then we don't need to defer packets at all!
>
> So perhaps you can write your PR as if this was implemented. Then I'll add it later.

rj00a (approved):

Sekky61:
Alright, it is no longer deferred. So once we redo packet dispatching, the example will work fully.

Issue #10, opened 2022-09-04
"packet contents were not decoded completely (4 bytes remaining)"


axionbuster:
Using commit `2ee9605` (latest `main` as of 11:50 PM Sat, Sep 3, 2022, PST), on an M1 mac, running example `terrain`, abridged log follows:

```
[2022-09-04T06:42:16Z DEBUG valence::protocol::codec] complete packet after partial decode: C2sPlayPacket(UpdateCreativeModeSlot { slot: 36 })
[2022-09-04T06:42:16Z ERROR valence::server] connection to 127.0.0.1:52167 ended: error during play: packet contents were not decoded completely (4 bytes remaining)
[2022-09-04T06:42:22Z DEBUG valence::server] error while flushing packet queue: Broken pipe (os error 32)
```

# Effects

The client is disconnected with a simple message, "Timed Out," while attempting to join.

It stops just short of the "Loading Terrain" stage.

# How to reproduce

Use Java client 1.19.2 from the official launcher.

I haven't been able to consistently reproduce it. However, repeatedly joining in-and-out can cause this problem by chance.

But to describe my actions,

1. Successfully enter the world. Destroy a grass block. Replace it with a stone block.
2. Quit world.
3. Attempt to join again.
4. (Get this error).
5. Attempt to join again.
6. (Fail to join due to another cause).
7. Inventory is cleared, and I'm allowed to join again.

# Difficulty reproducing the error

Often I get an initial disconnect with a "deadline has elapsed" error. Afterwards, I keep getting "conflicting UUID" errors so I have to reset the server. So entering the world successfully is kind of chance, and once I get kicked out I will keep getting kicked out, so it has to be perfect every time.

I haven't been able to reproduce this issue twice.

rj00a:
#38 fixes the "conflicting UUID" issue but I'm still getting `broken pipe` errors.

DarkWugWug:
I can look into this one. Can be reliably recreated by pressing your middle mouse button on any block or by putting something in your inventory. 

Done with (96f561494190338ed0186561135e1ce14cb9a177) patched with #38.

PR #52, opened 2022-09-12
Cleaner fix to #10

DarkWugWug:


rj00a:
I'll merge this but I don't think this resolves #10, at least not fully. I'm still getting `broken pipe` errors by disconnecting and reconnecting repeatedly.

rj00a:
Guess we should open a separate issue for that.

Issue #56, opened 2022-09-14
Migrate packet names to wiki.vg names


rj00a:
I've had a rule in place that the names of packets should match the names in [packets.json](https://github.com/valence-rs/valence/blob/main/extracted/packets.json). The reason for this was to make updating to new versions using the data extractor easier and to avoid potential legal issues because wiki.vg used the packet names from Mojang's mappings (this has mostly changed).

However, a lot of the yarn names are crap and it's way easier to just use wiki.vg. So we should transition all or most of the packet names to their names on wiki.vg for consistency.

Edit: This should be a gradual process so we don't cause too much pain right now.

EmperialDev:
Can I take this one? (I'm very new to Git)

TkTech:
> avoid potential legal issues

Just wanted to mention that some of those names were updated on the wiki _by_ Mojang employees, and they use the wiki internally. It would be very hard for them to argue infringement given both tacit (usage and contribution) and explicit (tweets, mentions) approval.

rj00a:
> Just wanted to mention that some of those names were updated on the wiki by Mojang employees

Interesting! Had no idea Mojang used it themselves. But yea you're right.

PR #62, opened 2022-09-14
Migrating the packets to wiki.vg

EmperialDev:
Adresses #56. Migrating all the packet names to wiki.vg's insted of fabric's.

I have also copied the documentation from wiki.vg into the code as doc comments, so it's easier to see what the diffrent packets are for.


rj00a:
I would avoid documenting the packets because that info can become outdated/incorrect very easily. Also there is no need to change the field names.

As I merge other PRs there are going to be a lot of merge conflicts so apologies in advance.

EmperialDev:
That's okay. I will remove the documentation and just rename the packets.

rj00a (changes requested):
- Packets that share identical names across groups should distinguish themselves with a suffix. For instance, `Disconnect` has login and play variants, so they should be named `DisconnectLogin` and `DisconnectPlay`.
- Packets that have the same name across serverbound and clientbound boundaries should be distinguished similarly with a `C2s` or `S2c` suffix.

Other than that looks good. Just need to `cargo fmt`.

rj00a (approved):

Issue #54, opened 2022-09-13
Broken pipe error when repeatedly connecting and disconnecting


rj00a:
Repeatedly connecting to the server and disconnecting while loading will eventually cause `Failed to log in: null` in the client and
```
[2022-09-13T22:46:24Z ERROR valence::server] connection to 127.0.0.1:36030 ended: error during login: Broken pipe (os error 32)
```
In the server.

I'm not sure if this is actually a bug in Valence or expected Linux socket behavior. It would be interesting to test this on different platforms.

rj00a:
> Just to confirm, this is happening on the main branch?
Yeah

dyc3:
I think a much easier way to reproduce this is just to run an example and spam refresh. Sometimes, valence doesn't respond to the ping.

rj00a:
I have been unable to reproduce the `broken pipe` by spam refreshing.

dyc3:
It doesn't print anything to the logs, it just kinda hangs. I'll record a video tomorrow

LenaGmbh:
I looked into this problem and it turns out it's actually a Minecraft client error, this happens also on vanilla servers.

PR #65, opened 2022-09-15
packet_inspector: fix UnexpectedEof errors deep inside packet decoding causing perfectly good connections to be aborted. fixes #64

### LINKED ISSUE #64: Can't join vanilla server through packet_inspector proxy.
## How to reproduce

start a vanilla mc server
```
docker run -e EULA=TRUE -e ONLINE_MODE=false -d -p 25565:25565 --name mc itzg/minecraft-server
```

run packet inspector
```
cargo r -r -p packet_inspector -- 127.0.0.1:25566 127.0.0.1:25565
```

Open mc and connect to `localhost:25566`
See that you never load into the world

## Notes

The disconnect occurs after the server sends the first `ChunkData` packet. The full error is
```
Error: Err(
    Error {
        context: "decoding packet after decompressing",
        source: Error {
            context: "failed to read field `block_light_arrays` from struct `ChunkData`",
            source: Error {
                kind: UnexpectedEof,
                message: "failed to fill whole buffer",
            },
        },
    },
)
```
This is not displayed by the packet inspector logs currently, because `UnexpectedEof` errors are explicitly ignored with this code.
```rust
if let Some(e) = e.downcast_ref::<io::Error>() {
    if e.kind() == ErrorKind::UnexpectedEof {
        return Ok(());
    }
}
```

dyc3:
fixes #64

### Test plan

start a vanilla mc server
```
docker run -e EULA=TRUE -e ONLINE_MODE=false -d -p 25565:25565 --name mc itzg/minecraft-server
```

run packet inspector
```
cargo r -r -p packet_inspector -- 127.0.0.1:25566 127.0.0.1:25565
```

Open mc and connect to `localhost:25566`
See that you can load into the world and play as normal.

### DIFF
--- a/packet_inspector/src/main.rs
+++ b/packet_inspector/src/main.rs
@@ -172,7 +172,7 @@ async fn handle_connection(client: TcpStream, cli: Cli) -> anyhow::Result<()> {
                         .await
                     {
                         if let Some(e) = e.downcast_ref::<io::Error>() {
-                            if e.kind() == ErrorKind::UnexpectedEof {
+                            if e.kind() == ErrorKind::BrokenPipe {
                                 return Ok(());
                             }
                         }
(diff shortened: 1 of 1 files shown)


rj00a:
Need to investigate further to make sure this doesn't cause issues elsewhere

dyc3:
I don't think it would affect #54 at all? This is just a change to the packet inspector, not to valence itself. Funnily enough, I can't seem to reproduce that issue with this setup.

rj00a:
Don't think this is needed anymore due to #146. Closing.

PR #67, opened 2022-09-16
Packet filtering

rj00a:
Adds a regex argument to the packet_inspector allowing it to be somewhat useful for once.

See the updated README for instructions.


dyc3 (approved):
lol I was literally about to implement something like this

PR #68, opened 2022-09-16
Custom resource packs

dessh:
Implements custom resource pack support.

Adds a `Client` method to set resource pack, and a `ClientEvent` for listening to resource pack status updates from clients.


src/client.rs
```diff
@@ -640,6 +640,30 @@ impl<C: Config> Client<C> {
...
     }
 
+    /// Requests that the client download and enable a resource pack.
+    ///
+    /// # Arguments
+    /// * `url` - The URL of the resource pack file.
+    /// * `hash` - The SHA-1 hash of the resource pack file.
+    /// * `forced` - Whether a client should be kicked from the server upon
+    ///   declining the pack (this is enforced client-side)
+    /// * `prompt-message` - A message to be displayed with the resource pack
```
dyc3, line 650:
```suggestion
    /// * `prompt_message` - A message to be displayed with the resource pack
```

dyc3 (commented):
Looks fine to me, but how do we test this? Could you write up some instructions to try this out?

dessh:
I updated the conway.rs example to show a working custom resource pack, for demonstration's sake. Let me know if you'd like this taken out of the example or for me to make a new example instead.

I've also noticed that we cannot set a client's resource pack immediately after they are created (in the same tick). I'd assume that deferring the call until the next tick would solve this issue, but am unsure if that is possible at the moment (something like Bukkit's `BukkitRunnable.runTask()`)

examples/conway.rs
```diff
@@ -218,6 +228,28 @@ impl Config for Game {
...
                     }
+                    ClientEvent::InteractWithEntity { kind, .. } => {
+                        if kind == InteractWithEntityKind::Attack {
+                            if client.state.resource_pack_active {
+                                client.send_message("You already have the resource pack!".italic());
+                                continue;
+                            }
+
+                            client.set_resource_pack(
+                                "https://download942.mediafire.com/j0ag4kfe1ftg/0ds8poaabjc8snw/pack.zip"
```
dyc3, line 239: Let's avoid pinging mediafire all the time. If you upload the pack as an attachment here, I think you'll get a usable URL.

dyc3 (changes requested):
It should definitely go in it's own example.

examples/conway.rs
rj00a: Should the resource pack go in `assets/`?

dyc3: Yeah, that would probably be better. We could just have the resource pack in the repo. For this PR, that will need to land first so we can have the correct URL from github.

src/client.rs
```diff
@@ -640,6 +640,30 @@ impl<C: Config> Client<C> {
...
+    /// * `url` - The URL of the resource pack file.
+    /// * `hash` - The SHA-1 hash of the resource pack file.
+    /// * `forced` - Whether a client should be kicked from the server upon
+    ///   declining the pack (this is enforced client-side)
+    /// * `prompt_message` - A message to be displayed with the resource pack
+    ///   dialog.
+    pub fn set_resource_pack(
+        &mut self,
+        url: String,
+        hash: Option<String>,
```
rj00a, line 655: You can make these `impl Into<String>` and `impl Into<Option<String>>`.

rj00a (changes requested):
As @dyc3 said, resource pack tests should go in a separate example.

You can use the mediafire link for now. In a separate commit we will add the resource pack to the project and change the URL.

examples/conway.rs
dessh: I'll just leave the URL as localhost in the new example, and open another PR to change it when the pack.zip has been merged into main.

dessh:
> As @dyc3 said, resource pack tests should go in a separate example.
> 
> You can use the mediafire link for now, but you should add the resource pack to `assets/`. In a separate commit we will change the URL.

I'll just leave the URL as localhost in the new example, and open another PR to change it when the pack.zip has been merged into main.

src/client/event.rs
```diff
@@ -128,6 +128,7 @@ pub enum ClientEvent {
         /// Sequence number
         sequence: VarInt,
     },
+    ResourcePackStatus(ResourcePackC2s),
```
rj00a, line 131: We should reexport the ResourcePackC2s enum here because stuff in the `protocol` module is private/hidden.

```rust
pub use crate::protocol::packets::c2s::play::ResourcePackC2s as ResourcePackStatus;
```

rj00a (changes requested):
> I've also noticed that we cannot set a client's resource pack immediately after they are created (in the same tick). I'd assume that deferring the call until the next tick would solve this issue, but am unsure if that is possible at the moment (something like Bukkit's BukkitRunnable.runTask())

The usual solution to this problem is to send the packet in `Client::update` instead of directly in `set_resource_pack`. The problem is that some packets don't work when sent before `LoginPlay`. Fixing this involves adding a field to `Client` to store the data you want to send.

What happens if the client receives the resource pack packet multiple times in the same tick? Does the new resource pack prompt immediately override the old prompt? In that case I think it might be better to have an `Option<ResourcePackS2c>` on `Client` rather than a `Vec<ResourcePackS2c>`.

dessh:
> What happens if the client receives the resource pack packet multiple times in the same tick? Does the new resource pack prompt immediately override the old prompt? In that case I think it might be better to have an `Option<ResourcePackS2c>` on `Client` rather than a `Vec<ResourcePackS2c>` if there is no observable difference.

Yeah, the new resource pack prompt overrides the old one on the client immediately, and the client sends back a "Failed" C2s for the old prompt, which means that `Option<ResourcePackS2c>` is probably the way to go.

src/client.rs
```diff
@@ -641,6 +643,30 @@ impl<C: Config> Client<C> {
...
+    /// * `url` - The URL of the resource pack file.
+    /// * `hash` - The SHA-1 hash of the resource pack file.
+    /// * `forced` - Whether a client should be kicked from the server upon
+    ///   declining the pack (this is enforced client-side)
+    /// * `prompt_message` - A message to be displayed with the resource pack
+    ///   dialog.
+    pub fn set_resource_pack(
+        &mut self,
+        url: impl Into<String>,
+        hash: impl Into<Option<String>>,
```
rj00a, line 658: I think it would be best to make this a required argument since it is required by the packet. It would also be good to describe the constraints on the value of the string.

src/protocol/packets/s2c.rs
```diff
@@ -667,6 +667,15 @@ pub mod play {
         }
     }
 
+    def_struct! {
+        ResourcePackS2c {
+            url: BoundedString<0, 32767>,
```
rj00a, line 672: 32767 is the max char length of strings so the `BoundedString` here is useless.

rj00a (changes requested):

src/client.rs
dessh: The reason I didn't make it required is because hash checking on server resource packs is optional, and will be skipped if the server sends a non-hash string (like ""). So we would need to explicitly state in the doc that to skip hash checking you need to pass an empty string in (which I felt like `Option` would be more suited for in this case)

rj00a: The use of `Option` here implies a semantic difference between `None` and `Some("")`, but there is not. There isn't any additional type safety gained by adding an Option. If we really wanted to we could use more precise types but that isn't really worth the effort.

dessh: Yeah that's true, I can see how people might get confused over `None` vs `Some("")`

dessh: > (except that the string should be lowercase).
Hex strings are case-insensitive and I checked that an uppercase hash works, so we can probably leave this out.

rj00a: Yeah I guess the wiki.vg info is outdated. I will have to update it.

rj00a (approved):

PR #70, opened 2022-09-17
Entity Hitbox Improvements

rj00a:
Includes the changes from #45 plus a large number of improvements
- Lots of cosmetic changes that don't change the semantics but improve readability, consistency, and reduce the number of lines of code.
- Redesigned the shulker hitbox code.
- Fixed rounding bug in painting rotation logic.
- Fixed bug in item frame hitbox code (there was a `+=` instead of a `-=`).
- Fixed glow item frame object data not being sent.
- Added a shulker bullet cursor to the `entity_raycast.rs` to show where the intersection is. This would probably be done better with particles but we don't have those yet.
- More entities in `entity_raycast.rs`.

# Testing
Run the `entity_raycast` example (replaces `raycast`).

# Things Left to Resolve
- [x] The slime/magma cube hitbox code was incorrect because the info on wiki.vg is outdated. I made a guess that appears to be correct but confirmation would be good.
- [x] The Warden emerging hitbox is wrong (@guac420 do you have the numbers?).


rj00a:
Btw @guac420 Check this out for the painting rotations https://www.desmos.com/calculator/m97ozjy5cz

PR #72, opened 2022-09-18
Add methods to Config to use a custom session server

### LINKED ISSUE #40: Configurable authentication server
The official Mojang authentication server address probably shouldn't be hardcoded. It should default to the official one in `Config` but be configurable.

tsatke:
Added new methods to the Config trait, that enables implementors to use a fully custom URL.

Closes #40

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,7 +1,7 @@
 //! Configuration for the server.
 
 use std::borrow::Cow;
-use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4};
 use std::panic::{RefUnwindSafe, UnwindSafe};
 
 use async_trait::async_trait;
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -208,6 +208,50 @@ pub trait Config: Sized + Send + Sync + UnwindSafe + RefUnwindSafe + 'static {
...
+        username: &str,
+        auth_digest: &str,
+        server_ip: &IpAddr,
+    ) -> Cow<'_, str> {
+        let host = self.session_server_host(server);
+        let path_and_args = format!(
+            "session/minecraft/hasJoined?username={}&serverId={}&ip={}",
+            username, auth_digest, server_ip
+        );
+        Cow::from(format!("https://{}/{}", host, path_and_args))
```
dyc3, line 252: It would be nice to have the URL scheme be configurable too.

tsatke: It is. The method can be overridden.

dyc3: oh right, duh. carry on

dyc3 (approved):

src/config.rs
```diff
@@ -208,6 +208,50 @@ pub trait Config: Sized + Send + Sync + UnwindSafe + RefUnwindSafe + 'static {
...
+    /// Providing a URL that does not return such a structure will result in a
+    /// disconnect for every client that connects.
+    ///
+    /// The arguments are described in the linked wiki article.
+    fn format_session_server_url(
+        &self,
+        server: &SharedServer<Self>,
+        username: &str,
+        auth_digest: &str,
+        server_ip: &IpAddr,
```
rj00a, line 245: The argument should be called `player_ip`. This can also be passed in by value since it is `Copy`.

src/config.rs
rj00a, line 252: We should use Rust's string interpolation feature. I've also eliminated one string allocation.
```rust
format!("https://{host}/session/minecraft/hasJoined?username={username}&serverId={auth_digest}&ip={server_ip}").into()
```

src/config.rs
```diff
@@ -208,6 +208,50 @@ pub trait Config: Sized + Send + Sync + UnwindSafe + RefUnwindSafe + 'static {
...
+    ///
+    /// This method is called from within the default implementation of
+    /// [`Self::session_server_url`]. If that method is overridden, this
+    /// method will not be called.
+    ///
+    /// # Default Implementation
+    ///
+    /// The official mojang session server (`sessionserver.mojang.com`)
+    /// is used.
+    fn session_server_host(&self, server: &SharedServer<Self>) -> Cow<'_, str> {
```
rj00a, line 223: Don't think it's worth having a second function for this. If you're overriding the session server you probably know what you're doing.

src/server.rs
```diff
@@ -704,9 +704,16 @@ async fn handle_login<C: Config>(
...
+        let hex_hash = auth_digest(&hash);
+
+        let url = C::format_session_server_url(
+            server.config(),
+            server,
+            &username,
+            &hex_hash,
+            &remote_addr.ip(),
+        )
+        .to_string();
```
rj00a, line 716: `to_string` is not needed because `get` can take a `&str`. (Also you would want to use `into_owned` since that avoids an unnecessary clone)

src/config.rs
```diff
@@ -208,6 +208,50 @@ pub trait Config: Sized + Send + Sync + UnwindSafe + RefUnwindSafe + 'static {
...
+    /// disconnect for every client that connects.
+    ///
+    /// The arguments are described in the linked wiki article.
+    fn format_session_server_url(
+        &self,
+        server: &SharedServer<Self>,
+        username: &str,
+        auth_digest: &str,
+        server_ip: &IpAddr,
+    ) -> Cow<'_, str> {
```
rj00a, line 246: It's probably best to return a `String` instead of a `Cow<str>` because realistically the returned value is always a formatted string.

rj00a (changes requested):

src/server.rs
tsatke: `to_string` and `to_owned` do the same thing, but `to_string` is more clear.
https://users.rust-lang.org/t/to-string-vs-to-owned-for-string-literals/1441/5

rj00a: I was actually referring to [`into_owned`](https://doc.rust-lang.org/stable/std/borrow/enum.Cow.html#method.into_owned), not `to_owned`.

rj00a:
Thanks

PR #73, opened 2022-09-18
Replace the bool in the retain functions with a retain decision

tsatke:
The functions for retain calls are now expected to return
a more explicit RetainDecision.
The RetainDecision implements Try, so that server implementors can use the ?-operator to short-circuit their implementation of those functions.

Helped me a lot in my little server project, especially because the retain functions can grow quite a bit, and this helped me splitting it up better.
Also improves readability IMO, since I don't have to figure out, what exactly the `return true;` in that place means.

### DIFF
--- a/examples/building.rs
+++ b/examples/building.rs
@@ -11,6 +11,7 @@ use valence::config::{Config, ServerListPing};
 use valence::dimension::{Dimension, DimensionId};
 use valence::entity::{EntityId, EntityKind};
 use valence::player_list::PlayerListId;
+use valence::retain::RetainDecision;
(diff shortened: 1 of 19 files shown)


tsatke:
Sorry for the other two commits, those are from #72 , and I shouldn't have done them on the main branch.

rj00a:
Probably won't be doing this because
- The `retain` functions in Valence are meant to be consistent with the standard library.
- Unstable Rust features are used. We are not using nightly Rust (but we are using unstable rustfmt features).
- This is a lot of additional code for what I consider to be a nonissue.

Issue #53, opened 2022-09-13
Inventory and Items


rj00a:
The lack of inventory and items can no longer be ignored. I will be working on this next.

- [x] Item data extractor
- [x] Item code generator
- [x] Inventory packets
- [x] Inventory client events
- [ ] `inventory.rs`?
- [x] New example

dyc3:
If we could split this out into smaller tasks, I'd like to take on some of the work.

rj00a:
I'll see if I can do that.

dyc3:
I've been thinking about how to structure the actual inventory types, and I think I've figured out something that might work.

Each player has an associated `PlayerInventory`:
```rust
pub struct PlayerInventory {
    armor_slots: [Slot; 4],
    offhand: Slot,
    crafting_slots: [Slot; 4],
    crafting_result: Slot,
    item_slots: Vec<Slot>,
}
```
This will allow us to keep track of the user moving items around in their inventory. The hard part is that (annoyingly) the slot numbers change depending on what type of window is open.

To account for this, we have a `WindowInventory` that will be created from the desired object's inventory, and the player's inventory. It should look something like this.

```rust
pub struct WindowInventory<T> {
    window_type: WindowType,
    object_inv: T,
    player: PlayerInventory,
}
```

Basically, this would provide a translation layer to the underlying player and object inventories. This would also mean that we _need_ to wrap all inventories in `Arc<Mutex<T>>` to make sure there aren't any data races, so really it would look like

```rust
pub struct WindowInventory<T> {
    window_type: WindowType,
    object_inv: Arc<Mutex<T>>,
    player: Arc<Mutex<PlayerInventory>>,
}
```

Reference: https://wiki.vg/Inventory

dyc3:
Another way we could do object inventories is have it be an enum with all possible inventory types instead of having the seperate window_type field. However, I think this would make it really hard to expand valence to handle modded, non-vanilla items and inventories.

rj00a:
I like this idea. I think you're on the right track but I have a few comments. I still don't have a complete understanding of inventories but here is what I think.

```rust
pub struct PlayerInventory {
    armor_slots: [Slot; 4],
    offhand: Slot,
    crafting_slots: [Slot; 4],
    crafting_result: Slot,
    item_slots: Vec<Slot>,
}
```

This can be represented more efficiently by using a single fixed-sized array of `Slot`, e.g. `[Slot; ?]`. This would also map more naturally to the way it's represented in the packets.

[...]

PR #75, opened 2022-09-20
Set up Inventory module

dyc3:
This is a very rough version of the inventory module and associated features. This will be a good base to start building something more robust.

- adds a player inventory to all clients
- makes clients keep track of what inventory they have open
- adds very naive state sync for inventories

related: #53 

# Test plans

### Building

1. run `building` example
2. pick any block from creative
3. place it
4. pick a different block from creative
5. place 2
6. break the last one
7. see that both remaining blocks don't become dirt

### Chest inventories

1. run `chest` example
2. sneak to give yourself a stone block
3. open the chest (it's invisible on top of the stone block)
4. place stone in chest
5. see the stone moving around in the chest

## Known Issues

To be fixed in later PRs

- It's very easy to duplicate items if you take the stone out of the chest on the same tick as the server trying to move the item.
- The chest is invisible when you spawn in. It appears if you break it.
- `get_slot` and `set_slot` can panic.


src/client.rs
```diff
@@ -962,7 +971,9 @@ impl<C: Config> Client<C> {
             C2sPlayPacket::SeenAdvancements(_) => {}
             C2sPlayPacket::SelectTrade(_) => {}
             C2sPlayPacket::SetBeaconEffect(_) => {}
-            C2sPlayPacket::SetHeldItemS2c(_) => {}
+            C2sPlayPacket::SetHeldItemS2c(e) => {
+                self.selected_hotbar_slot = PlayerInventory::hotbar_to_slot(e.slot.0).unwrap();
```
rj00a, line 975: A cheater can send a bad slot ID and crash the server due to `unwrap`.

src/inventory.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        Some(Self::hotbar_slots().start + hotbar_slot)
+    }
+}
+
+impl Default for PlayerInventory {
+    fn default() -> Self {
+        Self {
+            // Can't do the shorthand because Slot is not Copy.
+            slots: [
```
rj00a, line 49: as a workaround you can do `std::array::from_fn(|_| Slot::Empty)`

src/inventory.rs
```diff
@@ -0,0 +1,197 @@
...
+                Slot::Empty,
+                Slot::Empty,
+            ],
+        }
+    }
+}
+
+impl Inventory for PlayerInventory {
+    fn get_slot(&self, slot_id: SlotId) -> &Slot {
+        if slot_id < 0 {
```
rj00a, line 103: This checks the lower bound of `slot_id` but not the upper bound. The later array index will panic with a generic message.

src/inventory.rs
```diff
@@ -0,0 +1,197 @@
...
+
+/// Represents what the player sees when they open an object's Inventory.
+///
+/// This exists because when an object inventory screen is being shown to the
+/// player, it also shows part of the player's inventory so they can move items
+/// between the inventories.
+#[derive(Debug)]
+pub struct WindowInventory<T>
+where
+    T: Inventory,
```
rj00a, line 131: Nitpicking but I don't think it's idiomatic to put a trait bound on a struct definition unless it _needs_ to be that trait for the definition to work. Also maybe you can call the generic type `I` instead of `T` in this module.

src/inventory.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    fn set_slot(&mut self, slot_id: SlotId, slot: Slot) {
+        if slot_id < 0 {
+            // TODO: dont panic
+            panic!("invalid slot id")
+        }
+        self.slots[slot_id as usize] = slot;
+    }
+
+    fn capacity(&self) -> usize {
```
rj00a, line 118: Calling this `capacity` is confusing because `capacity` means something entirely different in the context of `Vec`. How about `slot_count`?

src/inventory.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// General slots are the slots that can hold all items, including the
+    /// hotbar, excluding offhand. These slots are shown when the player is
+    /// looking at another inventory.
+    pub fn general_slots() -> Range<SlotId> {
+        9..45
+    }
+
+    pub fn hotbar_slots() -> Range<SlotId> {
+        36..45
+    }
```
rj00a, line 34: These should be associated constants.
```suggestion
    pub const GENERAL_SLOTS: Range<SlotId> = 9..45;
    pub const HOTBAR_SLOTS: Range<SlotId> = 36..45;
```

src/inventory.rs
```diff
@@ -0,0 +1,197 @@
...
+
+pub trait CraftingInventory {
+    fn craft_result_slot() -> SlotId;
+    fn craft_table_slots() -> Range<SlotId>;
+}
+
+/// Represents a player's Inventory.
+#[derive(Debug, Clone)]
+pub struct PlayerInventory {
+    pub(crate) slots: [Slot; 46],
```
rj00a, line 21: If `PlayerInventory` is allowed to be owned by library users (Which is the case currently because of the `Default` impl) then we should probably box this field. Don't want to be passing this massive amount of data around on the stack.
```suggestion
    pub(crate) slots: Box<[Slot; 46]>,
```

rj00a (changes requested):
Not sure I want to merge this into `main` while the inventory API is still so uncertain. Can you continue development on this branch? Or maybe we can make a dedicated branch in this repo. Then I can be less nitpicky until we want to merge it into `main`.

examples/building.rs
```diff
@@ -218,8 +219,19 @@ impl Config for Game {
...
-                            world.chunks.set_block_state(place_at, BlockState::DIRT);
+                            if let Slot::Present {
+                                item_id,
+                                item_count: _,
+                                nbt: _,
+                            } = client.held_item()
+                            {
+                                // FIXME: this itemid to block state conversion does not work as you
+                                // would expect. we need some utilities for this
+                                world.chunks.set_block_state(
```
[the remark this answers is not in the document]
dyc3: I actually did not know this lmao.

src/inventory.rs
dyc3: The trait bound is necessary. `WindowInventory` acts like a translation layer between the 2 displayed inventories. In order to do that, the other object needs to impl `Inventory`.

dyc3:
I'm going to make this a draft for now while I work on other functionality to make this easier.

src/inventory.rs
rj00a: I mean the bound in the struct definition. Not in the later `impl`.

src/inventory.rs
dyc3: I think `len()` might imply that all the returned slots are filled, so I'll go with `slot_count()`

dyc3:
@rj00a, how would you feel about making user facing API out of scope for this PR? This is already getting pretty massive and it would help me nail down the internals. I can make a follow up PR later that refactors the `chest` example so that it's not manually sending packets.

rj00a:
Sure thing. Whatever is easier for you.

rj00a:
To start, ensure that the code conforms to the new CONTRIBUTING.md guidelines.

dyc3:
Could you be a little more specific with how you want the structs and impls ordered? I think it makes the most sense as is, but the top-down modules thing is throwing me off.

rj00a:
Hopefully things are a bit clearer now. If the guideline is still too weird I can remove it.

examples/chest.rs
```diff
@@ -0,0 +1,305 @@
...
+                        let cell_x = chunk_x * 16 + x;
+                        let cell_z = chunk_z * 16 + z;
+
+                        if cell_x < SIZE_X && cell_z < SIZE_Z {
+                            chunk.set_block_state(x, 63, z, BlockState::GRASS_BLOCK);
+                        }
+                    }
+                }
+            }
+        }
```
rj00a, line 131: You can initialize the platform with `chunks.set_block_state` which would be a bit simpler.

src/inventory.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    // TODO: `entry()` style api
+
+    fn slots(&self) -> Vec<Slot> {
+        (0..self.slot_count())
+            .map(|s| self.slot(s as SlotId))
+            .collect()
+    }
+
+    fn consume_one(&mut self, slot_id: SlotId) {
```
rj00a, line 27: Probably shouldn't have a dedicated function in the trait to do this.

rj00a (changes requested):
- In the chest example you can keep pressing crouch until you have a stack of 128, at which point the stack completely disappears. If you go to 256 the server crashes because an overflow happens.
  - This indicates to me that we might need to enforce a max of 128 items in the `ItemStack` type.
- The use of `Slot` in its current form in `inventory.rs` is problematic because
    - `Slot` is defined in the protocol module but exposed publicly.
    - `Slot` is essentially a less convenient version of `Option<ItemStack>`.
  
  However, I believe we can just ditch the current `Slot` enum and make `Slot` a type alias of `Option<ItemStack>`. We should still be able to implement `Encode` and `Decode` on `Slot` without overlapping impls.

---

I know you said you wanted to build this in smaller PRs, but this is going to need a lot more polish before I would feel comfortable merging this into `main`.

dyc3:
> I believe we can just ditch the current Slot enum and make Slot a type alias of Option<ItemStack>

I'm going to go ahead and make a separate PR for this to avoid merge conflicts later.

src/inventory.rs
dyc3: I agree, but I'm intentionally not focusing on the user API right now. I fully intend to change this later.

src/inventory.rs
```diff
@@ -0,0 +1,303 @@
...
+use rayon::prelude::ParallelIterator;
+
+use crate::client::Clients;
+use crate::config::Config;
+use crate::protocol::packets::s2c::play::SetContainerContent;
+use crate::protocol::{Slot, SlotId, VarInt};
+use crate::slab_versioned::{Key, VersionedSlab};
+
+pub trait Inventory {
+    fn slot(&self, slot_id: SlotId) -> Slot;
```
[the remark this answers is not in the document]
dyc3: Could've sworn there was something that was preventing me from doing that...

src/inventory.rs
```diff
@@ -0,0 +1,303 @@
...
+        self.slab.get(inv.0)
+    }
+
+    pub fn get_mut(&mut self, inv: InventoryId) -> Option<&mut ConfigurableInventory> {
+        self.slab.get_mut(inv.0)
+    }
+
+    pub(crate) fn sync<C: Config>(&mut self, clients: &mut Clients<C>) {
+        // sync open, dirty inventories to clients
+        let _obj_inventories_cleaned: HashSet<InventoryId> = clients
```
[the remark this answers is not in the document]
dyc3: We need to do this every tick to keep inventories in sync with clients. This is very much not the most optimal way to do it, but it is the most simple. Using [SetContainerSlot](https://wiki.vg/Protocol#Set_Container_Slot) instead of shoving the whole inventory down the pipe is on my todo list for later.

As for collecting, we need the collect because iterators are lazy by default. The HashSet part is left over from a previous implementation, I'll fix that.

rj00a:
Also realized we probably don't want to use `Slot` in `inventory.rs` because the getters should return `Option<&Slot>` rather than `&Option<Slot>`.

src/inventory.rs
```diff
@@ -0,0 +1,303 @@
...
+    }
+}
+
+impl Inventory for PlayerInventory {
+    fn slot(&self, slot_id: SlotId) -> Slot {
+        if slot_id < 0 || slot_id >= self.slot_count() as i16 {
+            // TODO: dont panic
+            panic!("invalid slot id")
+        }
+        self.slots[slot_id as usize].clone()
```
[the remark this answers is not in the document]
dyc3: I tried this and it just made it more complicated

src/client.rs
```diff
@@ -1568,6 +1589,29 @@ impl<C: Config> Client<C> {
...
+                        // However, it should be possible to avoid the clone if this packet
+                        // could consume refs
+                        .map(|s| s.cloned())
+                        .collect(),
+                    carried_item: self.cursor_held_item.clone(),
+                },
+            );
+            self.inventory.state_id = self.inventory.state_id.wrapping_add(1);
+            self.inventory.mark_dirty(false);
+        }
```
rj00a, line 1614: This should happen above where the packets are flushed. Otherwise the packets will arrive a tick late.

src/server.rs
```diff
@@ -448,6 +451,7 @@ fn do_update_loop<C: Config>(server: &mut Server<C>) -> ShutdownResult {
         });
 
         server.player_lists.update();
+        server.inventories.sync(&mut server.clients);
```
rj00a, line 454: The problem here is that this function sends packets to clients, but the client packets have already been flushed. I think the work of this function that sends packets to clients should be done in the client's `update()` function near where the client's own inventory is updated. Then this function can be renamed to `update()` and will only have to mark the inventories as clean.

rj00a (changes requested):

src/client.rs
dyc3: Just to clarify, is this the line that flushes the packets?

```rust
send_packet(&mut self.send, S2cPlayMessage::Flush);
```

rj00a: Yeah

rj00a:
👍

PR #76, opened 2022-09-20
Set time of day #55

### LINKED ISSUE #55: Set time of day
There needs to be a function on `Client` to set the time of day it sees as well as stop time from passing.

How do we stop time from passing? This can be done using a fixed time in the dimension config, but vanilla MC can change this dynamically using the `doDaylightCycle` gamerule.

dani162:
Add function to set client time
Fixes #55

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -35,7 +35,7 @@ use crate::protocol::packets::s2c::play::{
     SetEntityVelocity, SetExperience, SetHeadRotation, SetHealth, SetRenderDistance,
     SetSubtitleText, SetTitleText, SoundCategory, SynchronizePlayerPosition, SystemChatMessage,
     TeleportEntity, UnloadChunk, UpdateAttributes, UpdateEntityPosition,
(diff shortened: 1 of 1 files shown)


rj00a:
The formatting checks have failed. You will have to run `cargo fmt`.

src/client.rs
```diff
@@ -673,6 +673,15 @@ impl<C: Config> Client<C> {
         });
     }
 
+    /// Sets the world_age and the current ingame time.
+    /// 
+    /// To stop the time from passing, the time_of_day parameter must to be
+    /// called with a negativ value. The client stops the time at the
+    /// absolute value.
```
rj00a, line 680:
```suggestion
    /// Sets the world_age and the current in-game time.
    /// 
    /// To stop time from passing, the time_of_day parameter must be
    /// negative. The client stops the time at the absolute value.
```

rj00a (changes requested):

rj00a (approved):

PR #79, opened 2022-09-21
Listen on any network interfaces #77

### LINKED ISSUE #77: Cannot reach server from another host when config address function is not overwritten
Currently we can not reach the server from another host if you not overwrite the ip address in the config.
I think we should document this behavior and maybe also change the default behavior to accept request from all interfaces (use address `0.0.0.0` instead of `localhost`).

https://github.com/valence-rs/valence/blob/00668839bdfa026b15e7b5b0dd1cabb3f16d25a3/src/config.rs#L52-L60

To override the address we currently have to implement following function:
```rust
impl Config for Game {
    ...
    fn address(&self) -> SocketAddr {
        SocketAddrV4::new(Ipv4Addr::new(0, 0, 0, 0), 25565).into()
    }
}
```
See also: https://stackoverflow.com/questions/42435723/rust-tcplistener-does-not-response-outside-request

I can create a pr tomorrow, if needed

dani162:
Server was only reachable over localhost (127.0.0.1)
Fixes #77

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -54,9 +54,9 @@ pub trait Config: Sized + Send + Sync + UnwindSafe + RefUnwindSafe + 'static {
     ///
     /// # Default Implementation
     ///
-    /// Returns `127.0.0.1:25565`.
+    /// Returns `0.0.0.0:25565` to listen on every available network interface.
     fn address(&self) -> SocketAddr {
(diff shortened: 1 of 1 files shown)


dyc3 (approved):
Maybe this should be localhost for dev build, all interfaces for release builds?

Either way, lgtm

PR #87, opened 2022-09-26
Refactor `Slot`

### LINKED ISSUE #78: Refactor `Slot` so that it's easier to work with
Working with slots is somewhat cumbersome because sometimes you want to represent that the item is present no matter what. 

- [ ] Add a new type `ItemStack` with the fields that are in `Slot::Present`
- [ ] Change `Slot::Present { .. }` to `Slot::Present(ItemStack)`
- [ ] Implement `From<Slot> for Option<ItemStack>` and vice versa

OR, alternatively we could do this instead:
- [ ] Get rid of slot and just use `Option<ItemStack>`

dyc3:
- create ItemStack type
- move slot back into protocol
- change DropItemStack to use ItemStack, because it's imposible to drop nothing

closes #78


src/itemstack.rs
```diff
@@ -0,0 +1,9 @@
+use crate::nbt::Compound;
+use crate::protocol::VarInt;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct ItemStack {
+    pub item_id: VarInt,
```
rj00a, line 6: We will want this type to be an `ItemId` or something similar in the future.

PR #88, opened 2022-09-27
check gradlew for the extractor into git

dyc3:
This is how it's intended to be used. See https://stackoverflow.com/questions/20348451/why-should-the-gradle-wrapper-be-committed-to-vcs


rj00a:
Seems crazy to me that gradle is designed this way but I'll accept it.

PR #92, opened 2022-09-28
Add Items

EmperialDev:
Adding all the items, just like the blocks are. This will adress 1. and 2. bullet from issue #53  
This will also make it easier to convert between block <-> item.

Open to feedback, right now the implementation is just a copy of the way the blocks are done.

I'm thinking of adding a `Vec<Prop>` to store props in an enum, such as the hunger the item will restore, attack damage and mining level? and mabye other things, idk?


extractor/src/main/java/dev/_00a/valence_extractor/extractors/Items.java
```diff
@@ -0,0 +1,33 @@
...
+    public JsonElement extract() throws Exception {
+        var itemsJson = new JsonArray();
+
+        for (var item : Registry.ITEM) {
+            var itemJson = new JsonObject();
+            itemJson.addProperty("id", Registry.ITEM.getRawId(item));
+            itemJson.addProperty("name", Registry.ITEM.getId(item).getPath());
+            itemJson.addProperty("translation_key", item.getTranslationKey());
+            itemJson.addProperty("max_stack", item.getMaxCount());
+
```
dyc3, line 27: Is this everything we need to be able to convert items to block states? Go ahead and grab as much info as you can.

build/item.rs
```diff
@@ -0,0 +1,170 @@
...
+            // Added to make it the same as BlockKind
+            #[allow(clippy::should_implement_trait)]
+            pub fn from_str(name: &str) -> Option<Item> {
+                match name {
+                    #item_from_str_arms
+                    _ => None
+                }
+            }
+
+            /// Get the snake_case name of this block kind.
```
dyc3, line 145: fix this doc comment

build/item.rs
```diff
@@ -0,0 +1,170 @@
...
+                match id {
+                    #item_from_raw_id_arms
+                    _ => None
+                }
+
+                //if id <= #items_count.try_into().unwrap() {
+                //    Some(Self(id))
+                //} else {
+                //    None
+                //}
```
dyc3, line 123: nit: remove commented code

build/item.rs
```diff
@@ -0,0 +1,170 @@
...
+
+            /// Gets the translation key of an item.
+            pub const fn translation_key(self) -> &'static str {
+                match self {
+                    #item_to_translation_key_arms
+                }
+            }
+
+            /// Returns the max stack count
+            pub const fn get_max_count(self) -> i16 {
```
dyc3, line 160: nit: should this be `usize`?

dyc3 (changes requested):
Ah good, I was just about to get started on this.

build/item.rs
EmperialDev: I thought about just using an `u8` since the count shouldn't get above 64, but in the Minecraft Code it's stores as an int, so i thought it would be best to use an `i16` so they were compatible.

build/item.rs
EmperialDev: Ups 😅

build/item.rs
rj00a: The Minecraft protocol uses a byte to represent item counts so making `max_stack` a `u8` should be fine. There will be a serde error at build time if the value is out of range for some reason.

Also, since this function calls it `max_count` we should rename `max_stack` to `max_count`.

src/item.rs
```diff
@@ -0,0 +1,22 @@
...
+impl Encode for Item {
+    fn encode(&self, w: &mut impl std::io::Write) -> anyhow::Result<()> {
+        VarInt(self.to_raw() as i32).encode(w)
+    }
+}
+
+impl Decode for Item {
+    fn decode(r: &mut &[u8]) -> anyhow::Result<Self> {
+        let id = VarInt::decode(r)?.0;
+        let errmsg = "invalid block state ID";
```
rj00a, line 18: Not a block state.

build/item.rs
```diff
@@ -0,0 +1,170 @@
...
+                //} else {
+                //    None
+                //}
+            }
+
+            /// Gets the raw item ID from the item
+            pub const fn to_raw(self) -> u16 {
+                match self {
+                    #item_to_raw_id_arms
+                }
```
rj00a, line 130: If we make the discriminant of the `Item` enum be the item ID, we can implement this function as
```rust
self as u16
```

Then the definition of `Item` would look something like this. (If you know the IDs are sequential starting at 0 you can skip this part. But I think it would be good style to do this anyway)

```rust
enum {
    // (idk what the actual item IDs are)
    Apple = 0,
    Stone = 1,
    Potato = 2,
    ...
}
```

The implementation of `from_raw` would stay the same.

See also: https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations

rj00a (changes requested):
"varient" should be spelled as "variant"

rj00a:
I don't think a `Vec<Prop>` like thing is necessary here? Best to just add more associated functions on `Item`.

build/item.rs
EmperialDev: Yeah, the item ids are sequential form 0, so should I still add the item ids in the enum? I have implemented it now where there's no item ids in the enum, but `self as u16` works.

rj00a: Yea you can leave the item IDs off.

build/block.rs
```diff
@@ -522,9 +579,47 @@ pub fn build() -> anyhow::Result<TokenStream> {
...
+        ///
+        /// `Normal` is just a single BlockState
+        ///
+        /// `Wall` is to BlockStates one for the normal variant and one for the wall variant
+        ///
+        /// `Cauldren` is the diffrent cauldren varients put into one
+        pub enum BlockStateType {
+            Normal(BlockState),
+            Wall(WallBlockState),
+            Cauldren(CauldronBlockState)
```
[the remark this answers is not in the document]
EmperialDev: Ups, good catch 😅

EmperialDev:
Should I also add associated functions to convert `Item` <-> `BlockState` for Item, like I just did for BlockState?

rj00a (changes requested):
Looks like we should be mapping `BlockKind`s to items and items to `BlockKind`s (you can get a `BlockState` from a `BlockKind`). I don't think the `BlockStateType` type is needed for this. And yeah associated functions on `Item` would be great.

Also while you're at it could you rename `Item` to `ItemKind`? I think it would avoid some confusion between `Item` and `ItemStack` and would be consistent with `BlockKind`/`BlockState`.

EmperialDev:
> Looks like we should be mapping `BlockKind`s to items and items to `BlockKind`s (you can get a `BlockState` from a `BlockKind`). I don't think the `BlockStateType` type is needed for this.

I would still need to have some sort of custom return type, because for example Torch has two entries in the `BlockKind` enum, it has `BlockKind::Torch` and `BlockKind::WallTorch`, so something is needed, so both of them are returned.

And the Cauldron is special, because it has four diffrent entries: `BlockKind::Cauldron`, `BlockKind::WaterCauldron`, `BlockKind::LavaCauldron` and `BlockKind::PowderSnowCauldron`, so I don't know how to return all of them, without a custom return type.

A fix would mabye be to remove the wall variants from `BlockKind` and than have them stored inside `BlockState`, so `BlockState` will have an associated function like `pub fn wall_variant() -> Option<BlockState> { .. }` and `pub fn faced_varient(face: BlockFace) -> BlockState { .. }`?

And there could be a special associated function for Cauldrons like `pub fn cauldron_varients() -> Option<CauldronVarients> { .. }`?

rj00a:
I have a suggestion
```rust
// Maps items to zero or one block kinds.
// e.g. `ItemKind::Cauldron` would map to `Some(BlockKind::Cauldron)`.
pub const fn to_block_kind(self) -> Option<BlockKind> { ... }

// Maps items to any number of block kinds (including zero).
// e.g. `ItemKind::Cauldron` maps to `&[BlockKind::Cauldron, BlockKind::LavaCauldron, ...]`.
pub const fn to_block_kinds(self) -> &'static [BlockKind] { ... }
```

And the other way around would just be:
```rust
// Multiple BlockKinds can map to the same ItemKind.
// e.g. BlockKind::Cauldron and BlockKind::LavaCauldron both become ItemKind::Cauldron.
pub const fn to_item(self) -> Option<ItemKind> { ... }
```

It makes sense that the `minecraft:torch` item would become a `minecraft:torch` block kind because that is the only block that has that name.

Using this knowledge I believe it is possible to implement these functions without making special cases for torches, cauldrons, etc.

EmperialDev:
I just realized we don't need to make anything special for the different cauldron variants, since you can only place the empty variant from an item 😅.
> It makes sense that the minecraft:torch item would become a minecraft:torch block kind in the to_block_kind function because that is the only block that has that name.

Yeah but how would you then know that there is a wall varient? There are 44 blocks like torch, where they have a normal and wall varient.

>```rust
>pub const fn to_block_kinds(self) -> &'static [BlockKind] { ... }

I just don't think this is a very clean solution, since you would always need to check the length of the array, to know if there's wall varient.

Therefore I think it would be a good idea to make something like:
```rust
/// Returns the wall block state of this block state
///
/// If this block state dos not have a wall state, `None` is returned.
pub fn wall_state(Self) -> Option<BlockState> { ... }
```

rj00a:
> ```rust
> /// Returns the wall block state of this block state
> ///
> /// If this block state dos not have a wall state, `None` is returned.
> pub fn wall_state(Self) -> Option<BlockState> { ... }
> ```

Ok I agree functions like this are a good idea. But in that case we shouldn't need `BlockKindType`.

EmperialDev:
Yeah, I also think that `BlockKindType` was a very bad idea, but I just need something for it to work.

rj00a:
Can we merge the `itemstack` module into the `item` module?

EmperialDev:
Yeah of course, itemstack is also a bit empty. I will do that.

rj00a:
Some of the changes I've made:

- Changed the output of the item extractor a bit, removing some unnecessary `Option`s and adding status effects.
- The procedure used to determine the wall state of blocks was inherently brittle. To do it properly I believe we will need to extract more block information. It's a bit out of scope for this PR anyway, so I've removed it for now.
- Code generators no longer depend on each other.
- Lots of syntax changes for consistency.

If there are no objections, I'll go ahead and merge this.

rj00a (approved):

dyc3 (approved):

EmperialDev:
They look good the changes you made. So just merge it.

PR #93, opened 2022-09-28
Add basic enchantment extraction

TerminatorNL:
This adds basic enchantment extraction into the JSON file. Note: No rust parser has yet been written.

Do we want the top level to be an array or an object as currently defined in [enchants.json](https://github.com/valence-rs/valence/compare/main...TerminatorNL:valence:enchant-extract?expand=1#diff-bdab9517c426197356200e180fd7375d4598b135622b05cc63efbcdb62378c99)? I used the other extracted files as reference and they use objects.

### DIFF
--- a/extracted/enchants.json
+++ b/extracted/enchants.json
@@ -8,10 +8,11 @@
       "max_level": 4,
       "rarity_weight": 10,
       "cursed": false,
-      "sources": [
-        "enchantment_table",
-        "random_selection"
-      ]
+      "sources": {
+        "treasure": false,
+        "enchantment_table": true,
+        "random_selection": true
+      }
     },
     {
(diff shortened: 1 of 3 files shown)


dyc3:
> Do we want the top level to be an array or an object as currently defined in enchants.json? 

An object. Makes it less of a pain to refactor the code gen if we need to add more info in this json.

rj00a (changes requested):
I think it would make more sense to have the sources be individual boolean fields rather than a list of strings. It would be easier to work with using tools such as serde_json.

Edit: I see on discord that you originally went with what I'm suggesting. I think your original approach was just fine.

rj00a (approved):

PR #100, opened 2022-10-03
Skip spawn of non visible player characters

### LINKED ISSUE #22: Don't spawn player entities unless they're on the player list
Player entities are invisible when spawned unless they were on the client's player list. We should check that the player entity is on the list before spawning them in `client.rs`.

Agreon:
closes #22

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1453,15 +1453,11 @@ impl<C: Config> Client<C> {
                     .get(id)
                     .expect("entity IDs in spatial index should be valid at this point");
 
-                // We want to skip player entities that are not in the players player_list.
+                // Skip spawning players not in the player list because they would be invisible
+                // otherwise.
                 if entity.kind() == EntityKind::Player {
                     if let Some(list_id) = &self.new_player_list {
-                        let entry = player_lists.get(list_id).get_entry(&entity.uuid());
-
-                        #[allow(clippy::question_mark)]
-                        if entry.is_none() {
-                            return None;
-                        }
+                        player_lists.get(list_id).entry(entity.uuid())?;
                     }
                 }
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -1459,9 +1459,8 @@ impl<C: Config> Client<C> {
                     if let Some(list_id) = &self.new_player_list {
                         let entry = player_lists.get(list_id).get_entry(&entity.uuid());
 
-                        match entry {
-                            None => return None,
-                            _ => {}
+                        if let None = entry {
+                            return None;
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


Agreon:
@rj00a would you like to have a look?

src/client.rs
```diff
@@ -1453,6 +1453,19 @@ impl<C: Config> Client<C> {
                 let entity = entities
                     .get(id)
                     .expect("entity IDs in spatial index should be valid at this point");
+
+                // We want to skip player entities that are not in the players player_list.
+                if entity.kind() == EntityKind::Player {
+                    if let Some(list_id) = &self.new_player_list {
+                        let entry = player_lists.get(list_id).get_entry(&entity.uuid());
+
+                        #[allow(clippy::question_mark)]
```
dyc3, line 1462: would it be possible to not ignore this lint?

dyc3 (approved):
Seems pretty trivial, but I don't have enough context to know if this is OK.

src/client.rs
Agreon: Unfortunately not. Adding a question mark as suggested would change the implementation. We don't want to return the entity if it is found. Only return early if it is not. There is a bugfix in clippy that should come with the next release.

rj00a:
Made a few tweaks.

Also I don't think clippy is wrong here? If the entry is `None`, we need to return early with `None`, which is exactly what the question mark operators does.

rj00a (commented):
Made a few tweaks.

Also I don't think clippy is wrong here? If the entry is `None`, we need to return early with `None`, which is exactly what the question mark operator does.

rj00a:
Oh, and we also need to skip spawning the player if the player list is `None` (empty list).

src/client.rs
```diff
@@ -1452,6 +1452,17 @@ impl<C: Config> Client<C> {
                 let entity = entities
                     .get(id)
                     .expect("entity IDs in spatial index should be valid at this point");
+
+                // Skip spawning players not in the player list because they would be invisible
+                // otherwise.
+                if entity.kind() == EntityKind::Player {
+                    if let Some(list_id) = &self.player_list {
+                        player_lists.get(list_id).entry(entity.uuid())?;
```
Agreon, line 1460: Correct me if I'm wrong, but doesn't this line mean, that we would return the player entity if it is found? Then we would skip the spawn logic below (1470-1576)?

rj00a: These lines say:
1. If the entity is a player, then...
2. If the client has a player list, then...
3. Get the entry with the entity's UUID. If the value is `Some` (the entry exists), do nothing. Otherwise, return early with `None` (the entry does not exist).

dyc3 (approved):

Issue #74, opened 2022-09-20
Performance Improvements


rj00a:
- [x] #81
- [x] #86
- [x] #82
- [x] #83
- [x] #84
- [x] #85
- [ ] #141

PR #117, opened 2022-10-16
Wall blocks

### LINKED ISSUE #115: Wall blocks not placed right
Right now if you place a torch on the wall the client will show a ghost wall torch, but when you update it, it will go back to a normal torch. And this is also true for all other blocks that have a wall variant.

I think it would be easy to just check for the diffrent blocks that have a wall variant, and then just return the right variant based on the direction. Mabye also for the rotational blocks, e.g. logs etc.

I propose something like:
```rust
// this will be inside ClientEvent::InteractWithBlock
let state = match block_kind {
    BlockKind::Torch => face_wall_block(face, BlockState::TORCH, BlockState::WALL_TORCH),
    BlockKind::OakLog => face_directional_block(face, BlockState::OAK_LOG),
    ...
    kind => BlockState::from_kind(kind),
};

world.chunks.set_block_state(
    place_at,
    state,
);

// This will be outside impl Config
fn face_wall_block(face: BlockFace, normal: BlockState, wall: BlockState) -> BlockState {
    match face {
        BlockFace::Bottom => normal,
        BlockFace::Top => normal,
        BlockFace::North => wall.set(PropName::Facing, PropValue::North),
        BlockFace::South => wall.set(PropName::Facing, PropValue::South),
        BlockFace::West => wall.set(PropName::Facing, PropValue::West),
        BlockFace::East => wall.set(PropName::Facing, PropValue::East),
    }
}

fn face_directional_block(face: BlockFace, block: BlockState) -> BlockState {
    match face {
        BlockFace::Bottom | BlockFace::Top => block.set(PropName::Axis, PropValue::Y),
        BlockFace::North | BlockFace::South => block.set(PropName::Axis, PropValue::Z),
        BlockFace::West | BlockFace::East => block.set(PropName::Axis, PropValue::X),
    }
}
```

EmperialDev:
Fixes #115 
Make it so blocks are place right e.g torches stay on walls and keep their rotation, logs keep their axis and banners their rotation when updated.


examples/building.rs
```diff
@@ -211,10 +213,308 @@ impl Config for Game {
...
+                                        BlockKind::RedstoneTorch => face_wall_block(
+                                            face,
+                                            BlockState::REDSTONE_TORCH,
+                                            BlockState::REDSTONE_WALL_TORCH,
+                                        ),
+                                        BlockKind::SoulTorch => face_wall_block(
+                                            face,
+                                            BlockState::SOUL_TORCH,
+                                            BlockState::SOUL_WALL_TORCH,
+                                        ),
```
dyc3, line 232: There's got to be a better way to do this than having this massive match right? This is an example, they're supposed to be simple enough for someone to know what they do at a glance. Do we have extracted metadata for this? Can we provide it as a utility in valence?

examples/building.rs
```diff
@@ -230,3 +530,22 @@ impl Config for Game {
...
+    }
+}
+
+fn face_directional_block(face: BlockFace, block: BlockState) -> BlockState {
+    match face {
+        BlockFace::Bottom | BlockFace::Top => block.set(PropName::Axis, PropValue::Y),
+        BlockFace::North | BlockFace::South => block.set(PropName::Axis, PropValue::Z),
+        BlockFace::West | BlockFace::East => block.set(PropName::Axis, PropValue::X),
+    }
+}
```
dyc3, line 551: I think these should definitely go in the library itself, with unit tests.

examples/building.rs
EmperialDev: Yeah, its getting pretty long, and there is still missing blocks. I can't seem to find any metadata on the normal <-> wall variants, but there should be something somewhere in the java code. But I mabye found something inside `blocks.java`:
```java
public static final Block PLAYER_HEAD = Blocks.register("player_head", new PlayerSkullBlock(AbstractBlock.Settings.of(Material.DECORATION).strength(1.0f)));
public static final Block PLAYER_WALL_HEAD = Blocks.register("player_wall_head", new WallPlayerSkullBlock(AbstractBlock.Settings.of(Material.DECORATION).strength(1.0f).dropsLike(PLAYER_HEAD)));
```
The important part is the `.dropsLike(PLAYER_HEAD)`, this shows atleast that they are connected, so this could mabye be used?

I thought about moving it into `build/blocks.rs` or should it have it's own file inside Valence?

examples/building.rs
EmperialDev: Yeah, but where? Would it be best to put it inside `util.rs`? Or would it be better to put it inside `block.rs`?

examples/building.rs
EmperialDev: I have found the place where the metadata is, it's stored inside `item.java`, e.g:
```java
public static final Item TORCH = Items.register(new WallStandingBlockItem(Blocks.TORCH, Blocks.WALL_TORCH, new Item.Settings().group(ItemGroup.DECORATIONS)));
```
Here the constructor takes in two blocks, the `standingBlock` and `wallBlock` so I just need to extract these two (It's exactly identical for the other standing <-> wall blocks).

examples/building.rs
dyc3: Lets put it in block.rs for now, we can always move it later.

EmperialDev:
I han't touch the shape of stairs nor the connecting of wall blocks etc.

Is `face_block` too big and complicated? Should I try to cut it down in features and make it more simple?

As said in  #129 there should be a dedicated function for placing blocks, and most of this should properly be moved into that funtion.

rj00a:
I discovered in #74 that the block module is not performing as it should. This might require rethinking the entire block module, so I think we should hold off on these changes until that is resolved.

EmperialDev:
That's okay, it was also just a QoL for people trying to build something in the building example.

JoelEllis:
Hey - I did a similar thing, and I only noticed this when I went and looked through the PRs! :laughing: 

I implemented top/bottom for stairs & slabs, and the facing attribute for everything that has a direction - doors, stairs, etc.

https://github.com/JoelEllis/plotsirv/commit/3c77daf32af2cd9240e24e9a81ce1a6d30a31d59

It was simpler than I expected to be honest, although I would like an API that gives the facing cardinal direction of a client (like in the f3 menu) as that is pretty much directly set on the block. 

I haven't tested mine with wall blocks at all, though. Looking through this, it's likely that the `facing` logic is different for them. :disappointed: 

If it's all right with everyone here, I'd like to try and create a new version of this PR against the latest version of Valence, with the experience of implementing & testing a more complete creative server implementation in my own repository.

dyc3:
@JoelEllis Sure, go for it. This PR is very out of date and probably needs to be rewritten anyway. You can probably cherry-pick [d13be50](https://github.com/valence-rs/valence/pull/117/commits/d13be506f943f134f00475982b6e7d23627a22f5) into your new branch.

PR #118, opened 2022-10-16
add `slot_mut` to `Inventory`

dyc3:
- add doc comment to item stack
- add `slot_mut`

### Test Plan

- unit tests
also
- run `chest` example
- sneak to get stone in first hotbar slot

### DIFF
--- a/examples/chest.rs
+++ b/examples/chest.rs
@@ -259,11 +259,17 @@ impl Config for Game {
                     }
                     ClientEvent::StartSneaking => {
                         let slot_id: SlotId = PlayerInventory::HOTBAR_SLOTS.start;
-                        let stack = match client.inventory.slot(slot_id) {
-                            None => ItemStack::new(ItemKind::Stone, 1, None),
-                            Some(s) => ItemStack::new(s.item, s.count() + 1, None),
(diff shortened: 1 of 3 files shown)


rj00a (commented):
So `slot_mut` is forced to mark the slot as dirty. My only concern is that `set_slot` has the opportunity to avoid marking the slot as dirty when the new slot is equal to the old slot. But `slot_mut` can't have this ability.

So it might lead to a situation where someone is acquiring the slot mutably but doesn't actually end up modifying the slot. Then they would have to pay the cost of sending a packet anyway. Pseudocode:

```rust
if let Some(item) = inv.slot_mut(42) {
    if item.has_some_property_we_care_about() {
        item.mutate();
    } else {
        // Slot is dirty!
    }
} else {
    // Slot is dirty!
}
```

Avoiding a packet is probably more important than avoiding a clone.

rj00a:
Entities' tracked data don't have a `get_mut` for any of their fields for this reason.

dyc3:
That is true. I might be able to do something about that with a guard type around the slot. I'll play around with it.

dyc3:
After playing around with it for a little too long, this is too complicated.

PR #123, opened 2022-10-18
World border

Sekky61:
# World Border

This PR implements packets and methods to do with world borders.
World border is a wall defined by center point and diameter. It restricts client from going outside the border.
## Implemented packets
 - `InitializeWorldBorder` (this one was defined as `WorldBorderInitialize`, but unused)
 - `SetBorderCenter`
 - `SetBorderLerpSize`
 - `SetBorderSize`
 - `SetBorderWarningDelay`
 - `SetBorderWarningDistance`
## TODOs
I do not yet understand the behavior of `portal_teleport_boundary` field on `InitializeWorldBorder`. It seems to affect the shape of the border, sometimes making it rectangular instead of square, or making it disappear entirely. 

I could not replicate the appearance of the warning screen when getting close to the border.
## Discussion
Take this as a draft of the API. I considered two options, basic methods on `Client` for basic usecases (current state of PR) and copying packet API to the `Client` API.
Is it acceptable to allow unusual usecases with manual `send_packet` only?

### DIFF
--- a/examples/border.rs
+++ b/examples/border.rs
@@ -1,0 +1,233 @@
+use std::net::SocketAddr;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use log::LevelFilter;
+use num::Integer;
+use valence::biome::Biome;
+use valence::block::BlockState;
+use valence::chunk::UnloadedChunk;
+use valence::client::{handle_event_default, ClientEvent};
+use valence::config::{Config, ServerListPing};
+use valence::dimension::{Dimension, DimensionId};
+use valence::entity::{EntityId, EntityKind};
+use valence::player_list::PlayerListId;
+use valence::protocol::packets::s2c::play::{
+    InitializeWorldBorder, SetBorderCenter, SetBorderLerpSize, SetBorderSize,
+};
+use valence::server::{Server, SharedServer, ShutdownResult};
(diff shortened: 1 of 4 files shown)


dyc3:
> Is it acceptable to allow unusual usecases with manual `send_packet` only?

`send_packet` is publicly accessible _just in case_ we didn't account for a specific use case. It's a courtesy. That being said, I see no reason why this API wouldn't cover all the use cases as long as you expand it so that all the different kinds of packets can be sent without calling `send_packet` explicitly.

Are there any particular weird use cases you had in mind?

rj00a:
> Is it acceptable to allow unusual usecases with manual `send_packet` only?

If it's something obscure or behavior not really intended by Mojang then I think it's fine to require a manual `send_packet`.

CodeDoctorDE:
Any updates? This would be a nice feature for nice game modes

dyc3:
This PR is based on a really old version of valence. A lot of it will have to be reworked. @CodeDoctorDE, feel free to make a new PR for this.

CodeDoctorDE:
Hmm, I'm very new to rust and I'm not an expert with the minecraft protocol

rj00a:
World border was implemented in #364.

PR #124, opened 2022-10-19
Fix Inventory `consume_one`

guac420:
Since `ItemStack::set_count()` is bound to 1-127 inclusive it is impossible to set the count to 0. Previously consuming one from a stack of one would do nothing. Now the count is only set if there are more than one item in the stack.


guac420:
Just noticed that #118 resolves this issue. I'm not sure if it would be worth merging this if #118 is going to fix it soon anyways. If so I could just close this PR and wait for #118.

rj00a:
Yeah a lot of the inventory stuff including this function is temporary so I wouldn't worry about it.

Issue #114, opened 2022-10-16
Block desync in building example


rj00a:
Quickly breaking and placing blocks causes ghost blocks. This video should explain it.

https://user-images.githubusercontent.com/31678482/196025723-ebe5c760-ea31-43b0-9f97-efa95d562b3e.mp4

PR #127, opened 2022-10-20
Block sequence isn't getting synced when placing blocks

EmperialDev:
The `block_change_sequence` wasn't getting updated when placing blocks only when breaking blocks, this results in ghost block states on the client. Ghost block states is when you place a block e.g a torch, and then when you break a block, the torch changes block state, e.g loses it's facing. This also means that blocks are not being placed right, but I hope to fix this with #117.

I also added `cmp::max(...)`, because if a modded client sent a malicious packet, where the sequence is shorter then the previous they sent, thay could mabye revert block placement/breaking they previously had done?

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -2,8 +2,8 @@
 
 use std::collections::{HashSet, VecDeque};
 use std::iter::FusedIterator;
-use std::mem;
 use std::time::Duration;
+use std::{cmp, mem};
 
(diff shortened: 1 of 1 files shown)


dyc3:
Does this fix #114?

rj00a:
> Does this fix #114?

Don't think so.

EmperialDev:
It is more like ghost block states

PR #132, opened 2022-10-22
Add `Username<S>` type

rj00a:
This is analogous to the existing `Ident<S>` type.

### DIFF
--- a/src/username.rs
+++ b/src/username.rs
@@ -5,7 +5,8 @@ use std::fmt::Formatter;
 use std::io::Write;
 use std::str::FromStr;
 
-use serde::{Deserialize, Serialize};
+use serde::de::Error as _;
+use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::protocol::{Decode, Encode};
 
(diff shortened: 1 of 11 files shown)


src/username.rs
```diff
@@ -0,0 +1,174 @@
...
+/// ```
+/// use valence::prelude::*;
+///
+/// assert!(Username::new("00a").is_ok());
+/// assert!(Username::new("jeb_").is_ok());
+///
+/// assert!(Username::new("notavalidusername").is_err());
+/// assert!(Username::new("NotValid!").is_err());
+/// ```
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Serialize, Deserialize)]
```
rj00a, line 38: Derived `Deserialize` impl isn't calling `new`.

dyc3:
What exactly will this do for us?

rj00a:
> What exactly will this do for us?

There are a number of places where usernames are passed around. Using this type ensures that the contained string is actually a valid username and not some other kind of string. For instance you can use it as a function argument to indicate that only valid usernames are accepted, or return it from a function to indicate that only valid usernames are produced.

See: https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html

dyc3 (approved):

dyc3 (approved):

PR #137, opened 2022-10-28
Add SkyGrid example

UStuej:
This example generates terrain similar to the SkyGrid map by SethBling:  https://www.planetminecraft.com/project/skygrid-survival-map/ .

`BLOCK_TYPES: [BlockState, 547]` is intended to be an array of all `BlockState`s listed as opaque in `out/blocks.rs`.

(Much of `skygrid.rs` is copied from `terrain.rs`.)


examples/skygrid.rs
```diff
@@ -0,0 +1,748 @@
...
+
+fn has_terrain_at(x: i64, y: i64, z: i64) -> bool {
+    Y_RANGE.min().unwrap_or(-64) <= y
+        && y <= Y_RANGE.max().unwrap_or(319)
+        && x % BLOCK_SPACING == 0
+        && y % BLOCK_SPACING == 0
+        && z % BLOCK_SPACING == 0
+}
+
+const BLOCK_TYPES: [BlockState; 547] = [
```
rj00a, line 200: Any chance you can get rid of this giant array? You might even be able to use `const` evaluation to compute the array at compile time, but I don't know if Rust is capable of that yet. Otherwise, it would be better to create a `Vec<BlockState>` at runtime.

UStuej: I generated that array in Python using information from `out/blocks.rs`.  I don't know how to generate the array at runtime.

rj00a: Something like this should do it:
```rust
let blocks: Vec<BlockState> = BlockKind::ALL.into_iter().map(|k| k.to_state()).filter(|b| b.is_opaque()).collect();
```

Although I think using `is_solid` would be a better heuristic to use here since this currently excludes blocks such as glass. However, Valence has not extracted this information yet, so there is no `is_solid`.

UStuej: Ok, thank you.  (It wasn't clear to me that those methods were available.)

UStuej: > ```rust
> BlockKind::ALL.into_iter().map(|k| k.to_state()).filter(|b| b.is_opaque()).collect();
> ```

This doesn't appear to be able to evaluate within a constant context, as `[BlockKind; 933]` apparently does not implement `~const IntoIterator`, so I don't know of a way to store the result in a `const` or `static` variable outside of using a crate like `once_cell`.  The alternative is to not store the value in a static way, but then I don't know where else to store the result (for use by the function `terrain_column`).

UStuej:
I have committed working code which replaces the sizable `BLOCK_TYPES: [BlockState; 547]` array, using `once_cell::sync::Lazy` and the one-liner @rj00a provided to generate a vector at runtime of `BlockStates`s which satisfy `is_opaque`.

examples/skygrid.rs
UStuej: (This issue has been resolved with the `once_cell` library.)

UStuej:
@dyc3 You can try lowering your render distance.  The example causes my client to freeze at first.

I tried to imitate how the other examples (specifically `terrain.rs`) generate chunks and assign `BlockType`s within them, but I may not have wrote the chunk generation as efficiently as I could.  I haven't seen documentation on correctness for performing higher-level interactions with the Valence API.

UStuej:
I pulled the latest commits from `valence-rs/valence/main`, then later undid that action.  That's what the force-push above this message is for; it should have reverted my action.

UStuej:
(And now I decided to redo that action as I believe it might easier to merge, hopefully.)

dyc3 (approved):
I lowered my render distance to 4 and was able to get it to work. Previously I was using the default render distance 12.

We should try to figure out why the client is freezing though. But that seems out of scope of this PR.

UStuej:
Great!

rj00a:
Now that the ECS rewrite is complete, I wouldn't mind giving this another go.

PR #138, opened 2022-10-29
feature-website

kubo6472:
Add a basic website using docusaurus


kubo6472:
It won't build at mine, since you need to explicitly set GH org and repo name for it to work. There is a docusaurus-friendly way of doing the deploy, and I will look into it, but for now, this should work.

kubo6472:
Okay, I took my time and tested their recommended way of deploying the website. The CI is prefilled to work on THIS REPO (valence-rs/valence) and YOUR DOMAIN (https://valence.rs) so you should point a CNAME from WWW and @ (root) to valence-rs.github.io now, or before merging. Also enable github pages from the gh-pages branch in repo settings. If you want, [this is what I came up with.](https://d.kalr.tech) And [here's the source.](https://github.com/kalrtech/valence)

kubo6472:
It's in settings->pages, you don't need to set the custom domain there, it will be automatically set by this workflow. Just let it run and when you have a gh-pages branch, set it, and /(root) as the source.
image

.github/workflows/website.yml
```diff
@@ -0,0 +1,42 @@
...
+  deployments: write
+  id-token: write
+  issues: write
+  discussions: write
+  packages: write
+  pages: write
+  pull-requests: write
+  repository-projects: write
+  security-events: write
+  statuses: write
```
dyc3, line 19: Do we need _all_ of these permissions?

website/sandbox.config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "infiniteLoopProtection": true,
+  "hardReloadOnChange": true,
+  "view": "browser",
+  "template": "docusaurus",
+  "node": "16",
```
dyc3, line 6: we should probably also set the `engines` field in `package.json`

.github/workflows/website.yml
kubo6472: Not really, I was just frustrated with the build failing due to not having the right permissions.

website/src/pages/index.js
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <header className={clsx('hero hero--primary', styles.heroBanner)}>
+      <div className="container">
+        <h1 className="hero__title">{siteConfig.title}</h1>
+        <p className="hero__subtitle">{siteConfig.tagline}</p>
+        <div className={styles.buttons}>
+          <Link
+            className="button button--secondary button--lg"
+            to="/docs/Getting Started">
+            Get started with Valence - 5min ⏱️
```
qualterz, line 21: Should the brand name mention be avoided? I think a message like `Get started in 5 minutes ⏱️` would look much better.

website/sandbox.config.json
kubo6472: https://github.com/valence-rs/valence/blob/537ebbb71ddf4451961e136aaa98c9985cd4ea26/website/package.json#L42
it is set already, if this is what you mean

website/docs/Getting Started.md
```diff
@@ -0,0 +1,101 @@
...
+In the future we may decide to reimplement vanilla game mechanics as a separate project. If you're developing something
+like a minigame server without need for vanilla game mechanics, you can depend on Valence directly.
+
+# Goals
+
+Valence aims to be the following:
+
+* **Complete**. Abstractions for the full breadth of the Minecraft protocol.
+* **Flexible**. Your use case should be achievable without manually sending and receiving packets or other hacks.
+* **Minimal**. The API surface is small with only the necessities exposed. Opinionated features such as a
```
qualterz, line 21: From here the description about unopinionatedness can be moved into a separated goal like `Unopinionated`.

.github/workflows/website.yml
kubo6472: it works, with less...

website/docs/Getting Started.md
kubo6472: Sure, this wasn't too much focused on the contents, more on the system. Feel free to send PRs to the fork, if you have some exact ideas. Thanks.

kubo6472:
Closing in favor of #382 
I'm sad nobody mentioned their initiative to create something else instead.

PR #139, opened 2022-10-29
Fix yaw not matching client yaw

jivvy:
Saw this bug and thought it might be a good first issue.

I'm new to rust and contributing in general, so please guide me.

This seemed to fix the issue, but it might not apply to all cases, only for C2S packets regarding pos & yaw and entities

I'm also not sure what the best way of applying the wrap to when you set the yaw, though.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -46,7 +46,7 @@ use crate::server::{C2sPacketChannels, NewClientData, S2cPlayMessage, SharedServ
 use crate::slab_versioned::{Key, VersionedSlab};
 use crate::text::Text;
 use crate::username::Username;
-use crate::util::{chunks_in_view_distance, is_chunk_in_view_distance};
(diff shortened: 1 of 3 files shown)


jivvy:
I put a print statement in where the C2S packets are handled:
```rust
C2sPlayPacket::SetPlayerPositionAndRotation(p) => {
    println!("yaw: {}", p.yaw);
    if self.pending_teleports == 0 {
        self.position = p.position;
        self.yaw = wrap_yaw(p.yaw);
        self.pitch = p.pitch;

        self.events.push_back(ClientEvent::MovePositionAndRotation {
            position: p.position,
            yaw: p.yaw,
            pitch: p.pitch,
            on_ground: p.on_ground,
        });
    }
}
```
And interpreted this as that the client is actually sending unwrapped yaw values to the server

Issue #121, opened 2022-10-18
Entity yaw (mabye pitch) keeps going up


EmperialDev:
When you turn around your yaw (and properly also pitch) keeps goining up, so when you print out the player's yaw it is above 180 and if you turn the other way, you can get below -180.

Valence:
![Valence](https://user-images.githubusercontent.com/98261860/196517493-16e80ee9-1427-47f8-81f0-fdf163432ba1.png)
Vanilla:
![Vanilla](https://user-images.githubusercontent.com/98261860/196517495-a72e8fe9-6643-4eec-a507-ea21e9326698.png)

As you can see, this makes it so that the yaw (and mabye pitch) in the f3 doesn't match `client.yaw()` and `player.yaw()`.

Test plan
-
Print out the yaw and compare it to the f3 screen.

jivvy:
@EmperialDev Is it important to make the yaw match what is shown on the client?

Not sure what kind of application you're using it for.

Although the yaw is not the same number, when completing revolutions around a circle with it, it should still point in the same direction.

Does this need to be fixed?

EmperialDev:
I just discovered that the yaw kept going up, and thought that it was a bug, since when I printed out the value with a fabric mod, the yaw matched. But that is only if you log the yaw from the client, the server has the non capped yaw aswell on fabric.

But for what I use it for, it actually doesn't matter since the equation still works.

The only thing I can think of is the precision of the `f32` over time? Or what about if you want to teleport a player to another players yaw and pitch, than they will have the non capped yaw, is that okay to just send that, or does it need to be capped, I don't know? Another thing could be if someone else uses it, and just makes an if statment like this:
```rust 
if player.yaw > 90 && player.yaw < 180 {...}
```
But a simple comment would properly fix that.

So mabye not that important sry.

rj00a:
Closing this for now.

PR #140, opened 2022-10-29
Fix yaw not matching client yaw

jivvy:
Saw this bug ( #121 ) and thought it might be a good first issue.

I'm new to rust and contributing in general, so please guide me.

This seemed to fix the issue, but it might not apply to all cases, only for C2S packets regarding pos & yaw and entities

I'm also not sure what the best way of applying the wrap to when you set the yaw, though.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -111,7 +111,7 @@ pub fn wrap_yaw(yaw: f32) -> f32 {
     let abs_yaw = yaw.abs();
     let steps = (abs_yaw / 180.0).floor();
     let step_offset = steps % 2.0;
-    return abs_yaw - (steps + step_offset) * 180.0;
+    abs_yaw - (steps + step_offset) * 180.0
 }
 
 /// Calculates the minimum number of bits needed to represent the integer `n`.
--- a/src/util.rs
+++ b/src/util.rs
@@ -105,6 +105,15 @@ pub fn ray_box_intersect(ro: Vec3<f64>, rd: Vec3<f64>, bb: Aabb<f64>) -> Option<
     }
 }
 
+/// Takes a float and wraps it around -180..180 in a way that matches the
+/// client's interpretation of yaw values
+pub fn wrap_yaw(yaw: f32) -> f32 {
+    let abs_yaw = yaw.abs();
+    let steps = (abs_yaw / 180.0).floor();
+    let step_offset = steps % 2.0;
+    return abs_yaw - (steps + step_offset) * 180.0;
+}
+
 /// Calculates the minimum number of bits needed to represent the integer `n`.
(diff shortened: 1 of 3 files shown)


jivvy:
I put a print statement in where the C2S packets are handled:
```rust
C2sPlayPacket::SetPlayerPositionAndRotation(p) => {
    println!("yaw: {}", p.yaw);
    if self.pending_teleports == 0 {
        self.position = p.position;
        self.yaw = wrap_yaw(p.yaw);
        self.pitch = p.pitch;

        self.events.push_back(ClientEvent::MovePositionAndRotation {
            position: p.position,
            yaw: p.yaw,
            pitch: p.pitch,
            on_ground: p.on_ground,
        });
    }
}
```
And interpreted this as that the client is actually sending unwrapped yaw values to the server

jivvy:
I put a print statement in where the C2S packets are handled:
```rust
C2sPlayPacket::SetPlayerPositionAndRotation(p) => {
    println!("yaw: {}", p.yaw);
    if self.pending_teleports == 0 {
        self.position = p.position;
        self.yaw = wrap_yaw(p.yaw);
        self.pitch = p.pitch;

        self.events.push_back(ClientEvent::MovePositionAndRotation {
            position: p.position,
            yaw: p.yaw,
            pitch: p.pitch,
            on_ground: p.on_ground,
        });
    }
}
```
And interpreted this as that the client is actually sending unwrapped yaw values to the server

src/util.rs
```diff
@@ -105,6 +105,15 @@ pub fn ray_box_intersect(ro: Vec3<f64>, rd: Vec3<f64>, bb: Aabb<f64>) -> Option<
     }
 }
 
+/// Takes a float and wraps it around -180..180 in a way that matches the
+/// client's interpretation of yaw values
+pub fn wrap_yaw(yaw: f32) -> f32 {
```
dyc3, line 110: Do we need to make this public?

jivvy: I'm not sure. 

The `Client` and `Entity` structs are public, so the `yaw` field they possess is also public.
If someone wanted to set the yaw, they would need the function.

Perhaps there's a better way of hiding `wrap_yaw` and apply it discreetly?

rj00a:
Hmm I'm not sure this is an issue that should be fixed. If anything, I would expect all angles to be in `[0, 360)` instead of whatever Minecraft is doing. And since these are the actual values sent by the client, I don't think we should attempt to apply our own transformations.

jivvy:
That's true. I'm not sure what kinds of applications would require a yaw fitting the vanilla client

rj00a:
Thanks for opening a PR. Closing for the reasons discussed.

PR #144, opened 2022-10-30
Implement missing text component types

### LINKED ISSUE #19: Some text content types are not yet implemented.

SandroHc:
Implements the remaining text component types. Should hopefully close #19.

Some pending points:

1. Clippy is still complaining about the `PlayerChatMessage` struct size being much bigger than the other variants (336 bytes vs 72 bytes). Should #18 be revisited?
2. What should be the plain representations for the non-standard types? Currently using something like "component_type_name(param1=val, param2=val, ...)"
3. What is the best way to initialize a non-standard type with optional parameters? I created one version with the required fields and one with the optional fields - e.g. "type(required)" and "type_with_optional(required, optional)" - works fine for types with a single optional parameter but falls apart with multiple parameters, like in the NBT type.
4. Is the [component resolution](https://minecraft.fandom.com/wiki/Raw_JSON_text_format#Component_resolution) (like `@p`) done on the server side? If so, where and how should we do it?

Output of the new text example:

Text example output


rj00a:
Thanks for working on this.
1. I've resolved the problem by adding a layer of indirection with `Box`. I considered using `Arc` with clone-on-write semantics instead but concluded there wasn't much benefit.
2. This seems reasonable to me. I've updated the name of the function to reflect this.
3. For the sake of simplicity I've consolidated the constructors where possible. However, the `*_nbt` functions are still separated. A variant was created for each of these functions in `TextComponent` which makes more invalid states unrepresentable and reduces the total size of `TextComponent` slightly. The fact that multiple NBT types can be present at once seems to be a quirk of Minecraft's deserializer, so it is not supported.
4. Component resolution is still a mystery to me. A quick glance at Minecraft's `Text` class didn't answer any questions.

I've removed some of the tests because I didn't think they were worth their weight (read: I was too lazy to update them). Sorry about that.

If there are no objections, I can merge this.

SandroHc:
Sounds weird to say it as the original contributor, but LGTM! :)

Just did a small change to the translate display to make it consistent with the others, and added basic tests to cover all the text contents to make sure they serialize correctly.

rj00a:
Great! Here's a follow-up issue:
- [ ] #158

PR #147, opened 2022-11-01
Better `consume` for inventories

dyc3:
- rework `consume_one` into `consume`
- update `consume_one_held_item` to `consume_held_item` to match `consume`
- remove an unnecessary clone

Originally, I was planning on removing this function altogether. After thinking about it for a bit though, I think it provides enough of a convenience for end users to warrant keeping it.

This also implements the fix that was originally in #124

### DIFF
--- a/src/inventory.rs
+++ b/src/inventory.rs
@@ -31,8 +31,11 @@ pub trait Inventory {
     /// and no changes were made to the inventory.
     fn consume(&mut self, slot_id: SlotId, amount: impl Into<u8>) -> Result<(), ()> {
         let amount: u8 = amount.into();
-        let mut slot = self.slot(slot_id).cloned();
-        if let Some(stack) = slot.as_mut() {
+        let slot = self.slot(slot_id).cloned();
+        if slot.is_some() {
+            // Intentionally not using `if let` so stack can be moved out of the slot to
+            // avoid another clone later.
+            let mut stack = slot.unwrap();
             if amount > stack.count() {
                 return Err(());
             }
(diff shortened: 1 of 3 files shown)


src/inventory.rs
```diff
@@ -251,6 +267,8 @@ impl Inventories {
     }
 }
 
+pub struct InventoryError;
```
rj00a, line 270: Should `#[derive(Error)]` from the thiserror crate.

src/inventory.rs
```diff
@@ -23,17 +23,33 @@ pub trait Inventory {
...
-            stack.set_count(stack.count() - 1);
-            let slot = if stack.count() == 0 {
+    /// Decreases the count for stack in the slot by amount. If there is not
+    /// enough items in the stack to perform the operation, then it will fail.
+    ///
+    /// Returns `Ok` if the stack had enough items, and the operation was
+    /// carried out. Otherwise, it returns `Err` if `amount > stack.count()`,
+    /// and no changes were made to the inventory.
+    #[allow(clippy::unnecessary_unwrap)]
+    fn consume(&mut self, slot_id: SlotId, amount: impl Into<u8>) -> Result<(), InventoryError> {
```
rj00a, line 33: I think the clones and unwraps can be avoided in this function.

rj00a (commented):
Thinking we should have a more general function. How about passing in a closure which is given the previous count and returns the new count?

src/inventory.rs
dyc3: The clone cannot be avoided because we need `ItemStack`, not `&ItemStack`. Additionally, we can't use `if let Some(stack) = slot.as_mut()` because `stack` becomes `&mut ItemStack`, and we need to give `set_slot` an `Option<ItemStack>`.

That being said, moving the implementation of this function to the inventory types themselves would not really have this restriction. I'm thinking of consolidating `PlayerInventory` and `ConfigurableInventory` in another PR

PR #150, opened 2022-11-05
Add Particles

jivvy:
PR to add particles in a new `particles` module

 - [x] All the particles except Item particle
 - [x] Item particle
 - [x] Particle example
 - [x] `Encode` trait
 - [x] `Decode` trait
 - [ ] Tests (Encode -> Decode round trip)


jivvy:
Unfortunately, wiki.vg is outdated
These are the discrepancies I've found so far:
Particles missing:
 - SonicBoom, id 24
 - SculkSoul, id 29
 - SculkCharge, id 30, extra data 'roll: float'
 - SculkChargePop, id 31

src/particle.rs
```diff
@@ -0,0 +1,431 @@
...
+    pub particle_type: ParticleType,
+    pub long_distance: bool,
+    pub position: Vec3<f64>,
+    pub offset: Vec3<f32>,
+    pub max_speed: f32,
+    pub particle_count: u32,
+}
+
+#[derive(Clone, Debug)]
+pub enum ParticleType {
```
dyc3, line 20: Would it be possible to add and extractor and make this enum with codegen instead?

jivvy: I was hoping this would be possible.
I don't know how the extractors work, but here is a sample of how particles are defined in MC source:
```java
        public VibrationParticleEffect read(ParticleType<VibrationParticleEffect> particleType, StringReader stringReader) throws CommandSyntaxException {
            stringReader.expect(' ');
            float f = (float)stringReader.readDouble();
            stringReader.expect(' ');
            float g = (float)stringReader.readDouble();
            stringReader.expect(' ');
            float h = (float)stringReader.readDouble();
            stringReader.expect(' ');
            int i = stringReader.readInt();
            BlockPos blockPos = new BlockPos(f, g, h);
            return new VibrationParticleEffect(new BlockPositionSource(blockPos), i);
        }
```

src/particle.rs
```diff
@@ -0,0 +1,578 @@
...
+    Flash,
+    HappyVillager,
+    Composter,
+    Heart,
+    InstantEffect,
+    Item(u32), // TODO: field is 'Slot': 'The item that will be used.'
+    VibrationBlock {
+        // The 'Block' variant of the 'Vibration' particle
+        block_pos: BlockPos,
+        ticks: VarInt,
```
rj00a, line 73: Should not expose `VarInt` directly. Use `i32` instead.

src/particle.rs
```diff
@@ -0,0 +1,578 @@
...
+    Crit,
+    DamageIndicator,
+    DragonBreath,
+    DrippingLava,
+    FallingLava,
+    LandingLava,
+    DrippingWater,
+    FallingWater,
+    Dust {
+        rgb: Vec3<f32>,
```
rj00a, line 36: RGB values should use vek's `Rgb` type instead.

src/particle.rs
```diff
@@ -0,0 +1,578 @@
...
+    fn decode(_r: &mut &[u8]) -> anyhow::Result<Self> {
+        todo!("Is this even necessary?");
+        // let particle_id: VarInt = Decode::decode(_r).context(concat!("failed
+        // to read field `", stringify!(particle_id), "` from struct `",
+        // stringify!(Particle), "`"))?;
+    }
+}
+
+impl Encode for ParticleType {
+    fn encode(&self, _w: &mut impl Write) -> anyhow::Result<()> {
```
rj00a, line 407: `concat!` and `stringify!` macros can be removed.

src/particle.rs
```diff
@@ -0,0 +1,578 @@
...
+        let data_len = match self {
+            ParticleType::Block(block_state) => block_state.encoded_len(),
+            ParticleType::BlockMarker(block_state) => block_state.encoded_len(),
+            ParticleType::Dust { .. } => 4 * 4,
+            ParticleType::DustColorTransition { .. } => 7 * 4,
+            ParticleType::FallingDust(block_state) => block_state.encoded_len(),
+            ParticleType::SculkCharge { .. } => 4,
+            ParticleType::Item(_) => todo!("Item particle not yet implemented"),
+            ParticleType::VibrationBlock { block_pos, ticks } => {
+                "block".to_string().encoded_len() + block_pos.encoded_len() + ticks.encoded_len()
```
rj00a, line 562: No need to allocate new strings with `to_string`. `&str` should suffice.

examples/particles.rs
```diff
@@ -0,0 +1,309 @@
...
+                        client.play_particle(particle_type.clone(), pos, offset, 0.1, 100, true);
+                    });
+                }
+            }
+            server.state.particle_index += 1;
+        }
+    }
+}
+
+fn add_all_particles(server: &mut Server<Game>) {
```
rj00a, line 187: Looks like you want to use the `vec!` macro in the `main` function to initialize the vector instead.

examples/particles.rs
```diff
@@ -0,0 +1,309 @@
...
+        });
+
+        if !server.clients.is_empty() && server.shared.current_tick() % 30 == 0 {
+            if server.state.particle_index == server.state.particle_list.len() {
+                server.state.particle_index = 0;
+            }
+            let pos = Vec3::new(0.0, 100.0, -10.0);
+            let offset = Vec3::new(0.5, 0.5, 0.5);
+            let opt: Option<&ParticleType> =
+                server.state.particle_list.get(server.state.particle_index);
```
rj00a, line 161: Since you're panicking in the `None` case anyway, it may be more idiomatic to use `.expect()` or just index the vec. Up to you.

src/particle.rs
```diff
@@ -0,0 +1,482 @@
...
+    pub particle_type: ParticleType,
+    pub long_distance: bool,
+    pub position: Vec3<f64>,
+    pub offset: Vec3<f32>,
+    pub max_speed: f32,
+    pub particle_count: u32,
+}
+
+#[derive(Clone, Debug)]
+pub enum ParticleType {
```
rj00a, line 20: I think `Particle` would be a better name for this. To resolve the ambiguity with the packet, the packet can be renamed to  `ParticleS2c`.

rj00a, line 20: This should be added to the `prelude` module.

jivvy: I agree. Was only hung up on the packet struct being named differently from other packet structs, but I guess it's fine

jivvy:
Concerning implementing the final particle, Item, [this is how the client reads it](https://wiki.vg/Slot_Data)

It seems to me that we will need to `impl Encode for ItemStack`?

rj00a:
There is a `Slot` type alias that implements `Encode`.

williamdes:
TODO: update README ?

jivvy:
Anything else this PR needs to move forward?

Happy to do more

rj00a:
I'm working on some large architectural changes to the project at the moment. All of the protocol stuff including particles is going to go in a separate protocol library. Because of this, we will need a `Decode` implementation. Once that's done we can merge.

dyc3:
You could write round trip unit tests to test Encode -> Decode produces the same struct

src/particle.rs
```diff
@@ -0,0 +1,692 @@
...
+            85 => Particle::FallingDripstoneWater,
+            86 => Particle::GlowSquidInk,
+            87 => Particle::Glow,
+            88 => Particle::WaxOn,
+            89 => Particle::WaxOff,
+            90 => Particle::ElectricSpark,
+            91 => Particle::Scrape,
+            _ => {
+                bail!("Invalid particle id");
+            }
```
rj00a, line 680: Would be nice to include the invalid particle ID in the error message like
```suggestion
            id => bail!("invalid particle ID {id}"),
```
Error messages start with lowercase letters by convention.

rj00a (commented):
I'd go ahead and remove all the `.context()` calls since it's mostly noise.

And as @dyc3 said a unit test would be good.

rj00a:
> Also wondering if it would be better to define all the particle ids as constants so they won't be all spread out throughout the code

Since particle IDs are only used in this one place AFAICT it's probably not a problem. Otherwise we can write an extractor to create all the particle IDs for us.

rj00a:
The `valence_protocol` transition is now complete. If you could move `particle.rs` to the new `valence_protocol` crate that would be great. Also you'll have to replace the `vek` types with `[T; 3]`.

PR #152, opened 2022-11-08
Added ops implementations for BlockPos and ChunkPos

DanikVitek:
Applied `cargo clippy`

### DIFF
--- a/examples/text.rs
+++ b/examples/text.rs
@@ -1,0 +1,206 @@
+use std::net::SocketAddr;
+
+use valence::prelude::*;
+
+pub fn main() -> ShutdownResult {
+    tracing_subscriber::fmt().init();
+
+    valence::start_server(Game::default(), ServerState::default())
+}
+
+#[derive(Default)]
+struct Game {}
+
+#[derive(Default)]
+struct ClientState {
+    entity_id: EntityId,
+}
+
+#[derive(Default)]
+struct ServerState {
+    world: WorldId,
+}
+
+const FLOOR_Y: i32 = 1;
+const PLATFORM_X: i32 = 3;
+const PLATFORM_Z: i32 = 3;
+const SPAWN_POS: Vec3<f64> = Vec3::new(1.5, 2.0, 1.5);
+
+#[async_trait]
+impl Config for Game {
+    type ServerState = ServerState;
+    type ClientState = ClientState;
+    type EntityState = ();
+    type WorldState = ();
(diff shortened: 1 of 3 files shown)


dyc3 (approved):
Seems good, though I would love to see some unit tests.

DanikVitek:
> Seems good, though I would love to see some unit tests.

Ok, I'll add them

DanikVitek:
> Seems good, though I would love to see some unit tests.

Done

rj00a:
I don't think we should move forward with this. `BlockPos` is meant to represent a point in space rather than a displacement between two points (vector). Therefore the sum, multiplication, and division of two block positions is not meaningful. The operations that would make sense to me are subtraction between block positions returning a `Vec3<i32>` or addition/subtraction against a `Vec3<i32>` returning a new `BlockPos`. However, `BlockPos` is being moved to `valence_protocol` soon and the `vek` crate will not be available there.

Also I would like to comment on the use of `inline`. Often `#[inline]` is not necessary since the compiler is usually smart enough to know when something should be inlined or not. I would suggest only using `inline` if there is a _measurable_ performance increase with benchmarks to prove it. Otherwise we would end up sprinkling inlines haphazardly throughout the code base for little benefit.

DanikVitek:
Ok, the use of #[inline] is a topic for discussion. But as for being an API, I think presenting operations over such vectors is useful as it does not require intermediate conversions and can be easily used, for example, with dimensions (like 1 block in the Neather is 8 in the overworld). At least I don't think it is an odd feature. Except, probably, of the additional `safer` ops.

jivvy:
I think it's strange to have ops between two chunk / block pos's.

Side note, would it be weird to implement ops for integer tuples?

jivvy:
I was talking about something like
```rust
let other_pos: BlockPos = block_pos + (0, 2, 0);
```

DanikVitek:
> I don't think we should move forward with this. `BlockPos` is meant to represent a point in space rather than a displacement between two points (vector). Therefore the sum, multiplication, and division of two block positions is not meaningful. The operations that would make sense to me are subtraction between block positions returning a `Vec3<i32>` or addition/subtraction against a `Vec3<i32>` returning a new `BlockPos`. However, `BlockPos` is being moved to `valence_protocol` soon and the `vek` crate will not be available there.

So, should I reimplement the ops so that the return type is Vec3/Vec2 and clean up inlines, or just live it as is?

PR #160, opened 2022-11-23
Translation key extractor and code generator

### LINKED ISSUE #158: Provide translation keys in `valence_protocol`
An [extractor](https://github.com/valence-rs/valence/tree/main/extractor) and [code generator](https://github.com/valence-rs/valence/tree/main/valence_protocol/build) should be written to provide constants for all of Minecraft's translation keys. Each constant should have a doc comment providing the English translation of the key.

minecraft-data has an example of what the [generated JSON could look like](https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.19/language.json).

Example:

```rust
pub mod translation_key {
    /// Foo bar baz!
    pub const FOO_BAR_BAZ: &str = "foo.bar.baz";
    
    // ...
}
```

SandroHc:
Generates a new `translation_key.rs` with all bundled translations. Closes #158.

Decided to split the translation keys into tree files, as the generated code spans over 30K lines. The files are  [`translation_key_consts.rs`](https://paste.ee/r/y5YXz/0), [`translation_key_enum.rs`](https://paste.ee/r/y5YXz/1) and [`translation_key_enum_display.rs`](https://paste.ee/r/y5YXz/2).

Here's an example on how to use it:
```rust
// with enum
Text::translate(TranslationKey::NarratorJoining, [])

// with constant
Text::translate(TranslationKey::Custom(NARRATOR_JOINING), [])

// feeling adventurous
Text::translate(TranslationKey::Custom("custom.translation_key".to_string()), [])
```

### DIFF
--- a/examples/text.rs
+++ b/examples/text.rs
@@ -106,13 +106,20 @@ impl Config for Game {
                 client.send_message("\nTranslated Text");
                 client.send_message(
                     " - 'chat.type.advancement.task': ".into_text()
-                        + Text::translate("chat.type.advancement.task", []),
+                        + Text::translate(TranslationKey::ChatTypeAdvancementTask, []),
                 );
                 client.send_message(
                     " - 'chat.type.advancement.task' with slots: ".into_text()
                         + Text::translate(
-                            "chat.type.advancement.task",
+                            TranslationKey::ChatTypeAdvancementTask,
                             ["arg1".into(), "arg2".into()],
+                        ),
+                );
+                client.send_message(
+                    " - 'custom.translation_key': ".into_text()
+                        + Text::translate(
+                            TranslationKey::Custom("custom.translation_key".to_owned()),
+                            [],
                         ),
                 );
 
--- a/extractor/src/main/java/rs/valence/extractor/extractors/TranslationKeys.java
+++ b/extractor/src/main/java/rs/valence/extractor/extractors/TranslationKeys.java
@@ -1,0 +1,56 @@
+package rs.valence.extractor.extractors;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import net.minecraft.util.Language;
+import rs.valence.extractor.Main;
+
+import java.lang.reflect.Field;
+import java.util.Map;
+
+public class TranslationKeys implements Main.Extractor {
+
+    @Override
+    public String fileName() {
+        return "translation_keys.json";
+    }
+
+    @Override
+    public JsonElement extract() {
+        JsonArray translationsJson = new JsonArray();
+
+        Map<String, String> translations = extractTranslations();
+        for (var translation : translations.entrySet()) {
+            String translationKey = translation.getKey();
+            String translationValue = translation.getValue();
+
+            var translationJson = new JsonObject();
+            translationJson.addProperty("key", translationKey);
+            translationJson.addProperty("english_translation", translationValue);
+
+            translationsJson.add(translationJson);
+        }
+
+        return translationsJson;
+    }
+
+    @SuppressWarnings("unchecked")
+    private static Map<String, String> extractTranslations() {
+        Language language = Language.getInstance();
+
+        Class<? extends Language> anonymousClass = language.getClass();
+        for (Field field : anonymousClass.getDeclaredFields()) {
+            try {
+                Object fieldValue = field.get(language);
+                if (fieldValue instanceof Map<?, ?>) {
+                    return (Map<String, String>) fieldValue;
+                }
+            } catch (IllegalAccessException e) {
+                throw new RuntimeException("Failed reflection on field '" + field + "' on class '" + anonymousClass + "'", e);
+            }
+        }
+
+        throw new RuntimeException("Did not find anonymous map under 'net.minecraft.util.Language.create()'");
+    }
+}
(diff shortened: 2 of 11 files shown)


rj00a:
Should an enum be used here? Having a catch-all case (`Custom`) eliminates any benefit of using an enum IMO. At that point you may as well just use a string since they're essentially equivalent. A plain string would also be a bit more flexible as you can easily switch between `&str`, `String`, `Box<str>`, `Cow<'a, str>` etc.

extractor/src/main/java/rs/valence/extractor/extractors/TranslationKeys.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public String fileName() {
+        return "translation_keys.json";
+    }
+
+    @Override
+    public JsonElement extract() {
+        JsonArray translationsJson = new JsonArray();
+
+        Map<String, String> translations = extractTranslations();
```
dyc3, line 23: nit: prefer HashMap

valence_protocol/src/translation_key.rs
```diff
@@ -0,0 +1,3 @@
+include!(concat!(env!("OUT_DIR"), "/translation_key_consts.rs"));
```
dyc3, line 1: nit: I would prefer to have these outputs in a single file because we have IDEs that can fold code.

dyc3 (commented):
Yeah I'm not too sure about having the enum either. How much does it hurt compile times?

SandroHc:
> Should an enum be used here?

The idea behind the enum was to ensure type safety for the Mojang-provided translations, so any dev would know exactly what the expected values are. I added the catch-all variant for the hypothetical advanced user that wants to use translation keys provided by some other means, like resource packs. My expectation is that most devs would stick to the known values and ignore the catch-all variant.

> Yeah I'm not too sure about having the enum either. How much does it hurt compile times?

Profiled the build performance and, excluding dependencies, the `valence_protocol` crate build time saw a slowdown of 4.4s or 5% with the enum (took 100.3s to build with the enum and 95.9s without). The values provided are an average of three builds from my machine (i7-3770K, Windows 11 22000.1281) using `cargo clean && cargo build --package valence_protocol --timings`; raw results here: [timings.zip](https://github.com/valence-rs/valence/files/10084776/timings.zip).

> > Decided to split the translation keys into tree files, as the generated code spans over 30K lines.
>
> There are some existing generators that create much more than 30k so I wouldn't worry about it too much.

Originally split the generated code into smaller files due to my editor constantly freezing when opening the generated file; this is arguably caused by my poor choice of editor, but could happen to others too. @dyc3 seems to agree that a single file would suffice.

extractor/src/main/java/rs/valence/extractor/extractors/TranslationKeys.java
SandroHc: This specific map is initialized by the game with type `ImmutableMap<String, String>`, which can't be cast to `HashMap<String, String>`. Plus, we don't care about the implementation, just that it's a map of key-value pairs, so the generic type is good enough.

rj00a:
I'd argue that the `TranslationKey` type provides no additional benefit over a regular string for two reasons:

1. `TranslationKey` does not prevent the `Custom` variant from colliding with the other vanilla variants. `TranslationKey::ChatTypeAdvancementTask` and `TranslationKey::Custom("chat.type.advancement.task".into())` are semantically equivalent, but `TranslationKey` makes a distinction between them.

    Here's a situation I could see happening. Say you're given a `TranslationKey` and you want to know if its string representation is equal to `"chat.type.advancement.task"`. Some may feel inclined to write code that looks like this:
    ```rust
    let key: TranslationKey = ...;
    if matches!(key, TranslationKey::ChatTypeAdvancementTask | TranslationKey::Custom(k) if k == "chat.type.advancement.task") {
        // ...
    }
    ```

2. `TranslationKey::Custom` accepts invalid translation keys.  (Not to suggest that it has any way of knowing which are valid)

If you still want to give translation keys their own type, perhaps `TranslationKey` could look something like this instead:

```rust
#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
pub struct TranslationKey<S = String>(S);

// TODO: AsRef<str> impl, Borrow<str> impl, ...

impl<'a, S: From<&'a str>> TranslationKey<S> {
    pub fn chat_type_advancement_task() -> Self {
        Self("chat.type.advancement.task".into())
    }
    
    // ...

    pub fn custom(string: impl Into<S>) -> Self {
        Self(string.into())
    }
}
```

SandroHc:
I see your point. Equality matches and everything else related to the custom variant would be a mess to handle (and prone to errors), and there's no way around it if we want to be as open/unopiniated as possible. In that spirit, I've reverted the changes to the `TranslationKey`, leaving only the raw translation key strings. Here is the updated generated file: [`translation_key.rs`](https://github.com/valence-rs/valence/files/10097343/translation_key.zip).

examples/text.rs
```diff
@@ -1,6 +1,7 @@
 use std::net::SocketAddr;
 
 use valence::prelude::*;
+use valence_protocol::translation_key;
```
SandroHc, line 4: Should this be included in the prelude?

rj00a: I think so.

SandroHc: It was already in the prelude, but forgot to remove the import from the example.

PR #167, opened 2022-12-10
Use `Option<InventoryId>` for `open_inventory` on client

dyc3:
- automatically handle CloseContainer packet so the client updates open_inventory
- ~~remove `InventoryId::NULL` in favor of `Option<InventoryId>`~~ 
- Use `Option<InventoryId>` for `open_inventory` on client

I believe that doing stuff like `InventoryId::NULL` is an anti pattern, and better facilitated by using `Option` instead


rj00a:
I suggest replacing inventory's `VersionedSlab` with `RcSlab` to eliminate invalid keys entirely. Keys from versioned slabs are effectively weak pointers.

valence_protocol/src/packets/s2c.rs
```diff
@@ -185,6 +185,13 @@ pub mod play {
         pub reset: bool,
     }
 
+    #[derive(Copy, Clone, Debug, Encode, Decode, Packet)]
+    #[packet_id = 0x10]
+    pub struct CloseContainerS2c {
+        /// Ignored by notchian clients.
+        pub window_id: u8,
+    }
+
```
rj00a, line 194: Looks like this packet has not been added to the `S2cPlayPacket` enum at the bottom of the module.

rj00a (changes requested):

dyc3:
I'd like to defer the `RcSlab` change to another PR so I can better understand what exactly the side effects of that would be.

rj00a:
Inventory IDs are either valid or invalid. When an inventory (or entity, or world) is deleted, its corresponding ID becomes invalid. Attempting to do a lookup with an invalid key returns `None`. In this sense, an `InventoryId` can be thought of as an `Option<&mut Inventory>`.

Even if `InventoryId::NULL` isn't provided, creating a null value is trivial. Just create a new inventory and immediately remove it. The ID you get back is then invalid and can be used as a null pointer.

Continuing with the `Option<&mut Inventory>` analogy, what you are proposing to do with this PR is turn an `Option<&mut Inventory>` into an `Option<Option<&mut Inventory>>`. The `InventoryId` was already optional, so now you're adding an extra layer of optionality to the type.

`RcSlab` keys are refcounted, which means they are always valid. They are analogous to an `&mut Inventory`. My suggestion is to use RcSlab instead, which would make the `Option<InventoryId>` construction more reasonable. This is the reason I suggest doing this change in this PR.

dyc3:
Ok, I think I understand what you're saying, but I not sure that's the best way to frame the problem. 

Correct me if I wrong, but that setup could lead to a scenario where something is manipulating an inventory that is no longer supposed to exist, and we have no way to notify clients. For example:
- Client 1 opens a chest
- Client 2 breaks the chest
- Client 1 is now interacting with a chest that does not exist
- Client 1 is not notified of this change, because it still has a valid reference to the inventory via `open_inventory`
- Client 1 places item into chest and closes it
- The last reference to the inventory is gone, so valence cleans it up.
- The item no longer exists.

Really, the only place where a reference to an inventory becomes an `Option<Option<&mut Inventory>>` is `client.open_inventory`. But I still think this is the better way to model this information because both `Option`s convey different information:
- `Option<InventoryId>` conveys whether or not the client thinks that it has an inventory open.
- `Option<&mut Inventory>` conveys whether or not the desired inventory actually exists.

rj00a:
> Correct me if I wrong, but that setup could lead to a scenario where something is manipulating an inventory that is no longer supposed to exist, and we have no way to notify clients. For example:

  Yeah that's a good point. I think you'd need to maintain an external list of clients that are viewing the inventory and manually notify them that the inventory has been closed at the appropriate time. That seems pretty cumbersome.

  The details are still fuzzy, but I think we might need to keep the inventory alive somehow to synchronize with clients. We might be receiving packets for an inventory that doesn't exist anymore and we would need to translate slot indices to "undo" the operation on the client so that desync doesn't happen. Maybe the needed info could be stored inside the `InventoryId`? Not sure.

I still think we should keep `InventoryId::NULL` though. If we remove it, we should also remove the `Default` impl on `InventoryId` since `InventoryId::default() == InventoryId::NULL`. There would be an inconsistency with the other ID types too.

dyc3:
Ok, then lets keep `InventoryId::NULL`, but can we still keep the client's `open_inventory` as `Option<InventoryId>`?

rj00a:
Yeah sounds good.

rj00a (approved):

rj00a:
Oops, I think I missed my queue to merge this.

PR #168, opened 2022-12-10
add `swap_slot` to `Inventory`

dyc3:
This effectively exposes the slice `swap` method: https://doc.rust-lang.org/std/primitive.slice.html#method.swap

### DIFF
--- a/src/inventory.rs
+++ b/src/inventory.rs
@@ -126,6 +126,21 @@ impl<C: Config> Inventory<C> {
         mem::replace(old, new)
     }
 
+    pub fn swap_slot(&mut self, idx_a: u16, idx_b: u16) {
+        assert!(idx_a < self.slot_count(), "slot index out of range");
+        assert!(idx_b < self.slot_count(), "slot index out of range");
+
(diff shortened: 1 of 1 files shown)


src/inventory.rs
```diff
@@ -126,6 +126,21 @@ impl<C: Config> Inventory<C> {
         mem::replace(old, new)
     }
 
+    pub fn swap_slot(&mut self, idx_a: u16, idx_b: u16) {
+        assert!(idx_a < self.slot_count(), "slot index out of range");
+        assert!(idx_b < self.slot_count(), "slot index out of range");
+
+        if idx_a == idx_b {
```
rj00a, line 133: We can avoid marking the slots as modified and swapping if the contents of the slots at the indices are equal.

rj00a (changes requested):

src/inventory.rs
dyc3: Yes, that is what this does.

rj00a: I meant comparing the items themselves, not just the indices.

dyc3: oh the **contents** lmao. yeah I'll make that change

rj00a (approved):

PR #169, opened 2022-12-13
Re-visit `From` impl for native data types found in NBT data format.

TerminatorNL:
Now no longer returns `None` but `Result<>` with the old value if the type did not match the desired type.

Related: https://github.com/valence-rs/valence/pull/145#discussion_r1012527318


rj00a (commented):
- The `tag` module is private to the crate. I think the `NbtType` trait and the generic type parameter on `InvalidTypeError` should be removed to keep the interface as simple as possible.
- `InvalidTypeError` is missing a `std::error::Error` impl.
- If `TryFrom<Value> for T` exists, it would make sense to me that `TryFrom<&Value> for &T` and `TryFrom<&mut Value> for &mut T` should also exist.
- It is worth looking at [`serde_json::Value`](https://docs.rs/serde_json/latest/serde_json/enum.Value.html#). Notably, it does not have any trait impls to do `Value` to `T` conversions. The closest thing is the `as_*` methods.

To avoid the above problems, I think it would be best to just remove all the `TryFrom<Value> for T` and `From<Value> for Option<T>` impls on `Value` and recreate the needed conversion functions in `valence_anvil`. Or create an analogue of the `as_*` functions if that proves sufficient for your needs.

TerminatorNL:
Have a look at the latest change. I have ditched all generics and hard-coded the value conversions as seen in serde. It created the following methods:

`is_*(&self) -> bool`
`as_*(&self) -> Option<&*>`
`as_*_mut(&mut self) -> Option<&mut *>`
`take_*(&self) -> Option<*>`

IntelliJ IntelliSense works for all of the methods defined in the macro.

TerminatorNL:
Woops, I wanted to rename the branch. Guess not!

rj00a (commented):
I think the `take_*` functions should be renamed to `into_*` since "take" implies that the argument is not consumed. Other than that, looks good to me.

PR #176, opened 2022-12-22
reimpl chest example

dyc3:
- add `GENERAL_SLOTS`, `HOTBAR_SLOTS`, and `OFFHAND_SLOT` constants
- update the chest example so it compiles and runs
- add `PLAYER_INVENTORY_SLOTS` constant
- set dirty bit when replace_cursor_item is called
- update the default packet handler for ClickContainer to not crash when the player changes an open inventory

### Test plan

1. run `chest` example - `cargo r -r --example chest`
2. sneak to give yourself a stone block
3. open the chest (it's invisible on top of the stone block)
4. place stone in chest
5. see the stone moving around in the chest

### DIFF
--- a/src/client/event.rs
+++ b/src/client/event.rs
@@ -17,6 +17,7 @@ use valence_protocol::{BlockFace, BlockPos, Ident, ItemStack, VarLong};
 use crate::client::Client;
 use crate::config::Config;
 use crate::entity::{Entity, EntityEvent, TrackedData};
+use crate::inventory::PLAYER_INVENTORY_SLOTS;
 
 /// A discrete action performed by a client.
 ///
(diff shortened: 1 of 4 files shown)


src/client/event.rs
```diff
@@ -361,11 +362,20 @@ pub(super) fn next_event_fallible<C: Config>(
...
+                // we can only safely update the client's slots if there is no inventory open.
+                // Implementation-wise, this means that users will have to manually implement
+                // slot changes with an open inventory.
+                // TODO: It might be a good idea to just have users implement all of it, or maybe provide a helper.
+                if client.open_inventory().is_none() {
+                    for (idx, item) in &p.slots {
+                        if PLAYER_INVENTORY_SLOTS.contains(&((*idx) as u16)) {
+                            client.slots[*idx as usize] = item.clone();
+                        }
+                    }
```
dyc3, line 378: Looking for feedback on these lines specifically before I take the PR out of draft mode.

rj00a:
Guess we don't need this anymore.

PR #184, opened 2023-01-11
ECS Rewrite

### LINKED ISSUE #53: Inventory and Items

### LINKED ISSUE #69: Spawning entities triggered by client events is really tedious.
This is more of a usability problem, rather than a bug.

Supposed we have the following code:

```rust
let player = server.entities.get_mut(client.state.entity_id).unwrap();
while let Some(event) = handle_event_default(client, player) {
    match event {
        ...
    }
}
```

While inside this loop, it's not possible to call `server.entities.insert()` without getting this error:
```
error[E0499]: cannot borrow `*server.entities` as mutable more than once at a time
   --> examples/dropping.rs:211:39
    |
190 |             let player = server.entities.get_mut(client.state.entity_id).unwrap();
    |                          ----------------------------------------------- first mutable borrow occurs here
...
196 |             while let Some(event) = handle_event_default(client, player) {
    |                                                                  ------ first borrow later used here
...
211 |                         let (id, e) = server.entities.insert(EntityKind::Item, ());
    |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ second mutable borrow occurs here
```

It's technically possible to work around this, but its really annoying.

### LINKED ISSUE #179: Access to custom values in player's game profile
It would be wonderful if it were possible to access every value in the game profile provided by Mojang/Velocity, so that custom values can be passed to the sub-server without the need for an external messenger.

rj00a:
This PR redesigns Valence's architecture around the Bevy Entity Component System framework (`bevy_ecs` and `bevy_app`). Along the way, a large number of changes and improvements have been made.
- Valence is now a Bevy plugin. This allows Valence to integrate with the wider Bevy ecosystem.
- The `Config` trait has been replaced with the plugin struct which is much easier to configure. Async callbacks are grouped into their own trait.
- `World` has been renamed to `Instance` to avoid confusion with `bevy_ecs::world::World`.
- Entities, clients, player list, and inventories are all just ECS components/resources. There is no need for us to have our own generational arena/slotmap/etc for each one.
- Client events use Bevy's event system. Users can read events with the `EventReader` system parameter. This also means that events are dispatched at an earlier stage of the program where access to the full server is available. There is a special "event loop" stage which is used primarily to avoid the loss of ordering information between events.
- Chunks have been completely overhauled to be simpler and faster. The distinction between loaded and unloaded chunks has been mostly eliminated. The per-section bitset that tracked changes has been removed, which should further reduce memory usage. More operations on chunks are available such as removal and cloning.
- The full client's game profile is accessible rather than just the textures.
- Replaced `vek` with `glam` for parity with Bevy.
- Basic inventory support has been added.
- Various small changes to `valence_protocol`.
- New Examples
- The terrain and anvil examples are now fully asynchronous and will not block the main tick loop while chunks are loading.

# TODOs
- [x] Implement and dispatch client events.
- ~~[ ] Finish implementing the new entity/chunk update algorithm.~~ New approach ended up being slower. And also broken.
- [x] [Update rust-mc-bot to 1.19.3](https://github.com/Eoghanmc22/rust-mc-bot/pull/3).
- [x] Use rust-mc-bot to test for and fix any performance regressions. Revert to old entity/chunk update algorithm if the new one turns out to be slower for some reason.
- [x] Make inventories an ECS component.
- [x] Make player lists an ECS ~~component~~ resource.
- [x] Expose all properties of the client's game profile.
- [x] Update the examples.
- [x] Update `valence_anvil`.
- ~~[ ] Update `valence_spatial_index` to use `glam` instead of `vek`.~~ Maybe later
- [x] Make entity events use a bitset.
- [x] Update docs.

Closes #69
Closes #179
Closes #53


crates/valence_new/src/client.rs
```diff
@@ -553,7 +553,7 @@ fn update_one_client(
     if server.current_tick() % (server.tps() * 10) == 0 {
         if client.got_keepalive {
             let id = rand::random();
-            client.enc.write_packet(&KeepAliveS2c { id });
+            // client.enc.write_packet(&KeepAliveS2c { id });
```
dyc3: accident?

PR #187, opened 2023-01-12
[ECS] Add Inventory component, re-implement player inventories

dyc3:
- add missing field preventing build, format
- add Inventory ECS component
- add cursor_item fields, getters, setters to Client
- add new_with_title to Inventory instead of using an Option
- add Inventory component to client entities when they connect
- add system update_player_inventories

This does not implement opening other inventories yet, planning to do that in a follow up PR.

The idea here is to have an `Inventory` component that can be used for all inventories. If you want a chest to have an inventory, give the corresponding block entity an `Inventory`. If you don't want any inventory update systems to run, simply don't give clients a player inventory.

<details>

<summary>Code used to test</summary>

```rust
use bevy_ecs::prelude::*;
use bevy_ecs::schedule::ShouldRun;
use valence_new::client::Client;
use valence_new::config::Config;
use valence_new::dimension::DimensionId;
use valence_new::protocol::types::GameMode;
use valence_new::server::Server;

#[derive(Resource)]
struct GameState {
    instance: Entity,
}

fn main() -> anyhow::Result<()> {
    tracing_subscriber::fmt().init();

    valence_new::run_server(
        Config::default(),
        SystemStage::parallel()
            .with_system(setup.with_run_criteria(ShouldRun::once))
            .with_system(init_clients),
        (),
    )
}

fn setup(world: &mut World) {
    let instance = world
        .resource::<Server>()
        .new_instance(DimensionId::default());

    let id = world.spawn(instance).id();
    world.insert_resource(GameState { instance: id })
}

fn init_clients(mut clients: Query<&mut Client, Added<Client>>, state: Res<GameState>) {
    for mut client in &mut clients {
        client.set_instance(state.instance);
        client.set_game_mode(GameMode::Creative);
    }
}

```

</details>

### DIFF
--- a/crates/valence_protocol/src/inventory.rs
+++ b/crates/valence_protocol/src/inventory.rs
@@ -1,6 +1,65 @@
 use valence_derive::{Decode, Encode};
 
 #[derive(Copy, Clone, PartialEq, Eq, Debug, Encode, Decode)]
+pub enum WindowType {
+    Generic9x1,
+    Generic9x2,
+    Generic9x3,
+    Generic9x4,
+    Generic9x5,
+    Generic9x6,
+    Generic3x3,
+    Anvil,
+    Beacon,
+    BlastFurnace,
+    BrewingStand,
+    Crafting,
+    Enchantment,
+    Furnace,
+    Grindstone,
+    Hopper,
+    Lectern,
+    Loom,
+    Merchant,
+    ShulkerBox,
+    Smithing,
+    Smoker,
+    Cartography,
+    Stonecutter,
+}
+
+impl From<WindowType> for InventoryKind {
+    fn from(value: WindowType) -> Self {
+        match value {
+            WindowType::Generic9x1 => InventoryKind::Generic9x1,
+            WindowType::Generic9x2 => InventoryKind::Generic9x2,
+            WindowType::Generic9x3 => InventoryKind::Generic9x3,
+            WindowType::Generic9x4 => InventoryKind::Generic9x4,
+            WindowType::Generic9x5 => InventoryKind::Generic9x5,
+            WindowType::Generic9x6 => InventoryKind::Generic9x6,
+            WindowType::Generic3x3 => InventoryKind::Generic3x3,
+            WindowType::Anvil => InventoryKind::Anvil,
+            WindowType::Beacon => InventoryKind::Beacon,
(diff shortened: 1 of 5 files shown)


crates/valence_protocol/src/inventory.rs
```diff
@@ -26,6 +26,7 @@ pub enum InventoryKind {
     Smoker,
     Cartography,
     Stonecutter,
+    Player,
```
rj00a, line 29: This enum is supposed to match the [window type in the open inventory packet](https://wiki.vg/Inventory). By adding another variant it wouldn't match the packet anymore (Notice the `Encode` and `Decode` derives).

What should happen if a client "opens" a player inventory? Should they be shown a generic 9x3? 9x4 for the hotbar? Nothing at all? What happens if a client has an `Inventory` component added to it that isn't a player inventory (by removing the default one maybe)?  Should `update_player_inventories` check for this and skip over it?

It seems like the player inventory should be an inherent part of `Client` or perhaps a separate component.

dyc3: I _really_ don't like the idea of `Client` having slots or a special inventory that's not an `Inventory`. For one, it would make it impossible to "opt-out" of behavior that depends on it, like `update_player_inventories`, if a user decides they don't need it. Also, it just opens us up to implementing the same inventory code twice.

Adding multiple components of the same type is currently [not supported by bevy_ecs, and is still being debated](https://github.com/bevyengine/bevy/issues/1527).  We shouldn't worry about that until it actually becomes an issue.

Perhaps we should have a `WindowType` enum that matches what's supposed to go in the packet, and the `InventoryKind` enum as it currently is. I also think that for now, it would be best to just print a warning and/or return a `Result::Err` if a client tries to open a `Player` inventory. We can always change it later to open a 9x3 or 9x4 if we decide we want to do that.

rj00a: > Perhaps we should have a WindowType enum that matches what's supposed to go in the packet, and the InventoryKind enum as it currently is. I also think that for now, it would be best to just print a warning and/or return a Result::Err if a client tries to open a Player inventory. We can always change it later to open a 9x3 or 9x4 if we decide we want to do that.

Alright that sounds like the way to go.

PR #188, opened 2023-01-14
[ECS] re-implement opening inventories

dyc3:
This adds an `OpenInventory` component that indicates that an entity has opened and is viewing an inventory.

This doesn't yet implement actually allowing clients to modify the inventories. 

This makes it possible to query all entities for open inventories. We also don't have to manage a `_modified` flag, we can just use the tools bevy_ecs already provides. It also helps to avoid cluttering up the `Client` component more than it already is.
`window_id` is intensionally placed on `Client` because it's state that pertains to that specific connection, not the inventory that is currently open.

### Test Plan

1. Run the example.

```bash
cargo run --example inventory_test
```

2. Connect to the server.
3. An inventory will be shown to you when you finish joining.
4. Close the inventory to cycle to the next inventory.
5. All inventories should have a diamond pickaxe in slot 0.


crates/valence_new/src/inventory.rs
```diff
@@ -159,3 +162,108 @@ pub(crate) fn update_player_inventories(mut query: Query<(&mut Inventory, &mut C
...
+/// Used to indicate that the client with this component is currently viewing
+/// an inventory.
+#[derive(Debug, Clone, Component)]
+pub struct OpenInventory {
+    /// The Entity that has the Inventory that the client is currently
+    /// viewing.
+    pub(crate) entity: Entity,
+}
+
+impl OpenInventory {
```
rj00a, line 175: Should we add a getter for `entity`? Alternatively, make `entity` public?

crates/valence_new/src/inventory.rs
```diff
@@ -159,3 +162,108 @@ pub(crate) fn update_player_inventories(mut query: Query<(&mut Inventory, &mut C
         }
     }
 }
+
+/// Used to indicate that the client with this component is currently viewing
+/// an inventory.
+#[derive(Debug, Clone, Component)]
+pub struct OpenInventory {
+    /// The Entity that has the Inventory that the client is currently
+    /// viewing.
```
rj00a, line 171: How about phrasing it like this:
```suggestion
    /// The Entity with the `Inventory` component that the client is currently
    /// viewing.
```

crates/valence_new/src/inventory.rs
```diff
@@ -159,3 +162,108 @@ pub(crate) fn update_player_inventories(mut query: Query<(&mut Inventory, &mut C
...
+            continue;
+        }
+
+        // send the inventory to the client
+        let inventory = inventory.unwrap();
+        client.window_id = client.window_id % 100 + 1;
+
+        let packet = OpenScreen {
+            window_id: VarInt(client.window_id.into()),
+            window_type: VarInt(WindowType::from(inventory.kind) as i32),
```
rj00a, line 201: Oh, this field was supposed to be of type `WindowType` and not `VarInt`. While you're at it, I think `InventoryKind` should be moved to this inventory module because it doesn't really belong in `valence_protocol`.

crates/valence_new/src/inventory.rs
```diff
@@ -159,3 +162,108 @@ pub(crate) fn update_player_inventories(mut query: Query<(&mut Inventory, &mut C
...
+    for (mut client, open_inventory, _) in clients.iter_mut() {
+        // validate that the inventory exists
+        let inventory = inventories.get_component::<Inventory>(open_inventory.entity);
+        if inventory.is_err() {
+            warn!("Client is viewing an inventory that does not exist");
+            continue;
+        }
+
+        // send the inventory to the client
+        let inventory = inventory.unwrap();
```
rj00a, line 196: This can be better expressed with `let Ok(inventory) = ... else { ...; continue };` to avoid the `unwrap`.

crates/valence_new/src/inventory.rs
```diff
@@ -159,3 +162,108 @@ pub(crate) fn update_player_inventories(mut query: Query<(&mut Inventory, &mut C
...
+pub(crate) fn handle_close_container(
+    mut commands: Commands,
+    mut events: EventReader<CloseContainer>,
+) {
+    for event in events.iter() {
+        commands.entity(event.client).remove::<OpenInventory>();
+    }
+}
+
+/// Detects when an the client's `OpenInventory` component is removed, which
```
rj00a, line 257:
```suggestion
/// Detects when a client's `OpenInventory` component is removed, which
```

crates/valence_new/examples/inventory_test.rs
```diff
@@ -0,0 +1,85 @@
...
+        Inventory::new(InventoryKind::Crafting),
+    ];
+
+    for inv in inventories.iter_mut() {
+        inv.replace_slot(0, Some(ItemStack::new(ItemKind::DiamondPickaxe, 1, None)));
+    }
+
+    for inv in inventories {
+        world.spawn(inv);
+    }
```
rj00a, line 55: Combine these loops?

rj00a (commented):
My original plan was to cram everything into the `Client` component but I like what you've done here.

rj00a:
See the thread starting here: https://canary.discord.com/channels/998132822239870997/998132822864834613/1064057465118937099

crates/valence_new/src/inventory.rs
dyc3: I'll add a getter.

crates/valence_new/src/inventory.rs
dyc3: I actually need to rewrite most of this system because it doesn't account for multiple observers of the same inventory.

rj00a:
Note that we'll have to do something more complicated than a bitset to track changes when modifications by clients are supported because we don't want the client's changes to be sent back to them.

PR #195, opened 2023-01-23
Update Player List

rj00a:
The player list has been made into a resource rather than a component for the sake of simplicity. If we actually need to support multiple simultaneous player lists, we should be able to add that without too much hassle.

`cargo r -p valence_new --example player_list` for the example.


rj00a:
Something weird is going on with the ping. Sometimes it's a full bar for the joining player, but other times it's crossed out.

crates/valence_new/src/player_list.rs
```diff
@@ -0,0 +1,575 @@
...
+
+    let mut removed = vec![];
+
+    pl.entries.retain(|&uuid, entry| {
+        let Some(entry) = entry else {
+            removed.push(uuid);
+            return false
+        };
+
+        if entry.is_new {
```
rj00a, line 436: Forgot to set `is_new` to false.

crates/valence_new/src/player_list.rs
```diff
@@ -0,0 +1,575 @@
...
+    }
+
+    pub fn set_listed(&mut self, listed: bool) {
+        self.listed = listed;
+    }
+
+    fn clear_trackers(&mut self) {
+        self.old_game_mode = self.game_mode;
+        self.old_listed = self.listed;
+        self.is_new = true;
```
rj00a, line 325: Bruh

dyc3 (changes requested):
- Can you make the player list example accept offline connections by default? Makes it easier to test with 2 clients.
- The ordering of the "Hello!" entry sometimes appears in the first spot, and sometimes the second. Unable to reproduce reliably. Are we dependent on the client for this?
- Connecting with 2 clients using the same minecraft account in offline mode results in one of the entries for the clients having a blank name.

emortaldev:
>     * Sometimes when joining the example you fall through the floor. Unable to reproduce reliably.

sounds like client falling while chunks aren't loaded

rj00a:
> - Can you make the player list example accept offline connections by default? Makes it easier to test with 2 clients.
> - Connecting with 2 clients using the same minecraft account in offline mode results in one of the entries for the clients having a blank name.

Allowing a client to connect multiple times is going to cause problems generally because of duplicate UUIDs. Player list entries are keyed on the UUID, so connecting twice with the same account is going to overwrite the entry as expected. I would rather not set offline mode by default for this example because it should show the player skins when you press 
tab.

> - The ordering of the "Hello!" entry sometimes appears in the first spot, and sometimes the second. Unable to reproduce reliably. Are we dependent on the client for this?

I tried using an ordered collection (`IndexMap`) and the behavior was the same, so I'm assuming this is out of our control.

> - Sometimes when joining the example you fall through the floor. Unable to reproduce reliably.

I've fixed this by deferring teleports until after chunks are processed. Let me know if it's not actually fixed.

dyc3 (approved):
LGTM

Regarding the new systems, `add_new_clients_to_player_list` and `remove_disconnected_clients_from_player_list`, would it be possible to create something like a "feature bundle" to make it harder to forget one of these systems? This doesn't have to be fixed here, just brainstorming for next time.

rj00a:
Sounds like a good idea.

PR #196, opened 2023-01-25
Make Valence a Bevy Plugin

rj00a:
This PR adds `bevy_app` as a dependency and moves Valence's configuration and startup code into a Bevy `Plugin`.

# Benefits
- Enables us to fully integrate with the existing Bevy ecosystem. Think graphical frontends, physics engines (`bevy_rapier3d`), and who knows what else.
- Provides a standard interface for extending the functionality of Valence. (plugins!)
- An `App` is easier to use and configure than a plain `Schedule`.
- Users are now able to replace the default runner.


dyc3 (approved):
Seems good to me.

PR #197, opened 2023-01-26
Implement remaining clientbound packets

### LINKED ISSUE #186: Implement remaining clientbound packets in `valence_protocol`
Title. If done correctly, we should be able to use the [`packet_inspector`](https://github.com/valence-rs/valence/tree/main/crates/packet_inspector) proxy between vanilla client and vanilla server without any errors occurring. This will help us better understand the protocol and alert us to changes between protocol versions.

guac420:
Aims to close #186


rj00a:
Sorry for the trouble, but are you able to make this PR against the `ecs_rewrite` branch? `valence_protocol` is starting to diverge there.

rj00a:
If not then no worries.

rj00a:
Actually scratch that I'll make it work.

guac420:
If you need me to change it just let me know.

crates/valence_protocol/src/packets/s2c/look_at.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use crate::{Decode, DecodePacket, Encode, EncodePacket, VarInt};
+
+#[derive(Copy, Clone, PartialEq, Debug, EncodePacket, DecodePacket)]
+#[packet_id = 0x37]
+pub struct LookAt {
+    pub feet_eyes: FeetEyes,
+    pub target_position: [f64; 3],
+    pub entity_id: Option<VarInt>,
+    pub entity_feet_eyes: Option<FeetEyes>,
```
rj00a, line 11: Looks like you can bundle these two fields into an `Option<LookAtEntity>`. Then you wouldn't need a manual `Encode`/`Decode` impl and it would more accurately reflect the states this packet can be in.

```rust
struct LookAtEntity {
    pub entity_id: VarInt,
    pub entity_feet_eyes: FeetEyes
}
```

crates/valence_protocol/src/packets/s2c/look_at.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Encode, Decode)]
+pub enum FeetEyes {
```
rj00a, line 15: Maybe `FeetOrEyes` would be a better name?

crates/valence_protocol/src/packets/s2c.rs
```diff
@@ -608,6 +608,10 @@ pub mod play {
         pub blocks: &'a [VarLong],
     }
 
+    #[derive(Clone, Debug, Encode, EncodePacket, Decode, DecodePacket)]
+    #[packet_id = 0x40]
+    pub struct SelectAdvancementsTab<'a>(Option<Ident<&'a str>>);
```
rj00a, line 619: The inner value is not public. (As a side note, I'm probably going to change all the tuple structs into structs with named fields for consistency.)

guac420: I can change all of them real quick if you would like.

rj00a: That'd be great.

rj00a:
Nice!

rj00a:
Thanks for this. This was a huge help. If you'd like to keep working on `valence_protocol`, there is #209 now.

PR #198, opened 2023-01-26
[ECS] Infra for comprehensive unit tests

dyc3:
Motivation: There are a lot of edge cases we have to handle, and every new line of code we write increases the risk of regressions. Regressions slow down dev time: a perfect example of this is what happened in #189.

Unsolved issues:
- [x] `test_client_position` is flakey

### Test plan

```
cargo test -p valence_new
```


rj00a (commented):
If we are to move forward with my proposal in #199, here is what I think should happen (in no particular order):
- Move all of Valence's tokio code and initial login procedures into a separate crate, let's call it `valence_network`. `valence` should no longer have a dependency on tokio. It is the responsibility of `valence_network` to spawn `Client` entities as connections are made to the server. `valence_network` exposes a Bevy `Plugin`.
- The `OwnedSemaphorePermit` that is stored in `Client` can be moved into a separate component defined in `valence_network`.
- `Client`'s constructor now looks like this:
  ```rust
  pub fn new(info: NewClientInfo, stream: Box<dyn PacketStream>) -> Self { ... }
  ```
- `PacketStream` is a new trait that has `try_send` and `try_recv` methods similar to byte channels.

With those things done, setting up test infra should be easy. How much of that you want to address here or elsewhere is up to you.

dyc3:
That sounds good to me. It's going to be most beneficial if we have good test infra before we start writing tests, so I'm going to work on that here before we merge it in.

These are mostly notes for me:
```mermaid
sequenceDiagram
    actor Alice
    participant socket as TCP Socket
    participant stream as PacketStream
    Alice->>socket: receive bytes
    socket->>stream: buffer_recv
    Client->>stream: try_recv
    stream->>Client: Gets Packet
    Note over Client: Emit events, etc.
    Client->>stream: try_send
    Note over Client,socket: PostUpdate
    stream->>socket: flush_send
    socket->>Alice: send bytes
```

rj00a:
Actually I just realized it would make more sense to put the semaphore permit in `valence_network`'s impl of `PacketStream` so it can't be lost.

dyc3:
I've moved mostly everything related to accepting connections into `valence_network`. There's a few unsolved problems that I might need help with:

- what to do with dimensions and biomes?
- we need some way to configure a bundle to be added with a client, or defer actually creating the entities to a system in `valence`.
   - Because client entities need to also have an `Inventory` component, and importing `valence` would be a circular dep tree

rj00a:
Let me lay down a few things just so we're on the same page.
- `valence_network` depends on `valence`, and not the other way around.
- Dimensions and biomes stay in `valence`.
- All of the stuff in `byte_channel`, `connect`, and `packet_manager` goes in `valence_network`.
- Your `config.rs` in `valence_network` looks correct.
- `PlayPacketSender`/`PlayPacketReceiver` can go away.
- Within `valence` we can probably get rid of `SharedServer` and just merge all the remaining fields into `Server`.
- Here are the fields of `Server` that should end up in `valence_network`:
  - `address`
  - `connection_mode`
  - `compression_threshold` (probably needs to be copied in both places)
  - `max_connections`
  - `incoming_capacity`
  - `outgoing_capacity`
  - `tokio_handle`
  - `tokio_runtime`
  - `new_clients_send`
  - `new_clients_recv`
  - `connection_sema`
  - `rsa_key`
  - `public_key_der`
  - `http_client`

rj00a:
Just a thought: The [ability to enable/disable plugins in a general way](https://github.com/bevyengine/bevy/discussions/6744) in the "stageless" rearchitecture of Bevy (coming in the next release) would make a lot of sense for toggleable "Open to LAN" functionality in singleplayer mode. And even in regular servers I could see it being useful too.

dyc3:
From discussion on discord: this valence_network refactor is too complex for me to handle right now, so I'm just going to be adding the `PacketStream` trait and refactoring to integrate that.

crates/valence_new/src/packet_stream.rs
```diff
@@ -0,0 +1,181 @@
+use std::collections::VecDeque;
+
+use bytes::{BufMut, BytesMut};
+use valence_protocol::packets::S2cPlayPacket;
+use valence_protocol::{DecodePacket, EncodePacket, PacketDecoder, PacketEncoder};
+
+use crate::server::byte_channel::{ByteReceiver, ByteSender, TryRecvError};
+
+/// Represents a byte stream that packets can be read from and written to.
+pub trait PacketStream {
```
rj00a, line 10: `try_send` and `try_recv` need to operate on bytes of encoded packets rather than actual packet objects. `PacketEncoder` and `PacketDecoder` should stay in `Client`. It's meant to represent the interface that a pair of byte channels gives you. We couldn't do it this way even if we wanted because the trait is not object-safe and we are putting it in a `Box<dyn PacketStream>`.

(I didn't really make this clear sorry)

dyc3: Yeah I just discovered this :(

crates/valence_new/src/packet_stream.rs
```diff
@@ -0,0 +1,260 @@
...
+use bytes::BytesMut;
+use tokio::task::JoinHandle;
+use valence_protocol::packets::S2cPlayPacket;
+use valence_protocol::{DecodePacket, EncodePacket, PacketDecoder, PacketEncoder};
+
+use crate::packet::WritePacket;
+use crate::server::byte_channel::{ByteReceiver, ByteSender, TryRecvError, TrySendError};
+
+pub struct PacketStreamer {
+    pub stream: Arc<Mutex<dyn PacketStream + Send + Sync>>,
```
rj00a, line 12: Hmm this isn't quite right. We definitely don't want to be locking and unlocking a mutex every time we write or send/recv packets.

Let's see if we can untangle this a bit. First I would suggest renaming `PacketStream` to `ConnectionState` (or just `Connection`? `ClientConnection`? take your pick) since its job is to abstract away all the junk needed to maintain a real TCP connection. Here is what `Client` should look like:

```rust
struct Client {
    connection: Box<dyn ConnectionState>,
    enc: PacketEncoder,
    dec: PacketDecoder,
    ...
}
```

Notice that no matter what, we need packet encoders and decoders in `Client`. We are only using `ConnectionState` to flush all the packets in our encoder at the end of the tick, or to retrieve a hunk of bytes to put into our decoder at the beginning of the tick.

If we give `PacketEncoder` the `WritePacket` trait, we will be able to use it in most places we were previously using `PlayPacketSender`.

Here is what our real connection state should look like:

```rust
struct RealConnectionState {
    send: ByteSender,
    recv: ByteReceiver,
    permit: OwnedSemaphorePermit,
    writer_task: Option<JoinHandle<()>>,
    reader_task: JoinHandle<()>,
    handle: Handle,
}
```

There should also be a `Drop` impl on `RealConnectionState` to match what `PlayPacketSender` and `PlayPacketReceiver` were doing before.

Finally, here is what `ConnectionState` should look like:

```rust
trait ConnectionState: Send + Sync + 'static {
    fn try_send(&mut self, bytes: BytesMut) -> anyhow::Result<()>;
    fn try_recv(&mut self) -> anyhow::Result<BytesMut>;
}
```

I'm using `anyhow::Result` to hide details about the byte channel because it doesn't need to be backed by one necessarily. An error from either `try_send` or `try_recv` indicates that we should consider the client disconnected.

dyc3:
@rj00a This is good info, but I'm not really worried about that right now, because right now this prototype is non-functional. Sorry for not being clear. We can worry about structure and semantics once it's actually working and I'm confident that the API surface for unit tests is sufficient. 

Per my DM on discord: My primary concern is with `MockPacketStream::collect_sent`. The error that prevents this code from compiling is:

```
error[E0499]: cannot borrow `self.send_dec` as mutable more than once at a time
   --> crates/valence_new/src/packet_stream.rs:189:32
    |
143 | impl<'a> MockPacketStream {
    |      -- lifetime `'a` defined here
...
189 |         while let Ok(packet) = self.send_dec.try_next_packet::<S2cPlayPacket<'a>>() {
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                                |
    |                                `self.send_dec` was mutably borrowed here in the previous iteration of the loop
    |                                argument requires that `self.send_dec` is borrowed for `'a`
```
The current situation is that `PacketDecoder` is set up such that it's impossible to collect `S2cPlayPacket`s up into a Vec. This is necessary in order to make assertions on packet order and to assert for duplicate packets. I have run out of ideas for how to make it work.

crates/valence_new/src/packet_stream.rs
```diff
@@ -0,0 +1,260 @@
...
+            .append_packet(&packet)
+            .expect("failed to encode injected packet");
+        let bytes = self.recv_enc.take();
+        self.pending_recv.extend_from_slice(bytes.as_ref());
+    }
+
+    /// Collects all the packets that have been sent so assertions can be made
+    /// on what the server sent in unit tests.
+    #[allow(dead_code)]
+    pub fn collect_sent(&'a mut self) -> anyhow::Result<Vec<S2cPlayPacket<'a>>> {
```
rj00a, line 185: So there are a few ways I see to resolve this.
1. Create an entirely new version of `S2cPlayPacket` that owns all of its data. Not fun.
2. Create a new method on `PacketDecoder` that allows you to collect borrowed packets into a vec. I believe a method like this should be possible on `PacketDecoder`
    ```rust
    fn collect_into_vec<'a, P: DecodePacket<'a>>(&'a mut self, vec: &'a mut Vec<P>) -> WhateverResult
    ```
    This isn't great because the method is not very general and would likely require a lot of code duplication to implement. But it can be done.
3. Structure your program in a way that does not require you to collect all the packets into a buffer. Have you considered accepting a list of packets that you expect to match, then walking through that list in order and decoding as you go? It's not a flexible but it might be sufficient for what you're doing.

dyc3: I believe 2 is the best way to go about this. This method would effectively bypass the code paths used in production for compression and encryption. The purpose of setting up this infra is to make assertions on packet contents and their order.  Compression and encryption are already tested, so the omission of these code paths from all other unit tests is acceptable.

rj00a: Alright I've done (2) in https://github.com/valence-rs/valence/commit/04d9588a2d249175428014a31b7e13d5afe501db

dyc3:
Thanks for [04d9588](https://github.com/valence-rs/valence/commit/04d9588a2d249175428014a31b7e13d5afe501db), I'm just going to cherry pick it into this branch for now. Merging looks like it's going to be hell

rj00a:
Some things I'd love to be able to test once this is complete because the entity/chunk code is horribly complicated:
- Entity update packets are only sent for entities that are considered loaded by the client.
- Chunks are not unloaded twice.
- Chunks and entities are unloaded when leaving the client's view.

rj00a:
I've implemented my previous suggestions along with the `ClientConnection` trait in https://github.com/valence-rs/valence/commit/0b4e9085a32a4d4ac7ba58aa277cc92654785f35 because it was quite a bit more involved than I was imagining.

rj00a:
Sorry about the merge conflicts 😬

dyc3:
There's a couple of weird quirks that I've discovered when working on `example_test_open_inventory`. And they somewhat harm dev ergonomics for unit tests.

- If the client is not assigned an instance as soon as it connects, `update_one_client` fails the first check, and the client is immediately disconnected.
- MockClientHelper is unable to expose collect_sent without doing some lifetime voodoo. (Likely a similar problem as before when collect_sent wouldn't compile.)

And regarding the merge conflicts, it seems like I'm probably going to have to rewrite most of this.

rj00a:
> If the client is not assigned an instance as soon as it connects, update_one_client fails the first check, and the client is immediately disconnected.

Perhaps you could wrap up all the initialization boilerplate (create an app, create an instance and client, add client to instance, etc.) in a helper function? Immediately disconnecting the client when not in an instance was a deliberate design choice, but maybe we could afford to be a bit lax there.

dyc3:
I think this is ready to review. I'm happy with the verbosity of the example tests.

crates/valence_new/src/unit_test/example.rs
```diff
@@ -0,0 +1,137 @@
...
+    use crate::{assert_packet_count, assert_packet_order};
+
+    /// The server's tick should increment every update.
+    #[test]
+    fn example_test_server_tick_increment() {
+        let mut app = App::new();
+        app.add_plugin(ServerPlugin::new(()));
+        let server = app.world.resource::<Server>();
+        let tick = server.current_tick();
+        drop(server);
```
rj00a, line 70: Dropping a reference does nothing.

crates/valence_new/src/unit_test/util.rs
```diff
@@ -0,0 +1,192 @@
...
+        Self {
+            buffers: Arc::new(Mutex::new(MockClientBuffers {
+                recv_buf: BytesMut::new(),
+                send_buf: BytesMut::new(),
+            })),
+        }
+    }
+
+    pub fn inject_recv(&mut self, bytes: BytesMut) {
+        self.buffers.lock().unwrap().recv_buf.extend(bytes);
```
rj00a, line 65:
```suggestion
        self.buffers.lock().unwrap().recv_buf.unsplit(bytes);
```

crates/valence_new/src/unit_test/util.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    pub fn clear_sent(&mut self) {
+        self.buffers.lock().unwrap().send_buf.clear();
+    }
+}
+
+impl ClientConnection for MockClientConnection {
+    fn try_send(&mut self, bytes: BytesMut) -> anyhow::Result<()> {
+        self.buffers.lock().unwrap().send_buf.extend(bytes);
```
rj00a, line 79:
```suggestion
        self.buffers.lock().unwrap().send_buf.unsplit(bytes);
```

crates/valence_new/src/unit_test/util.rs
```diff
@@ -0,0 +1,192 @@
...
+        .id();
+    (client_ent, client_helper)
+}
+
+#[macro_export]
+macro_rules! assert_packet_order {
+    ($sent_packets:ident, $($packets:pat),+) => {{
+        let sent_packets: &Vec<valence_protocol::packets::S2cPlayPacket> = &$sent_packets;
+        let mut positions = Vec::new();
+        $(positions.push(sent_packets.iter().position(|p| matches!(p, $packets)));)*
```
rj00a, line 180:
```suggestion
        let positions = [
            $(sent_packets.iter().position(|p| matches!(p, $packets)),)*
        ];
```

rj00a (commented):
Looking good.

PR #201, opened 2023-01-28
Add a simple SNBT parser

CaveNightingale:
To implement the hoverEvent show_item, we need an SNBT parser and a SNBT serializer.
An SNBT parser is implemented in this commit.

### DIFF
--- a/crates/valence_nbt/src/snbt.rs
+++ b/crates/valence_nbt/src/snbt.rs
@@ -1,15 +1,17 @@
-use std::{
-    error::Error,
-    fmt::{Display, Formatter},
-    iter::Peekable,
-    str::Chars,
-};
+use std::error::Error;
+use std::fmt::{Display, Formatter};
+use std::iter::Peekable;
+use std::str::Chars;
 
-use crate::{tag::Tag, Compound, List, Value};
+use crate::tag::Tag;
+use crate::{Compound, List, Value};
 
 const STRING_MAX_LEN: usize = 32767;
+/// Maximum recursion depth to prevent overflowing the call stack.
+const MAX_DEPTH: usize = 512;
+
 #[derive(Debug, Clone, PartialEq, Eq, Copy)]
-pub enum SNBTErrorType {
+pub enum SnbtErrorKind {
     ReachEndOfStream,
     InvalidEscapeSequence,
     EmptyKeyInCompound,
@@ -20,35 +22,37 @@ pub enum SNBTErrorType {
     DifferentTypesInList,
     LongString,
     TrailingData,
+    DepthLimitExceeded,
 }
 
-impl Display for SNBTErrorType {
+impl Display for SnbtErrorKind {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        use SNBTErrorType::*;
+        use SnbtErrorKind::*;
         match self {
-            ReachEndOfStream => write!(f, "Reach end of stream"),
-            InvalidEscapeSequence => write!(f, "Invalid escape sequence"),
-            EmptyKeyInCompound => write!(f, "Empty key in compound"),
-            ExpectColon => write!(f, "Expect colon"),
-            ExpectValue => write!(f, "Expect value"),
-            ExpectComma => write!(f, "Expect comma"),
-            WrongTypeInArray => write!(f, "Wrong type in array"),
-            DifferentTypesInList => write!(f, "Different types in list"),
-            LongString => write!(f, "Long string"),
-            TrailingData => write!(f, "Extra data after end"),
+            ReachEndOfStream => write!(f, "reach end of stream"),
+            InvalidEscapeSequence => write!(f, "invalid escape sequence"),
+            EmptyKeyInCompound => write!(f, "empty key in compound"),
+            ExpectColon => write!(f, "expect colon"),
+            ExpectValue => write!(f, "expect value"),
+            ExpectComma => write!(f, "expect comma"),
+            WrongTypeInArray => write!(f, "wrong type in array"),
+            DifferentTypesInList => write!(f, "different types in list"),
+            LongString => write!(f, "long string"),
+            TrailingData => write!(f, "extra data after end"),
+            DepthLimitExceeded => write!(f, "depth limit exceeded"),
         }
     }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Copy)]
-pub struct SNBTError {
-    pub error_type: SNBTErrorType,
+pub struct SnbtError {
+    pub error_type: SnbtErrorKind,
     pub line: usize,
     pub column: usize,
 }
 
-impl SNBTError {
-    pub fn new(error_type: SNBTErrorType, line: usize, column: usize) -> Self {
+impl SnbtError {
+    pub fn new(error_type: SnbtErrorKind, line: usize, column: usize) -> Self {
         Self {
             error_type,
             line,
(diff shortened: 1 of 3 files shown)


crates/valence_nbt/src/snbt.rs
```diff
@@ -0,0 +1,685 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "@ {},{}: {}", self.line, self.column, self.error_type)
+    }
+}
+
+impl Error for SNBTError {}
+
+type Result<T> = std::result::Result<T, SNBTError>;
+
+pub struct SNBTReader<'a> {
```
rj00a, line 70: Looks like we're not checking the recursion depth in the parser. This can cause the program to crash with a stack overflow if given deeply nested data. You can see how I've guarded against this in `from_binary_slice.rs` with the `check_depth` method.

We don't really need to worry about this in `SnbtWriter` because we trust the input.

crates/valence_nbt/src/snbt.rs
```diff
@@ -0,0 +1,685 @@
...
+        self.skip_whitespace();
+        if self.peek().is_ok() {
+            return Err(self.make_error(SNBTErrorType::TrailingData));
+        }
+        Ok(value)
+    }
+
+    /// Get the number of bytes readed.
+    /// It's useful when you want to read a SNBT string from an command argument since there may be trailing data.
+    pub fn bytes_readed(&self) -> usize {
```
rj00a, line 413: The past tense of "read" is spelled "read" so this should be `bytes_read`.

crates/valence_nbt/src/snbt.rs
```diff
@@ -0,0 +1,685 @@
...
+    /// SNBT is quite similar to JSON, but with some differences.
+    /// See [the wiki](https://minecraft.gamepedia.com/NBT_format#SNBT_format) for more information.
+    /// # Example
+    /// ```
+    /// use nbt::Value;
+    /// use nbt::SNBTReader;
+    /// let value = SNBTReader::from_snbt("1f").unwrap();
+    /// assert_eq!(value, Value::Float(1.0));
+    /// ```
+    pub fn from_snbt(snbt: &str) -> Result<Value> {
```
rj00a, line 428: Rather than have this as a function on `SnbtReader`, it might be best to export a free function like `from_snbt_str` at the root of the crate. This would be consistent with `from_binary_slice` and `to_binary_writer`. Same goes for `SnbtWriter::stringify`, except that it should be named `to_snbt_string`.

crates/valence_nbt/src/snbt.rs
```diff
@@ -0,0 +1,685 @@
...
+        SNBTReader::new(snbt).read()
+    }
+}
+
+pub struct SNBTWriter {
+    output: String,
```
rj00a, line 434: It would be nice if this were a `&mut String` so we can append to an existing string buffer.

crates/valence_nbt/src/snbt.rs
```diff
@@ -0,0 +1,685 @@
...
+}
+
+impl Error for SNBTError {}
+
+type Result<T> = std::result::Result<T, SNBTError>;
+
+pub struct SNBTReader<'a> {
+    line: usize,
+    column: usize,
+    pub index: usize,
```
rj00a, line 73: This does not need to be public because we have the `bytes_read` method.

crates/valence_nbt/src/snbt.rs
```diff
@@ -0,0 +1,685 @@
...
+        self.output.push(']');
+    }
+
+    fn write_primitive(&mut self, postfix: &str, value: impl ToString) {
+        self.output.push_str(&value.to_string());
+        self.output.push_str(postfix);
+    }
+
+    fn write_list(&mut self, list: &List) {
+        macro_rules! variant_impl {
```
rj00a, line 491: It would be better to use a lambda here instead of a macro if possible.

crates/valence_nbt/src/snbt.rs
```diff
@@ -0,0 +1,685 @@
...
+                }
+            }};
+        }
+        let target = self.read_unquoted_string()?;
+        match target
+            .bytes()
+            .last()
+            .ok_or_else(|| self.make_error(SNBTErrorType::ExpectValue))?
+        {
+            b'b' | b'B' => try_ret!(target[..target.len() - 1].parse::<i8>()),
```
rj00a, line 369: You can replace `target[..target.len() - 1]` with `target`.

crates/valence_nbt/src/snbt.rs
```diff
@@ -0,0 +1,685 @@
...
+            }
+        };
+        if target.len() > STRING_MAX_LEN {
+            return Err(self.make_error(SNBTErrorType::LongString));
+        }
+        Ok(Value::String(target))
+    }
+
+    /// Read the next element in the SNBT string.
+    /// [`SNBTErrorType::TrailingData`] is impossible to be returned since it doesnot consider it's an error.
```
rj00a, line 391:
```suggestion
    /// [`SNBTErrorType::TrailingData`] cannot be returned because it is not considered to be an error.
```

rj00a (commented):
- You'll need to use `cargo +nightly fmt` to format the code.
- There are some clippy errors that need to be fixed.

crates/valence_nbt/src/snbt.rs
CaveNightingale: Help needed.

crates/valence_nbt/src/snbt.rs
CaveNightingale: We need to pop the last char before `match` and add it back after that to replace `target[..target.len() - 1]` with `target`, which I think the solution is not so clear.

crates/valence_nbt/src/snbt.rs
rj00a: Actually I'm alright with this.

crates/valence_nbt/src/snbt.rs
rj00a: Ah, you're right. I misread.

crates/valence_nbt/src/snbt.rs
```diff
@@ -0,0 +1,707 @@
...
+/// SNBT is quite similar to JSON, but with some differences.
+/// See [the wiki](https://minecraft.gamepedia.com/NBT_format#SNBT_format) for more information.
+/// # Example
+/// ```
+/// use valence_nbt::snbt::SnbtReader;
+/// use valence_nbt::Value;
+/// let value = SnbtReader::from_snbt("1f").unwrap();
+/// assert_eq!(value, Value::Float(1.0));
+/// ```
+pub fn from_snbt_string(snbt: &str) -> Result<Value> {
```
rj00a, line 447: Perhaps this should be named `from_snbt_str`

crates/valence_nbt/src/snbt.rs
```diff
@@ -0,0 +1,707 @@
...
+}
+/// Parse a string in SNBT format into a `Value`.
+/// Assert that the string has no trailing data.
+/// SNBT is quite similar to JSON, but with some differences.
+/// See [the wiki](https://minecraft.gamepedia.com/NBT_format#SNBT_format) for more information.
+/// # Example
+/// ```
+/// use valence_nbt::snbt::SnbtReader;
+/// use valence_nbt::Value;
+/// let value = SnbtReader::from_snbt("1f").unwrap();
```
rj00a, line 444: This won't compile.

rj00a:
👍

PR #205, opened 2023-02-03
Refactor PacketDecoder

mymatsubara:
There was little bit of code duplication inside of the `PacketDecoder`. Because of that I did the following changes:

- Extract the packet data and length extraction logic to `DecodeBuffer(BytesMut)`
- Extract the packet decompression logic to `DecompressionBuffer(Vec<u8>)`
- Extract packet decoding logic to `fn decode_packet`
- Refactor the following methods from `PacketDecoder` using the extracted code: `try_next_packet`, `collect_into_vec` and `has_next_packet`
- Sprinkle some `#[inline]` attributes
- Add some comments

NOTE: First I tried to extract all the logic into new `PacketDecoder` methods, but the borrow checker was complaining. Because of that I had to create the wrapper structs `DecodeBuffer` and `DecompressionBuffer` to give more granular control over mutation and make the borrow checker happy.

Anyways, I'm not entirely sure these modifications are worth merging, so feel free to reject them.

### DIFF
--- a/crates/valence_protocol/src/codec.rs
+++ b/crates/valence_protocol/src/codec.rs
@@ -1,3 +1,5 @@
+use std::ops::{Deref, DerefMut};
+
 #[cfg(feature = "encryption")]
 use aes::cipher::{AsyncStreamCipher, NewCipher};
 use anyhow::{bail, ensure};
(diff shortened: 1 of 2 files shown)


dyc3:
The way `PacketDecoder` is set up is specifically to minimize the bytes being copied in the heap. This is a high priority for @rj00a and I because it significantly reduces memory usage and increases performance. Could you tell us some more about the impact this has on those metrics?

mymatsubara:
If the compiler inlines all the extracted functions/methods, which is expected, there should be no change in performance and memory allocations, since there was no change in code behavior (only structure).

The only gains we might get from these changes is reducing code duplication and a little bit more code readability.

dyc3 (approved):
Seems fine to me.

PR #213, opened 2023-02-04
Add `ChunkView` abstraction

rj00a:
A chunk view is a view distance combined with a center `ChunkPos`. It represents the set of chunks that a client can see.

`ChunkView` simplifies some code internally as well as making some examples like `parkour` and `terrain` easier to implement.

The first commit has introduced a bug that I am looking for. Go to the `cow_sphere` example, set your render distance to 2, and then fly away from and towards the sphere. Eventually there will be a giant hole in the sphere.

I might need to set up unit tests for the chunk/entity code first.

- [x] Find and fix bug.


crates/valence_new/src/view.rs
```diff
@@ -0,0 +1,161 @@
...
+/// [`Instance`](crate::instance::Instance).
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Default, Hash, Debug)]
+pub struct ChunkPos {
+    /// The X position of the chunk.
+    pub x: i32,
+    /// The Z position of the chunk.
+    pub z: i32,
+}
+
+const EXTRA_VIEW_RADIUS: i32 = 2;
```
dyc3, line 14: What's this for?

crates/valence_new/src/view.rs
```diff
@@ -0,0 +1,161 @@
...
+    fn from([x, z]: [i32; 2]) -> Self {
+        Self { x, z }
+    }
+}
+
+impl From<ChunkPos> for [i32; 2] {
+    fn from(pos: ChunkPos) -> Self {
+        [pos.x, pos.z]
+    }
+}
```
dyc3, line 61: Some round trip unit tests for these would be nice to have.

crates/valence_new/src/view.rs
```diff
@@ -0,0 +1,161 @@
...
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn all_in_view() {
```
dyc3, line 148: This only asserts that 
- `for_each` produces `ChunkPos` that pass contains
- `iter` produces `ChunkPos` that pass contains

You should add more tests that
- assert exactly what `ChunkPos` should be emitted
- assert that `iter` and `for_each` produce the same output
- assert that `contains` returns false when given as pos outside the view.

dyc3 (changes requested):
At a glance, the logic in client.rs seems to be ported over correctly, so I think there's something wrong with how `ChunkView` works

crates/valence_new/src/view.rs
rj00a: Chunks will only render if all the adjacent chunks are also loaded, so this value corrects for that. After some experimentation, 2 appears to be the correct value for this but I'm not certain. I'll add a comment explaining this.

crates/valence_new/src/view.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    // The foreach-based methods are optimizing better than the iterator ones.
+
+    #[inline]
+    pub(crate) fn for_each(self, mut f: impl FnMut(ChunkPos)) {
+        let true_dist = self.dist as i32 + EXTRA_VIEW_RADIUS;
+
+        for z in self.pos.z - true_dist..=self.pos.z + true_dist {
+            for x in self.pos.x - true_dist ..= self.pos.z + true_dist {
```
rj00a, line 124: Swapped an `x` with a `z` here 🤦. Slap me if I write more code without tests.

PR #215, opened 2023-02-04
[ECS] port more examples

dyc3:
- add kill() and win_game() to client
- add death example (not a port in the interest of making the example as minimal as possible)
- port biomes example
- add kick method to client
- port parkour example


rj00a (commented):
On my phone but the in the parkour example the instance should be a component of the player since every player got their own world on the old example.

dyc3:
Yeah that might work a bit better.

emortaldev:
woo parkour example ported wooo woow wo oo wooo

PR #217, opened 2023-02-05
[ECS] port even more examples

dyc3:
- add client.set_resource_pack
- refactor resource pack events to unsplit them
- port resource pack example
- add set_title and set_action_bar to client
- add play_particle to client
- port particles example
- port conway example
- partially port entity_raycast example

I was unable to finish the entity_raycast example because I don't really know how the spatial index package works, but if you fix the type errors it should just work.

### DIFF
--- a/crates/valence_new/examples/conway.rs
+++ b/crates/valence_new/examples/conway.rs
@@ -10,7 +10,10 @@ pub fn main() {
     tracing_subscriber::fmt().init();
 
     App::new()
-        .add_plugin(ServerPlugin::new(()))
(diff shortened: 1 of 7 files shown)


rj00a (commented):
- Would be nice to position the player closer to the particle emitter without the big spawn platform in the particle example.
  - A `play_particle` method should be added to `Instance` which writes to the packet buffer of the chunk it's in, but you don't have to address this here.
- The grass should be made bright green in the conway example to make it more visible.
- Unrelated, but it would be nice if the death example showed how to respawn in different instances.

PR #219, opened 2023-02-05
Gamemode selector

AviiNL:
probably not really mission critical, but shows the usage of set_op_level and the gamemode command, allows using F3+F4 to set gamemode instead of sneaking in the building example.


dyc3 (changes requested):
Put this in a new example.

PR #220, opened 2023-02-05
[ECS] Game mode switching (F3+F4)

AviiNL:
Very crude command interpreter to make game mode switching work via F3 + F4 if client's OP level is high enough.


rj00a (commented):
We should remember to update this when we have an actual command parser.

PR #225, opened 2023-02-08
Add ping calculation system

### LINKED ISSUE #207: Calculate Client Ping
There should be a~~n optional~~ component for `Client` that is able to keep track of the client's current ping. If the component is present, it should be used to display the client's current ping in the tab list via one of the systems in `PlayerList::default_system_set`.

I'm assuming the `Ping` and `Pong` packets can be used for this, but wiki.vg claims that the vanilla server does not use this.

rileysu:
Fixes #207 

Adds a simple ping calculation system using instants.


rileysu:
The ping seems to be heavily dependent on the tickrate. Testing on a laptop and pc gives me 54ms ping with 20TPS and 10ms - 20ms ping with 640TPS.

crates/valence_new/src/client/event.rs
```diff
@@ -924,6 +924,9 @@ fn handle_one_packet(
                 );
             } else {
                 client.got_keepalive = true;
+                client.ping = client.keepalive_sent_time.elapsed().as_millis() as u32;
+
+                tracing::info!("Ping is: {}, Instant is: {:?}", client.ping, client.keepalive_sent_time);
```
dyc3, line 929: probably shouldn't log this, it'll fill up the output pretty fast.

rileysu: I only had this line for debugging. I will remove it!

crates/valence_new/src/client.rs
```diff
@@ -64,6 +65,8 @@ pub struct Client {
     entities_to_despawn: Vec<VarInt>,
     got_keepalive: bool,
     last_keepalive_id: u64,
+    keepalive_sent_time: Instant,
+    ping: u32,
```
rj00a, line 69: Might be good to use an `i32` here and default to `-1` since this is what the player list expects.

rj00a (commented):
Don't forget to update the displayed ping in `PlayerList::default_system_set`

rj00a:
ECS rewrite branch was deleted so this has been closed automatically. You can open it again against `main`.

Issue #208, opened 2023-02-04
Optimize `VarInt`/`VarLong` encode and decode procedules


rj00a:
Varints are prolific in the Minecraft protocol. It makes sense to try and optimize them as best as we can. The current implementation is straightforward but there is room for improvement.

See:
- [how fast can you encode a varint?](https://steinborn.me/posts/performance/how-fast-can-you-write-a-varint/)
- [Moulberry's fast branchless encode/decode functions](https://github.com/Moulberry/Graphite/tree/master/crates/graphite_binary/src/varint)

Benchmarks must be made with `criterion` to establish confidence.

rileysu:
Would it increase performance just to unwrap the loop in this case? Other than that I am not sure how to convert the `Read` to a statically sized buffer that can deliver some of these performance improvements.

rj00a:
We don't need to use `Read`. `&mut &[u8]` should be used instead.

PR #227, opened 2023-02-09
VarInt/VarLong Encode and Decode Optimizations

guac42:
Aims to improve upon VarInt/VarLong Encode and Decode procedures (#208).
# Baselines:
## VarInt:
VarInt::decode
VarInt::encode
## VarLong
VarLong::encode
VarLong::decode


guac42:
image
Performance improvement of VarInt::encode on my machine. The original version was running at about 39ns.

guac42:
I tried a similar encode approach on VarLong and it resulted in some regression probably because it requires the use of many u128 bitwise ops.

Moulberry:
The code is not mine, I must have removed some comments during refactoring.
It comes from https://github.com/as-com/varint-simd/, specifically 
https://github.com/as-com/varint-simd/blob/0f468783da8e181929b01b9c6e9f741c1fe09825/src/encode/mod.rs#L71

guac42:
image
Performance improvement of VarLong::encode on my machine. The original version was running at about 40ns.

guac42:
I tested branchless approaches to VarInt::decode and VarLong::decode and saw either no improvement or regression.

crates/valence_protocol/src/var_int.rs
```diff
@@ -48,16 +48,31 @@ pub enum VarIntDecodeError {
...
-        let mut val = self.0 as u32;
-        loop {
-            if val & 0b11111111111111111111111110000000 == 0 {
-                w.write_u8(val as u8)?;
-                return Ok(());
-            }
-            w.write_u8(val as u8 & 0b01111111 | 0b10000000)?;
-            val >>= 7;
-        }
+        let x = unsafe { std::mem::transmute::<i32, u32>(self.0) } as u64;
```
rj00a, line 54: If I'm not mistaken, this should be equivalent to `self.0 as u64`.

crates/valence_protocol/src/var_int.rs
```diff
@@ -48,16 +48,31 @@ pub enum VarIntDecodeError {
...
+        let bytes_needed = 8 - unused_bytes;
+
+        // set all but the last MSBs
+        let msbs = 0x8080808080808080;
+        let msbmask = 0xffffffffffffffff >> (((8 - bytes_needed + 1) << 3) - 1);
+
+        let merged = stage1 | (msbs & msbmask);
+        let bytes = merged.to_le_bytes();
+
+        w.write_all(&bytes[..bytes_needed as usize])?;
```
rj00a, line 73: There might be bounds checks happening here in the slicing that aren't optimized away. Try `get_unchecked`?

crates/valence_protocol/src/var_long.rs
```diff
@@ -26,7 +30,59 @@ impl VarLong {
...
+        // Shift it down 1 byte so the last MSB is the only one set, and make sure only
+        // the MSB is set
+        let shift = unsafe { _mm_bsrli_si128(mask, 1) };
+        let msbmask = unsafe { _mm_and_si128(shift, _mm_set1_epi8(128u8 as i8)) };
+
+        // Merge the MSB bits into the vector
+        let merged = unsafe { _mm_or_si128(stage1, msbmask) };
+        let bytes = unsafe { std::mem::transmute::<__m128i, [u8; 16]>(merged) };
+
+        w.write_all(&bytes[..bytes_needed as usize])?;
```
rj00a, line 74: Same bounds check issue.

rj00a (commented):
- Can you include the old approaches in the benchmarks for comparison?
- Have you tried the technique described in "how fast can you encode a varint"? The author reported a 30% perf improvement in their environment.

crates/valence_protocol/src/var_long.rs
```diff
@@ -26,7 +30,59 @@ impl VarLong {
 }
 
 impl Encode for VarLong {
+    // Adapted from VarInt-Simd encode
+    // https://github.com/as-com/varint-simd/blob/0f468783da8e181929b01b9c6e9f741c1fe09825/src/encode/mod.rs#L71
+    #[cfg(all(
+        any(target_arch = "x86", target_arch = "x86_64"),
+        not(target_os = "macos")
+    ))]
     fn encode(&self, mut w: impl Write) -> Result<()> {
```
rj00a, line 39: Looks like we're missing a fallback implementation for non-x86 or non-x86_64 platforms. If you decide to use this approach, please make a generic implementation that works on all platforms.

guac42: Would this work for the fallback implementation?
```
#[cfg(any(
    not(any(target_arch = "x86", target_arch = "x86_64")),
    target_os = "macos"
))]
```

crates/valence_protocol/src/var_int.rs
guac42: Yea that's correct. I should have noticed that LOL.

crates/valence_protocol/src/var_long.rs
rj00a: Oh wait I didn't read the rest of the code. You're fine.

crates/valence_protocol/src/var_long.rs
```diff
@@ -1,7 +1,11 @@
+#[cfg(target_arch = "x86")]
+use std::arch::x86::*;
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
```
rj00a, line 4: You can scope these imports to the body of `encode` so you don't have to use `cfg` here.

guac42:
> Have you tried the technique described in "how fast can you encode a varint"?

Yes, I think the performance the author gained was mostly due to java inefficiencies.
Here is the rust implementation's performance:
image

rj00a:
> Here is the rust implementation's performance:

Can you include this code (and the old VarInt VarLong functions) in the criterion benchmarks?

Don't forget to mark this PR as "ready for review" if it is finished.

guac42:
> Can you include this code in the criterion benchmarks?

I have since written a more optimized version of what I had previously benched. It's still based on [how fast can you encode a varint?](https://steinborn.me/posts/performance/how-fast-can-you-write-a-varint/).
### Here's the performance:
image
### Here's the code:
```rust
fn encode(&self, mut w: impl Write) -> anyhow::Result<()> {
    if (self.0 as u32 & (0xffffffff << 7)) == 0 {
        w.write_u8(self.0 as u8)?;
    } else if (self.0 as u32 & (0xffffffff << 14)) == 0 {
        let val = (self.0 as u32 & 0x7F | 0x80) << 8 | (self.0 as u32 >> 7);
        w.write_u16::<BigEndian>(val as u16)?;
    } else if (self.0 as u32 & (0xffffffff << 21)) == 0 {
        let val = (self.0 as u32 & 0x7F | 0x80) << 16 | ((self.0 as u32 >> 7) & 0x7f | 0x80) << 8 | (self.0 as u32 >> 14);
        w.write_u24::<BigEndian>(val)?;
    } else if (self.0 as u32 & (0xffffffff << 28)) == 0 {
        let val = (self.0 as u32 & 0x7F | 0x80) << 24 | (((self.0 as u32 >> 7) & 0x7f | 0x80) << 16)
        | ((self.0 as u32 >> 14) & 0x7F | 0x80) << 8 | (self.0 as u32 >> 21);
        w.write_u32::<BigEndian>(val)?;
    } else {
        let val = ((self.0 as u64) & 0x7f | 0x80) | ((self.0 as u64 >> 7) & 0x7f | 0x80) << 8
            | ((self.0 as u64 >> 14) & 0x7f | 0x80) << 16 | ((self.0 as u64 >> 21) & 0x7f | 0x80) << 24
            | ((self.0 as u64 >> 28) & 0x7f) << 32;
        w.write_all(unsafe { val.to_le_bytes().get_unchecked(..5usize) })?;
    }
}
```

guac42:
I even tried going as far as to bypass rusts arithmetic checks using ```u32::wrapping_add``` etc. And didn't achieve any real performance gains. :/ 
But 5% is still a fair amount, I guess.

rj00a:
Thanks for taking the time to check. Btw arithmetic checks only happen in debug mode so `wrapping_add` is equivalent to `+` in release mode.

PR #228, opened 2023-02-10
Port terrain example

rj00a:
- Added `viewed` boolean to chunks, which is used to efficiently track chunks that can be removed from an instance because nobody is looking at it.
- Added `to_unloaded` function to loaded chunks. Used to make an unloaded clone of a loaded chunk.
- Ported terrain example to `valence_new`.
- Terrain generation is now fully asynchronous and will no longer block the main tick loop.
  - Unloaded chunks closer to players are prioritized over chunks that are further away.
# Test Plan

```bash
cargo r -p valence_new --example terrain
```

### DIFF
--- a/crates/valence_new/examples/terrain.rs
+++ b/crates/valence_new/examples/terrain.rs
@@ -79,7 +79,9 @@ fn setup(world: &mut World) {
     });
 
     // Chunks are generated in a thread pool for parallelism and to avoid blocking
-    // the main tick loop.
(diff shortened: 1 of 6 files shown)


dyc3 (approved):
No unit tests/doc tests? Otherwise, looks fine.

rj00a:
Testing this effectively seems a bit difficult. I'll have to think about this.

PR #238, opened 2023-02-13
Packet inspector gui

### LINKED ISSUE #235: Add a GUI to the packet inspector

The packet inspector is a handy tool, but having all its output dumped to `stdout` is not very useful.

It would be better if we had a user interface to view the packets. I'm imagining something like [pakkit](https://github.com/Heath123/pakkit), but up-to-date with the rest of the project and easy to run. [egui]() could be used to build the interface. The interface will be simple enough that a pure Rust solution should work well.

### Additional context

We will likely need to use `bevy_reflect` on all the packets in `valence_protocol` to have enough information for the GUI. This could be placed behind a `reflect` feature.

AviiNL:
Fixes #235

## Description

A GUI Project for the packet inspector

## Test Plan

N/A

<details>

<summary>Playground</summary>

```rust
N/A
```

</details>

Steps:
1. Start (any) server
2. Start the proxy `cargo r -r -p packet_inspector_gui -- 127.0.0.1:25560 127.0.0.1:25565`
3. Join server (127.0.0.1:25560)
4. Magic


crates/packet_inspector_gui/README.md
```diff
@@ -0,0 +1,77 @@
+# What's This?
```
dyc3, line 1: We should just point people to the README in `packet_inspector` instead of maintaining 2 documents. Feel free to update that document to mention `packet_inspector_gui`.

dyc3 (commented):
I was running this along side the terrain example, and I couldn't join via the packet inspector.

```
Accepted connection to 127.0.0.1:58804
Connecting to 127.0.0.1:25565
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: IndeterminateOffset', crates/packet_inspector_gui/src/state.rs:63:53
```

crates/packet_inspector_gui/README.md
rj00a: I think we should just replace the old packet inspector entirely with this project.

JoelEllis: In my experience, the old packet inspector introduces a few bugs - for example, I've had it not send chat messages until the player moves and not send block interact events until the player hits the ground :/.
Given that, I'm not sure that keeping the old one around is even worth it. Perhaps adding a feature to disable the GUI and log the packets instead in this new project would be a better idea?

JoelEllis:
A small issue I've noticed is that Text elements get stringified / stripped  of their formatting. This isn't so bad in this example, but it might be more of an issue if you were testing more advanced formatting, like hover text. 

image

Otherwise, this is great! I really like it :smile:

rj00a:
> A small issue I've noticed is that Text elements get stringified / stripped of their formatting.

This is an issue with the `Debug` impl of `Text`. I agree that it should be changed.

crates/packet_inspector_gui/README.md
qualterz: > I think we should just replace the old packet inspector entirely with this project.

For me, CLI tools much more useful than GUI applications, so I would not appreciate the removal of `packet_inspector` package.

AviiNL: Then plan was to incorporate the CLI tool with this, with a `--nogui` or something cli switch

crates/packet_inspector/src/context.rs
```diff
@@ -0,0 +1,365 @@
...
+                    PacketDirection::ClientToServer => self.c2s_style,
+                    PacketDirection::ServerToClient => self.s2c_style,
+                };
+
+                println!(
+                    "[{}] ({}) {}",
+                    systemtime_strftime(packet.created_at),
+                    arrow.style(style),
+                    packet.get_packet_string_no_format().style(style)
+                )
```
rj00a, line 334: When redirecting the output to a file, the color escape codes are still present. You can fix this with [atty](https://crates.io/crates/atty). Would be nice to have a command line argument to override the default behavior.

rj00a (commented):
When running without args, it looks a bit weird when the window opens in fullscreen with the "Setup" widget displayed in the top left corner. Maybe in this case you can lock the size of the window to the size of the "Setup" widget? Barring that, center the widget in the window?

crates/packet_inspector/src/state.rs
```diff
@@ -0,0 +1,82 @@
...
+
+        self.enc.append_packet(&pkt)?;
+
+        let bytes = self.enc.take();
+        self.write.write_all(&bytes).await?;
+
+        let time = match OffsetDateTime::now_local() {
+            Ok(time) => time,
+            Err(_) => {
+                eprintln!("Unable to get local time, using UTC"); // this might get a bit spammy..
```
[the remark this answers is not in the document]
AviiNL: Using UTC is satisfactory though? just remove the log line?

crates/packet_inspector/src/context.rs
```diff
@@ -0,0 +1,365 @@
...
+    pub(crate) stage: Stage,
+    pub(crate) packet_type: i32,
+    pub(crate) packet_name: String,
+    pub(crate) created_at: OffsetDateTime,
+}
+
+impl From<&mut Packet> for String {
+    fn from(value: &mut Packet) -> Self {
+        if value.packet_data.len() > 1024 {
+            return "Packet too large".to_string();
```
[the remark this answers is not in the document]
AviiNL: Yeah I'll remove this entirely, it doesnt even work correctly since the length it returns is of the compressed packet. (Which even for chunk data, still fits in 1024 at least on terrain example)

crates/packet_inspector/src/state.rs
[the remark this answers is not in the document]
rj00a: Yeah I'd just remove the print.

AviiNL:
> > A small issue I've noticed is that Text elements get stringified / stripped of their formatting.
> 
> This is an issue with the `Debug` impl of `Text`. I agree that it should be changed.

Also think a separate issue should be created for this.

crates/packet_inspector/src/state.rs
```diff
@@ -0,0 +1,82 @@
...
+            }
+
+            self.dec.queue_bytes(buf);
+        }
+
+        let has_compression = self.dec.compression();
+        let pkt: P = self.dec.try_next_packet()?.unwrap();
+
+        self.buf.clear();
+        write!(&mut self.buf, "{pkt:?}")?;
```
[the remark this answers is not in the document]
AviiNL: What do you recommend to do with this for now? Leave it as-is and change it if/when `packet_name` on `Packet` becomes a thing? (or block this pr until it becomes a thing)

crates/packet_inspector/src/main.rs
```diff
@@ -23,107 +34,79 @@ use valence_protocol::packet::c2s::status::{QueryPingC2s, QueryRequestC2s};
...
-    eprintln!("Waiting for connections on {}", cli.client_addr);
-    let listen = TcpListener::bind(cli.client_addr).await?;
+    let client_addr = match cli.client_addr {
+        Some(addr) => addr,
+        None => panic!("client_addr must be set"),
+    };
+
+    let server_addr = match cli.server_addr {
+        Some(addr) => addr,
+        None => panic!("server_addr must be set"),
```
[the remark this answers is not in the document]
AviiNL: Technically, if server and client addr are None at this point, it's a bug, since Clap enforces a value with
```
#[arg(required_if_eq("nogui", "true"))]
```

But yeah, returning an error should not be that different in that both would do an exit.

PR #240, opened 2023-02-14
Minimal stresser implementation

### LINKED ISSUE #211: Create fake client(s) for testing purposes similar to `rust-mc-bot`.
[rust-mc-bot](https://github.com/Eoghanmc22/rust-mc-bot) is a tool I've been using to stress test the server. It would be nice if we had our own version of it so we don't have to work with a tool in a separate git repo.

Ideally our version will behave very similarly. It will be built on top of `valence_protocol`.

qualterz:
## Description

An implementation of a Minecraft server stresser for testing purposes.

The potential of this pull request is to implement a minimal stresser binary package that would be bound to the local `valence_protocol` package, so it would be always up to date with the latest Valence Minecraft protocol implementation.

The MVP version is going to be able concurrently connect headless clients to a target Minecraft server.

## Test Plan

Steps:
1. Ensure that the connection mode is offline
2. Run `cargo run --example bench_players` or any other example
3. Run `cargo run --package valence_stresser -- --target 127.0.0.1:25565 --count 1000`
4. Monitor the `bench_players` output
 
#### Related

closes #211


qualterz:
I've tested it with 1000 connections, and it works fine in my environment.

crates/valence_stresser/src/args.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub connections_count: usize,
+
+    /// Name prefix of connections.
+    #[arg(default_value = "Stresser")]
+    #[arg(short = 'n', long = "name")]
+    pub name_prefix: String,
+
+    /// Spawn cooldown of connections in milliseconds.
+    /// The lower the value, the more frequently connections are spawned.
+    #[arg(default_value = "100")]
```
dyc3, line 21: The default should be lower. Maybe 10?

crates/valence_stresser/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+    let target_addr = args.target_host.to_socket_addrs().unwrap().next().unwrap();
+
+    let mut last_thread: Option<JoinHandle<()>> = None;
+
+    for conn_index in 0..args.connections_count {
+        let conn_name = format!("{}{}", args.name_prefix, conn_index);
+
+        last_thread = Some(thread::spawn(move || {
+            make_connection(target_addr, conn_name.as_str());
+        }));
```
dyc3, line 24: This is really inefficient and makes it so that the kernel needs to be heavily relied on to manage thread execution. It would be way more preferable to make this async instead.

dyc3 (changes requested):
This is certainly a good start.

nit: The `bench_players` example would be a better choice for your test plan.
nit: Your test plan technically doesn't require the packet_inspector to be used, so I'd remove it.

crates/valence_stresser/src/args.rs
qualterz: Actually, it depends on how powerful the hardware on which the stresser is run. So there is a need to find out the average value that would fit most hardware. If the value is too low, there is a high potential for timed-out disconnections from a server due to overload.

crates/valence_stresser/src/main.rs
qualterz: I thought about using an asynchronous way of running connections, but I think that a dedicated thread for each connection would be the most effective way to stress a server, rather than relying on an asynchronous runtime.

qualterz:
> nit: Your test plan technically doesn't require the packet_inspector to be used, so I'd remove it.

I have used the `packet_inspector` to monitor the packet traffic.

crates/valence_stresser/src/main.rs
rj00a: @dyc3 is right. async should simply improve the performance of the stresser.

dyc3: In general, async programs will out perform threaded programs when there is IO involved, which is very much the case here. Valence's network code is async for this reason. The OS is allowed to switch the executing thread for any given CPU core at any time, which includes interrupting code that is not blocked by IO operations. These context swtiches are not free, not to mention that there is also a memory overhead for each thread.

I tried running the stresser with 4000 clients and a low join cooldown value. My computer seized up for a good 10 seconds because of the threads being created.

qualterz: Alright, I'll research it more, do some benchmarks, and apply the changes.

qualterz:
I'm not strong in asynchronous programming, and there's a problem that sessions are getting timed out disconnects.

qualterz:
Alright, I've fixed that bug with timed out disconnects. I tested everything in release mode, up to 4000 sessions were handled before the out-of-memory daemon killed the server.

dyc3 (approved):
Fantastic! Performs much better.

PR #243, opened 2023-02-15
Add ping calculation system

rileysu:
## Description

Add the ping computation system and capability to send ping to connecting clients.

Currently the player list will only update upon connection. Should another ticket be made for updating this since it seems out of scope for this issue?

## Test Plan

Printing ping per keepalive packet on a different computer on my network. It seems to work fine but would need extra feedback.

#### Related

#207


rj00a:
> Should another ticket be made for updating this since it seems out of scope for this issue?

That would be ideal.

PR #244, opened 2023-02-15
Add sounds

### LINKED ISSUE #206: Extracting Sound Data
We need to extract sound IDs. We should use the [extractor](https://github.com/valence-rs/valence/tree/main/extractor) and expose an API as part of `valence_protocol`. It should be used within the `SoundEffect` and `EntitySoundEffect` packets. Then, a `play_sound` method should be added to `Client`.

Gingeh:
This is my first time contributing here so I was pretty unfamiliar with the codebase and may have done some things incorrectly.

## Description
 - Added a sound extractor to extract sound event ids and identifiers
 - Added a `Sound` enum (with a build script) to represent sound effects
 - Added a `play_sound` method to `Instance` and `Client`
 - Re-implemented sound effects in the parkour example

## Test Plan
I tested this using the sounds I added to the parkour example.

#### Related
Hopefully fixes #206


crates/valence_protocol/src/sound.rs
```diff
@@ -0,0 +1,32 @@
...
+// associated const fn functions of `Sound` to access details about a sound.
+include!(concat!(env!("OUT_DIR"), "/sound.rs"));
+
+use crate::packets::s2c::play::SoundId;
+use crate::Ident;
+
+impl Sound {
+    pub fn to_id(self) -> SoundId<'static> {
+        SoundId::Direct {
+            id: Ident::new(self.to_str()).unwrap(),
```
rj00a, line 12: When `ident_str` is added in #32 it can be used here to avoid parsing the ident at runtime.

crates/valence/src/instance.rs
```diff
@@ -363,6 +364,32 @@ impl Instance {
...
+        let position = position.into();
+
+        self.write_packet_at(
+            &SoundEffect {
+                id: sound.to_id(),
+                category,
+                position: (position * 8.0).as_ivec3().into(),
+                volume,
+                pitch,
+                seed: rand::random(),
```
rj00a, line 387: We might want to add `play_sound_seeded` variants of this method that take the seed as a parameter in the future.

rj00a (approved):
Looks good to me 👍

PR #246, opened 2023-02-17
Add minimal chat example

JoelEllis:
## Description

I've added a minimal chat example - ignoring chat signing, etc. It's in the file `crates/valence/examples/chat.rs`.

## Test Plan

Manual, unfortunately :(

Steps:
1. Open two minecraft clients, with two separate accounts (or usernames if using offline mode)
2. Connect to the server in both clients
3. Type a chat message, checking it appears to both clients, with the correct username.
4. Type a command, checking it appears only to the sender.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,7 @@ Valence aims to be the following:
 
 ## Current Status
 
-Valence is still early in development with many features unimplemented or incomplete. However, the foundations are in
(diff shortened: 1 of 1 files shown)


JoelEllis:
I need to add a precommit hook for `cargo fmt`!

PR #248, opened 2023-02-17
Add a minimal chat example

JadedBlueEyes:
## Description

Take 3 of #247.

I've added a minimal chat example - ignoring chat signing, etc. It's in the file `crates/valence/examples/chat.rs`.

## Test Plan

Manual, unfortunately :(

Steps:
1. Open two minecraft clients, with two separate accounts (or usernames if using offline mode)
2. Connect to the server in both clients
3. Type a chat message, checking it appears to both clients, with the correct username.
4. Type a command, checking it appears only to the sender.

### DIFF
--- a/crates/valence/examples/chat.rs
+++ b/crates/valence/examples/chat.rs
@@ -46,7 +46,7 @@ fn init_clients(
     instances: Query<Entity, With<Instance>>,
 ) {
     for mut client in &mut clients {
-        client.set_position([0.0, SPAWN_Y as f64, 0.0]);
+        client.set_position([0.0, SPAWN_Y as f64 + 1.0, 0.0]);
         client.set_instance(instances.single());
         client.set_game_mode(GameMode::Adventure);
         client.send_message("Welcome to Valence! Talk about something.".italic());
--- a/crates/valence/examples/chat.rs
+++ b/crates/valence/examples/chat.rs
@@ -5,7 +5,6 @@ use valence::client::event::{default_event_handler, ChatCommand, ChatMessage};
 use valence::prelude::*;
 
 const SPAWN_Y: i32 = 64;
-const ARENA_RADIUS: i32 = 32;
 
 pub fn main() {
     tracing_subscriber::fmt().init();
@@ -19,7 +18,6 @@ pub fn main() {
         .add_system(init_clients)
         .add_system(despawn_disconnected_clients)
         .add_system_set(PlayerList::default_system_set())
-        .add_system(teleport_oob_clients)
         .run();
 }
 
@@ -34,24 +32,9 @@ fn setup(world: &mut World) {
         }
     }
 
-    // Create circular arena.
-    for z in -ARENA_RADIUS..ARENA_RADIUS {
-        for x in -ARENA_RADIUS..ARENA_RADIUS {
-            let dist = f64::hypot(x as _, z as _) / ARENA_RADIUS as f64;
-
-            if dist > 1.0 {
-                continue;
-            }
-
-            let block = if rand::random::<f64>() < dist {
-                BlockState::STONE
-            } else {
-                BlockState::DEEPSLATE
-            };
-
-            for y in 0..SPAWN_Y {
-                instance.set_block_state([x, y, z], block);
-            }
+    for z in -25..25 {
+        for x in -25..25 {
+            instance.set_block_state([x, SPAWN_Y, z], BlockState::BEDROCK);
         }
     }
 
@@ -103,13 +86,5 @@ fn handle_command_events(mut clients: Query<&mut Client>, mut commands: EventRea
             "You sent the command ".into_text() + ("/".into_text() + (message).into_text()).bold();
 
         client.send_message(formatted);
-    }
-}
-
-fn teleport_oob_clients(mut clients: Query<&mut Client>) {
-    for mut client in &mut clients {
-        if client.position().y < 0.0 {
-            client.set_position([0.0, SPAWN_Y as _, 0.0]);
-        }
     }
 }
--- a/crates/valence/examples/chat.rs
+++ b/crates/valence/examples/chat.rs
@@ -34,7 +34,7 @@ fn setup(world: &mut World) {
 
     for z in -25..25 {
         for x in -25..25 {
-            instance.set_block_state([x, SPAWN_Y, z], BlockState::BEDROCK);
+            instance.set_block([x, SPAWN_Y, z], BlockState::BEDROCK);
         }
     }
 
--- a/crates/valence/examples/chat.rs
+++ b/crates/valence/examples/chat.rs
@@ -84,8 +84,7 @@ fn handle_message_events(mut clients: Query<&mut Client>, mut messages: EventRea
             .color(Color::YELLOW)
             + message.into_text().not_bold().color(Color::WHITE);
 
-        // TODO! Is there a more efficient way to do this?
-        clients.for_each_mut(|mut client| {
+        clients.par_for_each_mut(16, |mut client| {
             client.send_message(formatted.clone());
         })
     }
--- a/crates/valence/examples/chat.rs
+++ b/crates/valence/examples/chat.rs
@@ -10,7 +10,7 @@ pub fn main() {
     tracing_subscriber::fmt().init();
 
     App::new()
-        .add_plugin(ServerPlugin::new(()))
+        .add_plugin(ServerPlugin::new(()).with_connection_mode(ConnectionMode::Offline))
         .add_startup_system(setup)
         .add_system_to_stage(EventLoop, default_event_handler)
         .add_system_to_stage(EventLoop, handle_message_events)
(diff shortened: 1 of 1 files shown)


JoelEllis:
[Video here!](https://cdn.discordapp.com/attachments/1076244963056169011/1076261657883447356/vid.mp4)

qualterz:
> [Video here!](https://cdn.discordapp.com/attachments/1076244963056169011/1076261657883447356/vid.mp4)

That low-quality sound on the background is amazing :D

JoelEllis:
> That low-quality sound on the background is amazing :D

Not deliberate though! The video recorder I used for it has a questionable encoder, but setting up OBS for a 30 second video seems overkill 😆

crates/valence/examples/chat.rs
```diff
@@ -0,0 +1,116 @@
...
+        };
+
+        let message = message.message.to_string();
+
+        let formatted = format!("<{}>: ", client.username())
+            .bold()
+            .color(Color::YELLOW)
+            + message.into_text().not_bold().color(Color::WHITE);
+
+        // TODO! Is there a more efficient way to do this?
```
qualterz, line 87: Talking about efficiency, there is a way to for each the clients in a parallel way:

```rust
clients.par_for_each_mut(16, |mut client| {
    client.send_message(formatted.clone());
})
```

But there is a doubt about what the batch size should be used. A similar approach: https://github.com/valence-rs/valence/blob/50018a52bf31eb5f1cacf8ede7c041e2eb83eafd/crates/valence/src/client.rs#L630-L631

dyc3:
For future reference, you can just update your original PR with any changes. There's no need to make another PR if you mess it up.

Also for this project, we squash commit each PR, so you don't need to worry about amending commits or anything to make the history clean.

crates/valence/examples/chat.rs
dyc3: Technically, the most efficient way to do this is for valence to have a broadcast method that avoids re-serializing the packet for every client it needs to send it to. But this is perfectly fine for an example.

JoelEllis:
> For future reference, you can just update your original PR with any changes. There's no need to make another PR if you mess it up.
> 
> Also for this project, we squash commit each PR, so you don't need to worry about amending commits or anything to make the history clean.

Git was picking up a directory it shouldn't, and that was breaking the build. To fix it, I was soft resetting to upstream, and then force pushing - but then Github would close the PR, because there were no commits.

crates/valence/examples/chat.rs
JoelEllis: I mean, if we were going for the most efficient possible, we could do it outside the event loop to avoid waiting for the next tick. That might even be useful for things that interact with an external database, running it outside of the loop to avoid blocking.

I'll update this with `par_for_each_mut` in the morning. 🙂

JoelEllis: > what the batch size should be used

I think this probably depends mainly on the amount of CPU cores and the amount of items, but it might also depend on eg. CPU cache size. There's probably no perfect number :woman_shrugging:

JoelEllis:
It should be good to merge now, unless anyone else has any feedback? :smile:

qualterz (commented):
I hope this hard coded bulk size wouldn't play a bad joke.

crates/valence/examples/chat.rs
```diff
@@ -0,0 +1,115 @@
...
+                BlockState::STONE
+            } else {
+                BlockState::DEEPSLATE
+            };
+
+            for y in 0..SPAWN_Y {
+                instance.set_block_state([x, y, z], block);
+            }
+        }
+    }
```
dyc3, line 56: Make this more simple. A square 1 block high would suffice.

Because this is an example, any code not related to what the example is showing off needs to be as minimal as possible.

crates/valence/examples/chat.rs
```diff
@@ -0,0 +1,115 @@
...
+        let message = command.command.to_string();
+
+        let formatted =
+            "You sent the command ".into_text() + ("/".into_text() + (message).into_text()).bold();
+
+        client.send_message(formatted);
+    }
+}
+
+fn teleport_oob_clients(mut clients: Query<&mut Client>) {
```
dyc3, line 109: System is not necessary for this example.

dyc3 (changes requested):

crates/valence/examples/chat.rs
qualterz: Actually, you can generate infinity layer of barrier block.

crates/valence/examples/chat.rs
```diff
@@ -0,0 +1,115 @@
...
+    for command in commands.iter() {
+        let Ok(mut client) = clients.get_component_mut::<Client>(command.client) else {
+            warn!("Unable to find client for message: {:?}", command);
+            continue;
+        };
+
+        let message = command.command.to_string();
+
+        let formatted =
+            "You sent the command ".into_text() + ("/".into_text() + (message).into_text()).bold();
```
qualterz, line 103: I think would be good to have an example of command like `/echo`.

AviiNL: That is literally what this command is, in my opinion though, Chat Example != Command Example, and I think the command code doesn't (shouldn't) be in this example. Especially since a proper command system is yet to be built. and when it does, the more examples that uses/demonstrates commands will have to be changed, so to reduce the overhead on that, I would strongly suggest removing it from this example.

dyc3: No, this is perfectly fine. We don't have command parsing implemented yet. An example implementing commands would go in a separate example.

JoelEllis: I agree with @dyc3 for this - even if I added a (complex) command system, it would be incomplete until brigadier nodes are sent to the client, and I hardly even know where to start with that other than #35.

dyc3: For reference: we have a thread on discord discussing the design for a `valence_command` crate: https://discord.com/channels/998132822239870997/1072277129183625256

crates/valence/examples/chat.rs
```diff
@@ -0,0 +1,90 @@
...
+use valence::client::event::{default_event_handler, ChatCommand, ChatMessage};
+use valence::prelude::*;
+
+const SPAWN_Y: i32 = 64;
+
+pub fn main() {
+    tracing_subscriber::fmt().init();
+
+    App::new()
+        .add_plugin(ServerPlugin::new(()))
```
dyc3, line 13: We should default this example to offline mode to reduce the friction of testing with 2 clients.

crates/valence/examples/chat.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    world.spawn(instance);
+}
+
+fn init_clients(
+    mut clients: Query<&mut Client, Added<Client>>,
+    instances: Query<Entity, With<Instance>>,
+) {
+    for mut client in &mut clients {
+        client.set_position([0.0, SPAWN_Y as f64, 0.0]);
```
dyc3, line 49: Whenever I spawn in, it appears that I'm spawning under the platform. Add a couple blocks to the y parameter here.

dyc3 (changes requested):

crates/valence/examples/chat.rs
qualterz: I think there is need in development config where offline mode is enabled by default, so all examples that use this config would be easier to play.

JoelEllis: Yeah, I edited this to the default for consistency with the rest of the examples. I do have to edit this myself too :)

For my actual project, I ended up doing this:

https://github.com/JoelEllis/plotsirv/blob/0d6695c55bc164cca3f1fb5c61a691b9c8a7ac56/src/main.rs#L12-L60

It'd be my preference to have this in offline mode too. If that's OK with @rj00a?

qualterz: Also, I had an idea of having an environment variable `VALENCE_MODE` that has possible values `production` and `development`, so in development environment offline mode would be enabled and things like encryption would be disabled by default.

JoelEllis: Yeah, that's a pretty good idea, but not in scope for this PR. Make an issue or a discord thread? :slightly_smiling_face:

qualterz: Ah, I prefer not to use Discord, but feel free to leave this idea there. Also, I need to think about this solution more thoroughly before submit the feature request.

JoelEllis: IMO you should create one as soon as you can - either place is good for letting other people throw their ideas into the ring and then come up with a good solution!

rj00a: > It'd be my preference to have this in offline mode too. If that's OK with @rj00a?

Yeah you can go ahead and change it.

JadedBlueEyes:
Everything should be good to merge now.

dyc3 (approved):
LGTM

Issue #199, opened 2023-01-26
Dividing Clients and Entities into smaller components


rj00a:
Brainstorming some ideas for how the monolithic `Client` and `McEntity` components could be split up into a set of smaller components. Here are some reasons you would want to do this:
- Improve query performance by increasing parallelism and useful data in the CPU cache.
- More modular design can split code into smaller understandable pieces.
- Integrate with Bevy's change detection.
- Share data between client and Minecraft entity when they share the same ECS entity. (Typically the client will have a player entity associated with it).
- Save memory by spawning entities without unused components.

Some reasons you wouldn't want to do this:
- Many corner cases and interactions that are not accounted for can lead to bugs.
- Potentially confusing and unintuitive.
- Invalid states are very much representable.

---

| Component Name | Description | Effect When Added | Effect When Removed | Effect When Modified |
| -------------- | ----------- | -------------------------------- | ------------------------------------ | ---------------------- |
| `Despawned` | Any ECS Entities with this component are despawned at the end of the tick. This gives Valence an opportunity to perform deinitialization of components that can't be removed explicitly. In particular, `McEntity` and its other components. |  |  |
| `Client` | The main client component. Contains the TCP connection and packet buffers. Also contains immutable info such as username, UUID, and IP among other things.| Joins the client (Should only be done by Valence). Maybe we want to allow attaching `Client` to existing entities somehow? | No effect, but the client is disconnected when dropped. Removing only this component from an `McEntity` will allow you to take control. |  |
| `Disconnected` | Marker component to indicate that a client is disconnected. | | |
| `Location` | Contains the instance an Entity or Client is inside. Also contains the instance from the previous tick.  |  | Disallowed at least for entities because we need it to clean up the instance. | Changes the instance a client or entity is in. Makes clients respawn even if `old_instance == new_instance`. Can be used to respawn the client after death. |
| `Position` | The XYZ coordinates (old and new) of a client or entity. Automatically updated by Valence as a client moves around. | | Disallowed at least for entities because we need it to clean up the instance. | Teleports clients and entities when modified |
| `Direction` | Contains the pitch/yaw pair of an entity or client. Automatically updated by clients as they look around. | | Disallowed(?) | Sets the pitch/yaw of entities and teleports clients to set their pitch/yaw. Can be combined with `Position` modifications to form a single teleport. |
[...]

PR #252, opened 2023-02-20
Implement Drop Item Events

dyc3:
## Description

Dropping items is heavily coupled to inventories. Clients also predict state changes when they try to drop items, so we need to be able to replicate that change in order to stay in sync.

This will also remove `DropItem` events in favor of just `DropItemStack` events. Having 2 event streams that basically mean the same thing seems verbose and error prone.

As of right now, these changes require the event loop to have a reference to the client's inventory. This seems like something we are going to need to do a lot more of to complete #199.

## Test Plan

<details>

<summary>Playground</summary>

```rust
use tracing::info;
use valence::client::event::{default_event_handler, DropItemStack};
use valence::prelude::*;

#[allow(unused_imports)]
use crate::extras::*;

const SPAWN_Y: i32 = 64;

    app.add_plugin(ServerPlugin::new(()).with_connection_mode(ConnectionMode::Offline))
        .add_system_to_stage(EventLoop, default_event_handler)
        .add_startup_system(setup)
        .add_system(init_clients)
        .add_system(toggle_gamemode_on_sneak)
        .add_system(drop_items);
}

fn setup(world: &mut World) {
    let mut instance = world
        .resource::<Server>()
        .new_instance(DimensionId::default());

    for z in -5..5 {
        for x in -5..5 {
        }
    }

    for z in -25..25 {
        for x in -25..25 {
        }
    }

    world.spawn(instance);
}

fn init_clients(
    mut clients: Query<&mut Client, Added<Client>>,
) {
    let instance = instances.get_single().unwrap();

    for mut client in &mut clients {
        client.set_position([0.5, SPAWN_Y as f64 + 1.0, 0.5]);
        client.set_instance(instance);
        client.set_game_mode(GameMode::Creative);
    }
}

fn drop_items(clients: Query<&Client>, mut drop_stack_events: EventReader<DropItemStack>) {
    if drop_stack_events.is_empty() {
        return;
    }

    for event in drop_stack_events.iter() {
        info!("drop stack: {:?}", event);
    }
}

```

</details>

Steps:
1. `cargo test -p valence --tests`
2. Run playground `cargo run -p playground`
3. Open creative menu
4. Pick an item and click to drop it outside of the creative menu
5. Pick an entire stack of an item, place it in your hotbar
6. Hover over the item, press your drop key to drop an item from the stack
7. Press shift to switch to survival
8. Select the item stack in your hotbar, press your drop key to drop an item from the stack
9. Open your inventory, grab the stack, hover outside the window and click to drop the entire stack
10. Grab another stack from creative, place it in your hotbar
11. switch back to survival, select the stack, and press your control + drop key to drop the entire stack
12. For each item you dropped, you should see a log message with the event

### DIFF
--- a/crates/playground/src/extras.rs
+++ b/crates/playground/src/extras.rs
@@ -1,2 +1,24 @@
 //! Put stuff in here if you find that you have to write the same code for
 //! multiple playgrounds.
+
+use valence::client::event::StartSneaking;
+use valence::prelude::*;
+
+/// Toggles client's game mode between survival and creative when they start
+/// sneaking.
+pub fn toggle_gamemode_on_sneak(
+    mut clients: Query<&mut Client>,
+    mut events: EventReader<StartSneaking>,
+) {
+    for event in events.iter() {
+        let Ok(mut client) = clients.get_component_mut::<Client>(event.client) else {
+            continue;
+        };
+        let mode = client.game_mode();
+        client.set_game_mode(match mode {
+            GameMode::Survival => GameMode::Creative,
+            GameMode::Creative => GameMode::Survival,
+            _ => GameMode::Creative,
+        });
+    }
+}
--- a/crates/playground/src/main.rs
+++ b/crates/playground/src/main.rs
@@ -1,5 +1,6 @@
 use valence::bevy_app::App;
 
+#[allow(dead_code)]
 mod extras;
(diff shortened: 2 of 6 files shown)


rj00a (commented):
👍🏻

PR #253, opened 2023-02-20
Refactor `valence_protocol`

rj00a:
## Description

- Remove duplicate packet definitions by using `Cow`.
- Rename packets to match yarn mappings.
- Remove some top-level re-exports.
- Move every packet into its own module for consistency.
  - Move packet-specific types from the `types` module into the appropriate packet module.
- Remove internal use of `EncodePacket`/`DecodePacket` derives and move packet identification to `packet_group`. This can be done because there are no duplicate packets anymore.
- Simplify some events.

In a future PR I plan to clean things up further by properly bounding packet data (to prevent DoS exploits) and fixing any remaining inconsistencies with the game's packet definitions.

## Test Plan

Behavior of `valence_protocol` should be the same.

Steps:
1. Use the packet inspector against the vanilla server to ensure packet behavior has not changed.
2. Run the examples.
3. Run `valence_stresser`.


rj00a:
I have no idea why the tests are failing on macos.

PR #255, opened 2023-02-21
Provide API for getting the wall variant of blocks

JadedBlueEyes:
## Description

This adds an API for getting the wall / attachable variant of a block. 
Unlike #117 it does not add convenience methods for orientation, and it does not modify the example to use this code.

## Test Plan

The extractor is not tested. 
The API has some tests in the test module.

#### Related

#115 - it does not fix this, as it doesn't modify the example, but it could lead to it.

### DIFF
--- a/extractor/src/main/java/rs/valence/extractor/extractors/Blocks.java
+++ b/extractor/src/main/java/rs/valence/extractor/extractors/Blocks.java
@@ -4,9 +4,11 @@ import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import net.minecraft.registry.Registries;
+import net.minecraft.item.VerticallyAttachableBlockItem;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.EmptyBlockView;
 import rs.valence.extractor.Main;
+import rs.valence.extractor.mixin.ExposeWallBlock;
 
 import java.util.LinkedHashMap;
 import java.util.Locale;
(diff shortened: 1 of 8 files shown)


extractor/src/main/java/rs/valence/extractor/extractors/Blocks.java
```diff
@@ -37,6 +39,13 @@ public JsonElement extract() {
             blockJson.addProperty("translation_key", block.getTranslationKey());
             blockJson.addProperty("item_id", Registries.ITEM.getRawId(block.asItem()));
 
+            if (block.asItem() instanceof VerticallyAttachableBlockItem wsbItem) {
+                if (wsbItem.getBlock() == block) {
+                    var wallBlock = ((ExposeWallBlock) wsbItem).getWallBlock();
+                    blockJson.addProperty("wall_variant_id", Registries.BLOCK.getRawId(wallBlock));
```
JadedBlueEyes, line 45: I'm not sure whether to call it wall variant or something from `VerticallyAttachableBlockItem`.

In the original commit, the name of the import was `WallStandingBlockItem`, but this didn't work so I had to change this - did the mappings change?

rj00a:
Looks good to me.

> https://github.com/valence-rs/valence/issues/115 - it does not fix this, as it doesn't modify the example, but it could lead to it.

Generally these things are known as "placement rules" which can become [complicated](https://github.com/Minestom/Minestom/tree/master/src/main/java/net/minestom/server/instance/block/rule/vanilla) from what I understand. We will probably need to provide a holistic solution.

extractor/src/main/java/rs/valence/extractor/extractors/Blocks.java
rj00a: There might have been some tweaks in a later version of the yarn mappings.

rj00a: Actually `wall_block_id` is probably a better name for this.

JadedBlueEyes:
> Looks good to me.
> 
> > #115 - it does not fix this, as it doesn't modify the example, but it could lead to it.
> 
> Generally these things are known as "placement rules" which can become [complicated](https://github.com/Minestom/Minestom/tree/master/src/main/java/net/minestom/server/instance/block/rule/vanilla) from what I understand. We will probably need to provide a holistic solution.

I've just been slowly adding to my [ball of mud](https://github.com/JadedBlueEyes/plotsirv/blob/49d03925b2b6f11963aedeff22e2abc907bfb877/src/main.rs#L174) :laughing: 
Things can only go downhill from here!

I think things like stair bending and fence connecting happen in a [block update](https://minecraft.fandom.com/wiki/Block_update) and can be ignored until then, as long as orientation, etc is right. I'm not sure how you would actually implement block updates in Bevy though (another stage?)

extractor/src/main/java/rs/valence/extractor/extractors/Blocks.java
JadedBlueEyes: Renamed the public function, although I've left the internals alone.

PR #258, opened 2023-02-24
Changed `setup()` to not be an exclusive system

Supermath101:
## Description

In the [examples of the `valence` crate](https://github.com/valence-rs/valence/tree/main/crates/valence/examples), the `setup()` functions are currently [exclusive systems](https://bevy-cheatbook.github.io/programming/exclusive.html). My reasoning for making these a standard system is twofold:
- Exclusive systems can be considered an advanced Bevy topic. I don't think you should be using advanced concepts in examples meant to be an introduction to the way this project/framework is used.
- In instances where developers try to use these examples, adding another startup system would result is running the systems sequentially. That leaves performance on the table (albeit only during startup).

## Test Plan (or lack thereof)

I've chosen to _not_ include any tests besides the ones already present, because I feel that it's unnecessary. My justification is as follows: Since `setup()` was the only _startup_ system in these examples (that I'm aware of), the changes are functionally equivalent to anything before. This is because startup systems run _before_ all non-startup systems. See [this docs.rs section](https://docs.rs/bevy/0.9.1/bevy/app/struct.App.html#the-stages) for some information on startup system scheduling.

### DIFF
--- a/crates/playground/src/playground.template.rs
+++ b/crates/playground/src/playground.template.rs
@@ -30,7 +30,7 @@ fn setup(mut commands: Commands, server: Res<Server>) {
         }
     }
 
-    world.spawn(instance);
+    commands.spawn(instance);
 }
 
 fn init_clients(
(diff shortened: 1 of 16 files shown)


dyc3:
FYI, a test plan is not "What tests do I need to write?", think of it as "What should the maintainers do to make sure this change works as expected?"

Issue #210, opened 2023-02-04
[ECS] Optional component for `Instance` used to control time, weather, world border, etc.


rj00a:
It would be convenient if `Instance` had some optional components for controlling state which is typically expected to be global to the instance. This includes the time of day, whether or not time is paused (`doDaylightCycle`), The world border for the instance, the weather (raining or thundering), etc.

My current thinking is that we can have the following components:
- `TimeOfDay`
- `Weather`
- `WorldBorder`

rj00a:
> Might be more convenient to make these components smaller for the purpose of change detection.

Yes, each field that needs change detection should be its own component.

dyc3:
I'd consider that to be vanilla functionality, which would be out of scope for valence. Do gamerules get conveyed to clients and affect them in any way? If not, gamerules are definitely out of scope.

arseeeeN:
That makes sense, so the developers that use Valence would handle the updating of the time of day themselves? Including sending that information to the players of course.
Should the time of day be a field of the `InstanceInfo` struct or a component that could be spawned alongside the entity of the Instance component?

dyc3:
Valence would be the one sending packets. As the issue says, the desired solution is a `TimeOfDay` component that goes on the entity that has the `Instance`. A system inside valence would then be able to handle sending those packets.

MrlnHi:
> I'd consider that to be vanilla functionality, which would be out of scope for valence. Do gamerules get conveyed to clients and affect them in any way? If not, gamerules are definitely out of scope.

reducedDebugScreen or however it's called gets conveyed to clients (and actually makes a difference). Other than that I can't think of any tho

tachibanayui:
Is anyone already working on time of day? I want to take a shot at implementing it.

arseeeeN:
I initially wanted to work on it but couldn't find the time, so feel free to take over.

1bacon:
> > I'd consider that to be vanilla functionality, which would be out of scope for valence. Do gamerules get conveyed to clients and affect them in any way? If not, gamerules are definitely out of scope.
> 
> reducedDebugScreen or however it's called gets conveyed to clients (and actually makes a difference). Other than that I can't think of any tho

For completeness:
Only `reducedDebugInfo`, `doLimitedCrafting` and `doImmediateRespawn` are sent to the Client.

All others are checked by the server internally.

PR #106, opened 2022-10-09
Implemented Weather

Tert0:
This PR implements Weather.
# Issue
Closes #105

Im currently not sure how design the api for weather.
**Current State** (by Game Event)
```rust
client.set_raining(true);
client.set_rain_level(0.5); // does not depend on set_raining
client.set_thunder_level(0.3); // does depend on set_raining (or set_rain_level)
```
**Idea 2** (without set_raining)
```rust
client.set_rain_level(0.5);
client.set_thunder_level(0.3); // does depend on set_rain_level
```
**Idea 3**
```rust
client.set_rain(0.5);
client.set_thunder(0.3); // does depend on set_rain_level
client.set_thunder(0.5/*rain level*/, 0.3/*thunder level*/); // does depend on set_rain_level
```

**Idea 4**
```rust
client.set_weather(Weather::Rain);
client.set_weather(Weather::Rain, 0.5);
client.set_weather(Weather::Thunder); // depends on enabling the rain first
client.set_weather(Weather::Thunder, 0.3); // depends on enabling the rain first
```
I would be happy if someone would give their opinion on this.

PR #260, opened 2023-02-25
Weather implementation

qualterz:
## Description

An implementation of basic weather systems.

The weather component attached to a world instance would be handled for all clients, except those that have their own weather component; these clients would be handled separately.

## Test Plan

<details>

<summary>Playground</summary>

```rust
fn handle_command_events(
    mut exec_cmds: EventReader<CommandExecution>,
    mut commands: Commands,
) {
    for cmd in exec_cmds.iter() {
        let msg = cmd.command.to_string();
        let ent = instances.single();

        match msg.as_str() {
            "ar" => {
                commands.entity(ent).insert(Rain(WEATHER_LEVEL.end));
            }
            "rr" => {
                commands.entity(ent).remove::<Rain>();
            }
            "at" => {
                commands.entity(ent).insert(Thunder(WEATHER_LEVEL.end));
            }
            "rt" => {
                commands.entity(ent).remove::<Thunder>();
            }
            _ => (),
        };
    }
}
```

</details>

Steps:
1. Run `cargo test --package valence --lib -- weather::test`

#### Related
Part of #210
Past approach #106


crates/valence/src/instance.rs
```diff
@@ -431,6 +434,46 @@ impl Instance {
...
+            kind: GameEventKind::EndRaining,
+            value: f32::default(),
+        });
+    }
+
+    /// Sets a level of the rain in the instance.
+    pub fn set_raining_level(&mut self, level: f32) {
+        self.write_packet(&GameStateChangeS2c {
+            kind: GameEventKind::RainLevelChange,
+            value: num::clamp(
```
qualterz, line 458: I have doubts about should the value be clamped or not.

crates/valence/src/weather.rs
```diff
@@ -0,0 +1,19 @@
...
+    /// Should be between [`MIN_WEATHER_LEVEL`] and [`MAX_WEATHER_LEVEL`].
+    ///
+    /// The [`None`] value means no raining event.
+    pub raining: Option<f32>,
+    /// Contains the thunder level.
+    /// Should be between [`MIN_WEATHER_LEVEL`] and [`MAX_WEATHER_LEVEL`].
+    ///
+    /// The [`None`] value means no thunder.
+    pub thunder: Option<f32>,
+}
```
rj00a, line 19: Can you make this module a child of the `instance` module?

crates/valence/src/instance.rs
qualterz: Also there is a core `clamp` function implementation, which should be used?

crates/valence/src/instance/weather.rs
```diff
@@ -0,0 +1,337 @@
+use bevy_ecs::prelude::*;
+use bevy_ecs::schedule::SystemConfigs;
+use valence_protocol::packet::s2c::play::game_state_change::GameEventKind;
+use valence_protocol::packet::s2c::play::GameStateChangeS2c;
+
+use super::Instance;
+use crate::prelude::*;
+
+pub const WEATHER_LEVEL_MIN: f32 = 0_f32;
+pub const WEATHER_LEVEL_MAX: f32 = 1_f32;
```
dyc3, line 10: Can this be a `Range` instead?

dyc3 (commented):
Seems good. Fix the merge conflict and we can merge it

crates/valence/src/instance.rs
```diff
@@ -28,6 +29,7 @@ use crate::view::ChunkPos;
 mod chunk;
 mod chunk_entry;
 mod paletted_container;
+pub mod weather;
```
rj00a, line 32: Keeping the weather module in the `instance` module doesn't really fit since it works on clients too. So it's original place is probably fine.

crates/valence/src/weather.rs
```diff
@@ -0,0 +1,338 @@
...
+
+use bevy_ecs::prelude::*;
+use bevy_ecs::schedule::SystemConfigs;
+use valence_protocol::packet::s2c::play::game_state_change::GameEventKind;
+use valence_protocol::packet::s2c::play::GameStateChangeS2c;
+
+use crate::packet::WritePacket;
+use crate::prelude::*;
+
+pub const WEATHER_LEVEL_RANGE: Range<f32> = 0_f32..1_f32;
```
qualterz, line 11: Is not the name of this constant too wordy? Maybe just `WEATHER_LEVEL` would be better?

crates/valence/src/weather.rs
```diff
@@ -0,0 +1,338 @@
...
+        instance.set_weather(weather);
+    });
+}
+
+fn handle_weather_for_joined_player(
+    mut clients: Query<&mut Client, Added<Client>>,
+    weathers: Query<&Weather, With<Instance>>,
+) {
+    clients.for_each_mut(|mut client| {
+        if let Ok(weather) = weathers.get_single() {
```
qualterz, line 150: How to get the weather component of the instance where the client is joined?

qualterz: Alright, I got the instance entity from `McEntity` using `Query<(&mut Client, &McEntity), Added<Client>>`, and I used the instance entity to get the weather component from `Query<&Weather, With<Instance>>`.

rj00a: You'll want to use the `Location` component instead.

crates/valence/src/weather.rs
```diff
@@ -0,0 +1,337 @@
...
+            client.end_raining();
+        }
+    })
+}
+
+fn handle_weather_change_per_client(mut query: Query<(&mut Client, &Weather), Changed<Weather>>) {
+    query.for_each_mut(|(mut client, weather)| {
+        client.set_weather(weather);
+    });
+}
```
rj00a, line 177: What happens if an entity has both client components and an `Instance` component? Changes to `Weather` will be sent twice -- once for the client and once for the instance. Looks like you'll want to add `Without<Instance>` to these systems.

crates/valence/src/weather.rs
```diff
@@ -0,0 +1,337 @@
...
+        handle_weather_for_joined_player,
+        handle_weather_begin_per_instance.before(handle_weather_change_per_instance),
+        handle_weather_change_per_instance.before(handle_weather_end_per_instance),
+        handle_weather_end_per_instance,
+        handle_weather_begin_per_client.before(handle_weather_change_per_client),
+        handle_weather_change_per_client.before(handle_weather_end_per_client),
+        handle_weather_end_per_client,
+    )
+        .into_configs()
+}
```
rj00a, line 190: Following #289, these systems should be wrapped up in a `Plugin`.

qualterz:
My final step is to document weather module.

qualterz:
Something weird with tests execution, in some cases they may fail. I'll just increase the number of ticks and cross my fingers that this will be a permanent fix.

qualterz:
I don't know should the framework handle the `Thunder` component removal or not. I think would be better to leave this to a user and the weather components would be just an ECS API for weather events emitting.

qualterz:
During my play tests, I preferred to leave the `Thunder` component handling for ergonomics.

qualterz:
Not sure about scheduling but tests should pass. If there's no problems, the pull request can be merged.

dyc3 (approved):

PR #261, opened 2023-02-25
Replace `EncodePacket` and `DecodePacket` with `Packet`

rj00a:
## Description

Combines the `EncodePacket` and `DecodePacket` trait into a single `Packet` trait. This makes `valence_protocol` simpler and easier to use. This can be done because all packets were made to be bidirectional in #253.

Additionally, a `packet_id` method has been added. This should help with #238.

## Test Plan

Steps:
1. Run examples, packet_inspector, etc. Behavior should be the same.

### DIFF
--- a/crates/valence/src/unit_test/util.rs
+++ b/crates/valence/src/unit_test/util.rs
@@ -135,7 +135,7 @@ impl MockClientHelper {
 
     /// Inject a packet to be treated as a packet inbound to the server. Panics
     /// if the packet cannot be sent.
-    pub fn send<'a>(&mut self, packet: &(impl Packet<'a>)) {
+    pub fn send<'a>(&mut self, packet: &impl Packet<'a>) {
         self.enc
             .append_packet(packet)
             .expect("failed to encode packet");
--- a/crates/valence/src/unit_test/util.rs
+++ b/crates/valence/src/unit_test/util.rs
@@ -6,7 +6,7 @@ use bytes::BytesMut;
 use valence_protocol::codec::{PacketDecoder, PacketEncoder};
 use valence_protocol::packet::S2cPlayPacket;
 use valence_protocol::username::Username;
-use valence_protocol::EncodePacket;
+use valence_protocol::Packet;
 
 use crate::client::{Client, ClientConnection};
 use crate::config::{ConnectionMode, ServerPlugin};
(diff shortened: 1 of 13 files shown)


dyc3 (approved):

crates/valence_protocol_macros/src/lib.rs
```diff
@@ -8,13 +8,14 @@ use proc_macro::TokenStream as StdTokenStream;
 use proc_macro2::TokenStream;
```
JadedBlueEyes, line 8: Just above here - 

```rust
//! This crate provides derive macros for [`Encode`], [`Decode`],
//! [`EncodePacket`], and [`DecodePacket`].
//! It also provides the procedural macro [`ident_str!`]
```

Broken links

crates/valence_protocol/src/lib.rs
```diff
@@ -277,111 +278,75 @@ pub trait Decode<'a>: Sized {
 ///
 /// [macro]: valence_protocol_macros::DecodePacket
```
JadedBlueEyes, line 279: Broken link

crates/valence/src/unit_test/util.rs
```diff
@@ -135,7 +135,7 @@ impl MockClientHelper {
 
     /// Inject a packet to be treated as a packet inbound to the server. Panics
     /// if the packet cannot be sent.
-    pub fn send(&mut self, packet: &(impl EncodePacket + ?Sized)) {
+    pub fn send<'a>(&mut self, packet: &(impl Packet<'a>)) {
```
JadedBlueEyes, line 138: unnecessary parentheses around type

crates/valence_protocol/src/lib.rs
```diff
@@ -396,7 +361,7 @@ mod derive_tests {
     #[allow(unconditional_recursion)]
     fn has_impls<'a, T>()
```
JadedBlueEyes, line 362: type parameter goes unused in function definition

JadedBlueEyes (changes requested):
Hey, I've just gone through and ran clippy + tests. I've marked out a few changes, and the playground is broken from a previous PR, but otherwise LGTM! :+1: :smile:

crates/valence_protocol/src/lib.rs
rj00a: The type param asserts that the input types have the expected bounds.

PR #263, opened 2023-03-02
Support for schematics file format

### LINKED ISSUE #193: Adding support for schematics file format
## Description :

.schem file format, is meant to store part of the minecraft world as file and can be use in a large variety of community programs.

[Specification link](https://github.com/SpongePowered/Schematic-Specification/blob/master/versions/schematic-3.md)

## Usability : 

Adding support for this file format could be used to : 
- Let the server dev enhance his world generation by using premade structure
- Let minigame creator use premade arena
- Simplify the creation of tools for users to copy as or paste shematics 

## Implementation 

For a simple implementation, I would see something like : 
```rust
let schem = Schematics::load("<path>");

instance.paste_schem(schem: Schematics, basePos: Vec3, option: PasteOption);

let copy_schem = instance.copy_schem(pos1 : Vec3, pos2 : Vec3, option: CopyOption);

Schematics::save(schem: Schematics, "<path>");
```

The option would be like : 
- paste_air_block : bool
- copy_entities : bool
- ...

MrlnHi:
## Description
- Added auxiliary crate `valence_schem` that supports copying and pasting schematics to an `Instance` and also allows for parsing them from a `Compound` and serializing it back to one

## Test Plan

The examples `schem_loading` and `schem_saving` are sufficient for testing. Also, as for a third party created schematic, I used [SimpleSpongeSchematics](https://github.com/masagameplay/SimpleSpongeSchematics) to do so.

Steps:
1. Create a Schematic file using the sponge mod above
2. Start the `schem_loading` with the path of your schematic, like so: `cargo r --example schem_loading -- "my/path/to/something_cool.schem"`

#### Related

Closes #193


qualterz:
I think would be good to have an example schematic in `valence/assets` directory.

MrlnHi:
> I think would be good to have an example schematic in `valence/assets` directory.

I agree 👍. Will do once I get back at it

crates/valence_schem/examples/schem_loading.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    match Schematic::load(path) {
+        Ok(schem) => {
+            schem.paste(&mut instance, SPAWN_POS, |_| BiomeId::default());
+        }
+        Err(err) => {
+            eprintln!("Error loading schematic: {err}");
+            world.send_event(AppExit);
+        }
+    }
```
dyc3, line 55: It would be nice to include this kind of thing in doc comments so it's visible to people looking at the docs.

dyc3 (commented):
Haven't given it a super hard look yet.

crates/valence_schem/examples/schem_loading.rs
```diff
@@ -0,0 +1,83 @@
...
+        .add_plugin(ServerPlugin::new(()))
+        .add_system_to_stage(EventLoop, default_event_handler)
+        .add_system_set(PlayerList::default_system_set())
+        .add_startup_system(setup)
+        .add_system(init_clients)
+        .add_system(despawn_disconnected_clients)
+        .run();
+}
+
+fn setup(world: &mut World) {
```
rj00a, line 32: This should not be an exclusive system (#258).

MrlnHi:
Question: do we want to support all spec versions? (1-3) 
I don't think we should for serialization, but for deserialization probably. The deserialization would then check the version and act accordingly.
So basically by deserializing an older schematic and serializing it again, we would upgrade it to version 3. I don't think we should support DataVersion (meaning knowing how to handle 1.8 schematics with blocks like minecraft:planks) but probably the Version property. What do you think?

crates/valence_schem/examples/schem_loading.rs
MrlnHi: What exactly do you think should be in doc comments?

rj00a:
> I don't think we should support DataVersion (meaning knowing how to handle 1.8 schematics with blocks like minecraft:planks) but probably the Version property. What do you think?

That seems like the most reasonable thing to do now.

crates/valence_schem/examples/schem_loading.rs
dyc3: A short explanation of what schematics are, examples for basic usage with a little context. You can look at the doc comments for the inventory module for reference.

MrlnHi:
@rj00a ready for review

rj00a:
Currently working on a redesign of the `Instance` API to fix a number of issues. Once that's mostly finished I can take a closer look at this and see what impact it has.

rj00a:
I've (finally) finished the redesign in #402, so we can move forward with this.

MrlnHi:
> I've (finally) finished the redesign in #402, so we can move forward with this.

Ok, I updated the branch to main

dyc3 (commented):
I gave it a quick look over again. This might need to be updated to reference chunk layers instead of instances.

crates/valence_generated/src/block.rs
dyc3, line 1: @rj00a The changes in this file don't seem right. `valence_generated` should be completely codegen right?

MrlnHi: I guess so, but I can't implement the trait anywhere else, right?

crates/valence_schem/src/lib.rs
```diff
@@ -0,0 +1,1112 @@
...
+impl<'a> Write for VarIntWriteWrapper<'a> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.0.extend(buf.iter().map(|byte| *byte as i8));
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
```
rj00a, line 224: I'm going to expose some functions in `valence_nbt` that should make these wrappers unnecessary.

crates/valence_generated/src/block.rs
```diff
@@ -48,6 +50,53 @@ fn fmt_block_state(bs: BlockState, f: &mut fmt::Formatter) -> fmt::Result {
...
+    #[error("unknown prop name '{0}'")]
+    UnknownPropName(String),
+    #[error("unknown prop value '{0}'")]
+    UnknownPropValue(String),
+}
+
+impl FromStr for BlockState {
+    type Err = ParseBlockStateError;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
```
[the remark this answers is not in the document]
MrlnHi: The problem that it's not necessarily `Ident`s which we get in here, but also the block state props
(so like `fence[west=true]` which wouldn't parse as `Ident`, right?)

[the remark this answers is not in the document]
MrlnHi: > We should make this tolerant of the minecraft: namespace in the input.
as for this: I believe that would be a better fit for the `BlockKind::from_str` method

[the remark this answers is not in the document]
rj00a: The whole thing isn't an ident, but presumably the string before the `[` is, which you have access to in the `kind` variable.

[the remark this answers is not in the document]
MrlnHi: > The whole thing isn't an ident, but presumably the string before the `[` is, which you have access to in the `kind` variable.

ahh right I misread part of your first message :smile: I see know

crates/valence_generated/src/block.rs
```diff
@@ -48,6 +50,53 @@ fn fmt_block_state(bs: BlockState, f: &mut fmt::Formatter) -> fmt::Result {
...
+        let state = match s.split_once('[') {
+            Some((kind, props)) => {
+                let Some(kind) = BlockKind::from_str(kind) else {
+                    return Err(ParseBlockStateError::UnknownBlockKind(kind.to_string()));
+                };
+                props[..props.len() - 1]
```
rj00a, line 74: This will panic if `props` is empty.

PR #265, opened 2023-03-03
Update to Bevy 0.10

### LINKED ISSUE #264: Update to Bevy 0.10

Bevy 0.10 (unfinished as of writing) includes some very large changes to the scheduling API. These changes should help make #199 and #257 more achievable.

---

- [ ] https://github.com/bevyengine/bevy/issues/7880

Allowing system sets to be nested helps when there are a large number of systems that need to be scheduled. Sets of systems (e.g. all the systems needed to update inventories) can be encapsulated together in a sane way without having to create an explicit label for it.

Use a recent commit on the main branch of bevy or 0.10 when that comes out.

### Additional context

_No response_

rj00a:
## Description

Fix #264

Bevy version is pinned to a recent commit on the main branch until 0.10 is released.

## Test Plan

Run examples and tests. Behavior should be the same.

scheduling is a bit tricky so I may have made some subtle mistakes.

### DIFF
--- a/crates/playground/src/playground.template.rs
+++ b/crates/playground/src/playground.template.rs
@@ -9,7 +9,7 @@ const SPAWN_Y: i32 = 64;
 
 pub fn build_app(app: &mut App) {
     app.add_plugin(ServerPlugin::new(()))
-        .add_system_to_stage(EventLoop, default_event_handler)
+        .add_system(default_event_handler.in_schedule(EventLoopSchedule))
         .add_startup_system(setup)
         .add_system(init_clients)
         .add_system(despawn_disconnected_clients);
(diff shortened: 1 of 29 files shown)


rj00a:
Hmm, not sure if it's possible to create a looping system set with 0.10. I need to recreate the behavior of the event loop which previously used run criteria to do this. Might have to reorganize things and use a dedicated `Schedule` for the event loop instead.

qualterz:
Here is also used hardcoded batch size:

https://github.com/valence-rs/valence/blob/cc2571d7e618172ea719430d62a49d3207fe64cc/crates/valence/examples/chat.rs#L68

Should it be changed to `par_iter_mut().for_each_mut()`?

rj00a:
I'm still updating the examples, but it probably shouldn't be running in parallel to begin with.

PR #268, opened 2023-03-04
Inventory module docs and more helper functions

dyc3:
## Description

This mostly adds docs, but it also adds (and documents) some new helper functions:

- `set_title()` - Allows us to put `must_use` on `replace_title`
- `set_slot()` - Allows us to put `must_use` on `replace_slot`
- `set_slot_amount()` - Allows users to modify stack counts without cloning the entire stack and replacing it. Useful if the stack has a lot of NBT data.
- `first_empty_slot()`
- `first_empty_slot_in()` - Useful, trivial to provide

## Test Plan

<details>

<summary>Playground</summary>

```rust
N/A
```

</details>

Steps:
1. `cargo test -p valence --doc`

### DIFF
--- a/crates/valence/src/inventory.rs
+++ b/crates/valence/src/inventory.rs
@@ -67,6 +67,14 @@ impl Inventory {
             .as_ref()
     }
 
+    /// Sets the slot at the given index to the given item stack.
+    /// ```
+    /// # use valence::prelude::*;
+    /// let mut inv = Inventory::new(InventoryKind::Generic9x1);
+    /// inv.set_slot(0, ItemStack::new(ItemKind::Diamond, 1, None));
+    /// let old = inv.replace_slot(0, ItemStack::new(ItemKind::IronIngot, 1, None));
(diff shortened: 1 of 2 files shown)


crates/valence/src/inventory.rs
```diff
@@ -67,7 +96,37 @@ impl Inventory {
...
+    /// See also [`Inventory::replace_slot`].
+    ///
+    /// ```
+    /// # use valence::prelude::*;
+    /// let mut inv = Inventory::new(InventoryKind::Generic9x1);
+    /// inv.set_slot(0, ItemStack::new(ItemKind::Diamond, 1, None));
+    /// assert_eq!(inv.slot(0).unwrap().item, ItemKind::Diamond);
+    /// ```
+    #[track_caller]
+    #[allow(unused_must_use)]
```
rj00a, line 110: It's more idiomatic to use `let _ = ...;` to disable the warning.

PR #279, opened 2023-03-08
Replaced `valence_protocol` and `valence_nbt` imports with `valence` ones

wjwzpeajhc:
## Description

This PR replace `valence_protocol` and `valence_nbt` imports in `crates/valence/examples` with `valence` ones so you don't have to change imports or import the crates when you copy the examples and have only `valence` as a dependency.


AviiNL:
This may have been merged a touch too hastily, there's a few more examples that reference `valence_protocol` instead of `valence::protocol`.

- `parkour.rs`
- `resource_pack.rs`
- `text.rs`

PR #293, opened 2023-03-19
Rough click slot packet validation

### LINKED ISSUE #292: Inventory Packet Validation

At the moment, we trust the client a little too much when it comes to modifying inventories. For example, a client can send a click slot packet that just adds any item to any slot out of thin air.

There are 2 types of assertions we need to make on incoming modifications:
1. plain invalid info that shouldn't be processed ever
2. making sure items aren't being duplicated/spawned/deleted, aka conservation of mass

Type 1 validations:
- [x] slot ids are in-bounds
- [x] item stack amounts are in-bounds
- [ ] ~~some slots are "outbound only" eg. crafting result slots~~ decided not to do this because it's not that useful and pretty hard to assert for

Type 2 validations:
- [x] while the cursor is holding an item
    - [x] and clicking an empty slot, the item should go into the clicked slot
    - [x] and clicking a filled slot, 
        - [x] if the item is the same type, and same nbt data, an amount in the cursor item's stack be transferred to the slot's stack such that the resulting slot's stack reaches 64 (maybe make this configurable? some items can only stack to 1)
        - [x] otherwise, the slot and cursor item should swap
    - [x] and the user clicks and drags, the resulting differences in item stacks should add up to the correct amount
        - example: user is holding 5 diamonds, slot 1 is holding 1 diamond, slot 2 is empty. user clicks and drags the 5 item stack across slot 1 and 2. now the user should be holding 1 diamond, slot 1 should have 3 diamonds, slot 2 should have 2 diamonds
- [x] while the cursor is **not** holding an item
    - [x] and clicking an empty slot, nothing should happen
    - [x] and clicking a filled slot
       - [x] and pressing shift, using left click, the slot's stack should go directly to another slot (it doesn't matter which one)
       - [x] using right click, half the stack should be picked up (which slot gets the remainder if the amount in the stack is an odd number)
       - [x] otherwise, the slot's stack should be transferred to the cursor

The type 2 validations may be expensive, so ideally we should allow the user to disable it.

There's probably some more that I missed, but that should be enough to catch most cases.

We could also not do this, and defer this to end users, but I think that could end up being pretty error prone.

## Additional context

For #288 

https://wiki.vg/Protocol#Click_Container

dyc3:
## Description

This adds some validation for incoming inventory packets that makes it so that you can't just spawn items by sending malicious packets. It adds type 1 and type 2 validations as outlined in #292.

This also adds some new helpers, `InventoryWindow` and `InventoryWindowMut`.

fixes #292

<details>

<summary>Playground</summary>

```rust
use valence::client::{default_event_handler, despawn_disconnected_clients};
use valence::prelude::event::PlayerInteractBlock;
use valence::prelude::*;

#[allow(unused_imports)]
use crate::extras::*;

const SPAWN_Y: i32 = 64;
const CHEST_POS: [i32; 3] = [0, SPAWN_Y + 1, 3];

    app.add_plugin(ServerPlugin::new(()).with_connection_mode(ConnectionMode::Offline))
        .add_startup_system(setup)
        .add_system(default_event_handler.in_schedule(EventLoopSchedule))
        .add_system(init_clients)
        .add_systems((toggle_gamemode_on_sneak, open_chest).in_schedule(EventLoopSchedule));
}

fn setup(mut commands: Commands, server: Res<Server>) {
    let mut instance = server.new_instance(DimensionId::default());

    for z in -5..5 {
        for x in -5..5 {
        }
    }

    for z in -25..25 {
        for x in -25..25 {
        }
    }
    instance.set_block(CHEST_POS, BlockState::CHEST);

    commands.spawn(instance);

    let mut inventory = Inventory::new(InventoryKind::Generic9x3);
    inventory.set_slot(0, ItemStack::new(ItemKind::Apple, 100, None));
    inventory.set_slot(1, ItemStack::new(ItemKind::Diamond, 40, None));
    inventory.set_slot(2, ItemStack::new(ItemKind::Diamond, 30, None));
    commands.spawn(inventory);
}

fn init_clients(
    mut clients: Query<(&mut Position, &mut Location, &mut Inventory), Added<Client>>,
) {
    for (mut pos, mut loc, mut inv) in &mut clients {
        pos.0 = [0.5, SPAWN_Y as f64 + 1.0, 0.5].into();
        loc.0 = instances.single();

        inv.set_slot(24, ItemStack::new(ItemKind::Apple, 100, None));
        inv.set_slot(25, ItemStack::new(ItemKind::Apple, 10, None));
    }
}

// Add new systems here!

fn open_chest(
    mut commands: Commands,
    inventories: Query<Entity, (With<Inventory>, Without<Client>)>,
    mut events: EventReader<PlayerInteractBlock>,
) {
    let Ok(inventory) = inventories.get_single() else {
        return;
    };

    for event in events.iter() {
        if event.position != CHEST_POS.into() {
            continue;
        }
        let open_inventory = OpenInventory::new(inventory);
        commands.entity(event.client).insert(open_inventory);
    }
}

```

</details>

## Test Plan

Steps:
1. `cargo test`


crates/valence/src/inventory/validate.rs
```diff
@@ -0,0 +1,643 @@
+use valence_protocol::packet::c2s::play::click_slot::ClickMode;
+use valence_protocol::packet::c2s::play::ClickSlotC2s;
+
+use super::{Inventory, InventoryWindow};
+use crate::prelude::CursorItem;
+
+/// Validates a click slot packet enforcing that all fields are valid.
+pub(crate) fn validate_click_slot_impossible(
```
rj00a, line 8: Can this return an `anyhow::Result<()>` instead of `bool`? Would help with debugging. Same goes for `validate_click_slot_item_duplication`.

crates/valence/src/client/event.rs
```diff
@@ -852,7 +860,27 @@ fn handle_one_packet(
...
+                &q.inventory,
+                open_inv,
+                &q.cursor_item,
+            ) {
+                debug!(
+                    "client {:#?} click slot packet tried to incorrectly modify items: {:#?}",
+                    q.entity, p
+                );
+                return Ok(true);
+            }
```
rj00a, line 882: If a validation error does occur, the client can go out of sync. We probably need to send a `InventoryS2c` to fix this.

rj00a (commented):
Getting some validation failures when splitting stacks with right clicks and shift-clicking stacks to/from the chest in the chest example.

crates/valence/src/inventory/validate.rs
```diff
@@ -0,0 +1,643 @@
...
+
+    let max_slot = match open_inventory {
+        Some(inv) => inv.slot_count() + 36,
+        None => player_inventory.slot_count(),
+    };
+
+    // check all slot ids and item counts are valid
+    if !packet.slots.iter().all(|s| {
+        (0..=max_slot).contains(&(s.idx as u16))
+            && (0..=64).contains(&(s.item.as_ref().map(|i| i.count()).unwrap_or(0)))
```
rj00a, line 25: I see two problems here:
1. This assumes all items have a max stack size of 64.
2. This can't handle the situation where a stack with a weird stack count _already exists_ in the inventory. (Try giving yourself a stack with 100 items using `.set_count(100)`. Vanilla client seems to handle this just fine.)

crates/valence/src/inventory/validate.rs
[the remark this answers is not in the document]
dyc3: They're separate because I wanted to make the item duplication checks configurable by the end user for performance and i forgot to actually do that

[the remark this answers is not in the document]
rj00a: Also #296 will probably add the needed escape hatches.

dyc3:
Whoops, I gotta revert the changes to the chest example

crates/valence/src/inventory/validate.rs
```diff
@@ -0,0 +1,675 @@
...
+        None => player_inventory.slot_count(),
+    };
+
+    // check all slot ids and item counts are valid
+    ensure!(
+        packet.slots.iter().all(|s| {
+            (0..=max_slot).contains(&(s.idx as u16))
+                && s.item
+                    .as_ref()
+                    .map_or(true, |i| (1..=127).contains(&i.count()))
```
rj00a, line 32: The `1..=127` range fixes the issue with the weird stack counts, but now clients can stack items in ways they shouldn't.

You can check the max stack count with `ItemKind::max_stack`. For handling the weird counts, I think you need to check that the new stack does not have a count greater than `max(old_stack.item.max_stack(), old_stack.count())` or something like that.

rj00a:
I'm getting `Client state id mismatch, resyncing` when adding items to my inventory with code, switching to survival, and clicking on it. (But no validation errors when doing that with pickaxes? Not sure what's going on here).

dyc3:
> It turns out the client will actually add NBT data to clicked items in some circumstances.

Ugh, that's really annoying. Do we extract data for what nbt data an item is supposed to have?

rj00a:
> Do we extract data for what nbt data an item is supposed to have?

Not yet.

dyc3:
Would you be opposed to loosening the assertions on nbt data so we can defer this problem for later?

rj00a:
Still getting the `Client state id mismatch, resyncing` as mentioned above. (Not sure if you're finished with this)

rj00a:
Also doesn't look like the item count check is working properly.

```
client 2v1 invalid click slot packet: "invalid carried item count" ClickSlotC2s {
    window_id: 0,
    state_id: VarInt(
        2,
    ),
    slot_idx: 24,
    button: 0,
    mode: Click,
    slots: [
        Slot {
            idx: 24,
            item: None,
        },
    ],
    carried_item: Some(
        ItemStack {
            item: Apple,
            count: 77,
            nbt: None,
        },
    ),
}
```

(I added a stack of 77 apples to the inventory)

dyc3:
I'm not able to reproduce that. I am getting different issues though, like I can't manually set a stack of 100 apples, it just shows up as 64 in the client.

Can you list out all the steps to repro using the playground I added to the PR?

rj00a:
I Made an edit to the playground. When you click on the apple stacks in the player's inventory you'll see the issues manifest.

dyc3:
Ok, I fixed the issue you were talking about, but from now on I think we should consider behavior related to overfull stacks (eg stacks that have `amount > max stack`) to be undefined given that the notchian client is inconsistent in those scenarios.

rj00a:
> but from now on I think we should consider behavior related to overfull stacks (eg stacks that have amount > max stack) to be undefined given that the notchian client is inconsistent in those scenarios.

Yeah agree.

crates/valence_protocol/src/item.rs
```diff
@@ -16,8 +16,8 @@ pub struct ItemStack {
     pub nbt: Option<Compound>,
 }
 
-const STACK_MIN: u8 = 1;
-const STACK_MAX: u8 = 127;
+pub const STACK_MIN: u8 = 1;
+pub const STACK_MAX: u8 = 127;
```
rj00a, line 20: These could be associated constants on `ItemStack` for convenience.

rj00a (commented):
- Still seeing the desync when clicking on one of the apple stacks in the player inventory for the fist time. It only happens when the chest is not open. Guessing a variable wasn't initialized properly somewhere.

dyc3:
I can repro this bug with the item dupe check turned off, so it's unrelated to this PR. I'll open another issue for it.

PR #298, opened 2023-03-21
fix key input with filters, also removed regex from gui and make text…

### LINKED ISSUE #297: Packet inspector: Arrow keys are not functioning correctly.
## Valence Version

c9fbd1a24e86e6df4698640ef08ae59303ab728a

## What You Did

Clicked on a packet from the list on the left and used the arrow keys.

## What Went Wrong

Does not immediately go to the next or previous packet. It appears to deselect on the first input, then select the next packet on the next input.

AviiNL:
… filters case-insensitive

Closes #297

## Description

Up and down arrow to select previous/next packet now ignore filtered packets.

## Test Plan

1. Query the server using refresh on the server browser
2. Uncheck QueryResponseS2c in the packet-selector list
3. Click HandshakeC2s
4. Press down arrow until last packet is selected
5. It shouldn't go "blank" anymore when going from QueryRequestC2s to QueryPingC2s

As an additive request, I've removed the regex check from the main GUI filter box, such that this is now case-insensitive when filtering specific packets.
CLI still uses regex to be able to filter on multiple packets (unchanged)

### DIFF
--- a/crates/packet_inspector/src/context.rs
+++ b/crates/packet_inspector/src/context.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::RwLock;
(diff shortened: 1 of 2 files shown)


rj00a:
👍🏻

PR #302, opened 2023-03-24
Update to 1.19.4

### LINKED ISSUE #291: Update to Minecraft 1.19.4

rj00a:
## Description

Closes #291

- Update extractors to support Minecraft 1.19.4
- Update code generators.
- Changed generated entity component names to avoid name collisions.
- Update `glam` version.
- Added `Encode` and `Decode` for `glam` types in `valence_protocol`.
- Fixed inconsistent packet names and assign packet IDs automatically.
- Remove `ident` and rename `ident_str` to `ident`.
- Rework registry codec configuration. Biomes and dimensions exist as entities.`BiomeRegistry` and `DimensionTypeRegistry` resources have been added. The vanilla registry codec is loaded at startup.

### Issues
- Creating new instances has become more tedious than it should be. This will be addressed later.

## Test Plan

Steps:
1. Boot up a vanilla server with online mode disabled.
2. Run the `packet_inspector`.
3. Connect to the vanilla server through the packet inspector to ensure all packets are updated correctly.
4. Close the vanilla server and try some valence examples.


rj00a:
Not sure I'm happy with this approach to biome and dimension configuration.

In summary, biomes and dimensions exist as entities that can be queried. The main benefit is that user-defined components can be attached. `BiomeRegistry` and `DimensionTypeRegistry` contain lookup tables. However, there is the possibility that the tables become stale for a time after biomes/dimensions are added/removed.

Issue #311, opened 2023-03-30
Extract vanilla registry codec


rj00a:
## Describe the problem related to your feature request.

#302 Makes use of the default vanilla registry codec. However, the registry codec is not currently extracted automatically and must be acquired manually for each new MC version. It is placed in the `extracted/` directory.

## What solution would you like?

Create a new extractor for the serializable registries. It should be written in binary form uncompressed and placed in `extracted/registry_codec.dat`.

## Additional context

The registry codec is the NBT `Compound` sent in the `GameJoinS2c` packet.

PR #316, opened 2023-04-08
Registry codec extractor

Bafbi:
## Description

- #311 .
- Update the script to cp all the file of the output dir .

## Change to the gradle project

- Adding the fabric API .
- The serveur now need to start starting .

## Test plan

Do the same as before and compare the 1.9.4 output file and my generated one .


rj00a:
Thanks.

PR #318, opened 2023-04-10
Fix player bugs

rj00a:
## Description

Fix regressions from #317 and #315 

- Make sure player list packets get sent before player entity spawn packets are sent to prevent invisible players.
- Correctly update pose and hand swing.

## Test Plan

Steps:
1. Load any example with two players.
2. See that players are visible.
3. See that crouching and hand swings are functioning.


dyc3:
It would be good to add some unit tests so these regressions don't happen again.

PR #320, opened 2023-04-18
Fixed Panic when dropping item from inventory while chest is open #295

### LINKED ISSUE #295: Panic when dropping item from inventory while chest is open
## Valence Version

c9fbd1a24e86e6df4698640ef08ae59303ab728a

## What You Did

1. Run `cargo run --example chest`
2. give yourself any item and place it on hotbar
3. open the chest
4. mouse over the item in your hotbar
5. press drop key
6. crash and burn

<details>

<summary>Playground</summary>

```rust
PASTE YOUR PLAYGROUND CODE HERE
```

</details>

## What Went Wrong

The server should not panic.

## Additional Information

Caused by this line:

https://github.com/valence-rs/valence/blob/c9fbd1a24e86e6df4698640ef08ae59303ab728a/crates/valence/src/client/event.rs#L865

because it fails to consider that an inventory might be open, so slot ids for the player inventories may be different

SeseMueller:
## Description

Fixed #295 by checking the slot ids of the drop event and only updating the player inventory/the target inventory accordingly.

## Test Plan
Steps:
1. Rerun the test from #295, it works
2. Test other inventories in game
...

(No details, it only uses the chest example.)

### DIFF
--- a/crates/valence/src/inventory.rs
+++ b/crates/valence/src/inventory.rs
@@ -763,28 +763,86 @@ fn handle_click_slot(
             // The client is dropping an item by pressing the drop key.
 
             let entire_stack = pkt.button == 1;
-            if let Some(stack) = client_inv.slot(pkt.slot_idx as u16) {
-                // TODO: is the use of `replace_slot` here causing unnecessary packets to be
-                // sent?
-                let dropped = if entire_stack || stack.count() == 1 {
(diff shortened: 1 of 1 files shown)


dyc3 (changes requested):
This needs unit tests. See the ones at the bottom of `inventory.rs`.

dyc3 (approved):
LGTM

PR #321, opened 2023-04-19
Reorganize Project

### LINKED ISSUE #241: Specify the convention of crates naming

It seems that there are several crates that have a name without the `valence` prefix, such as `playground` and `packet_inspector`, so it's not obvious what the difference between crates with this prefix and without.

The crates with prefix `valence` should indicate that they are public crates for external usage, and the crates without this prefix should indicate that they are only for internal usage.

The public crates in this context means that they are published to the [crates.io](https://crates.io/) crates registry, also they are supposed to be a library with well documented public API, or a binary that solves problems not only specific to the framework. And the crates for internal usage are supposed to be libraries or binaries for framework development purposes only.

### Additional context

_No response_

### LINKED ISSUE #308: Merge `valence` and `valence_protocol`

The current `valence_protocol` architecture has felt somewhat flawed and messy. I have identified a few reasons for this. It is for these reasons that I believe a general-purpose protocol crate that covers everyone's use case is not really feasible.

- Packets are often inherently context sensitive. Their full definition can rely on things well outside the scope of `valence_protocol`. A few examples:
  - The entity tracker update packet is just a meaningless blob of bytes in `valence_protocol`. The actual structure and data types used in the packet are defined in `valence`.
  - The update recipes packet depends on things defined in a registry to parse correctly. All solutions seem bad from the perspective of a general-purpose library.
  - The chunk data packet is a blob of bytes.
- There is tension between making the packets precise and making them generally applicable. e.g. should text component fields use the `Text` type, or should they just be strings?
- It's not clear when something should be a part of `valence` or `valence_protocol`.
- Some types are duplicated between the crates. The `GameMode` enum is defined in `valence` because we want it to be a `Component`. But `valence_protocol` needs it and it shouldn't depend on `bevy_ecs`. So it's defined in both.
- `glam` is not used in the packets because maybe some hypothetical user of `valence_protocol` doesn't want/need it.
- Some random types used in packets are defined in the `types` module, but they are unrelated to each other and don't really belong together. It also feels strange to import them from `valence::protocol` in user code, so they should probably be re-exported in the appropriate locations.
- Code generators are split between `valence_protocol` and `valence`.
- Packets are bidirectional while also trying to be as zero-copy as possible. This means sprinkling `Cow` around randomly even when we don't really need it.

- Define packets in the `valence` modules where they're actually used. This should be doable because packets aren't really shared between different contexts. This makes sense with #296.
- Tweak the packets to suit the context where they're actually used.
- Get rid of `valence_protocol` and move top-level modules like `text` and `ident` into `valence`, also at the top level.
- Move `Encode`, `Decode` and the packet encoder/decoders to the `packet` module in `valence`.
- `valence_protocol_macros` becomes `valence_macros`.
- Make the packet inspector depend on `valence`.
- Introduce `encryption` and `compression` features to `valence`.
- Move the packet enums into `packet_inspector`.

There are some problems that may make us want to avoid doing this at all:
- **Increases compile times.** This can be solved by moving large modules into appropriate subcrates. Think `valence_connect`, `valence_command`, `valence_inventory`, `valence_anvil`, etc. This can happen transparently by introducing `valence_core` to avoid circular dependencies.
- **Utility crates like the packet inspector drag in more dependencies**. Meh, whatever.

rj00a:
## Description

There's a dizzying number of changes in this PR, but here are the highlights:
- `valence` and `valence_protocol` have been divided into smaller crates in order to parallelize the build and improve IDE responsiveness. In the process, code architecture has been made clearer by removing circular dependencies between modules. `valence` is now just a shell around the other crates.
- `workspace.packages` and `workspace.dependencies` are now used. This makes dependency managements and crate configuration much easier.
- `valence_protocol` is no more. Most things from `valence_protocol` ended up in `valence_core`. We won't advertise `valence_core` as a general-purpose protocol library since it contains too much valence-specific stuff. Closes #308.
- Networking code (login, initial TCP connection handling, etc.) has been extracted into the `valence_network` crate. The API has been expanded and improved with better defaults. Player counts and initial connections to the server are now tracked separately. Player counts function by default without any user configuration.
- Some crates like `valence_anvil`, `valence_network`, `valence_player_list`, `valence_inventory`, etc. are now optional. They can be enabled/disabled with feature flags and `DefaultPlugins` just like bevy.
- Whole-server unit tests have been moved to `valence/src/tests` in order to avoid [cyclic dev-dependencies](https://github.com/rust-lang/cargo/issues/4242).
- Tools like `valence_stresser` and `packet_inspector` have been moved to a new `tools` directory. Renamed `valence_stresser` to `stresser`. Closes #241.
- Moved all benches to `valence/benches/` to make them easier to run and organize.

Ignoring transitive dependencies and `valence_core`, here's what the dependency graph looks like now:

```mermaid
graph TD
	network --> client
	client --> instance
	biome --> registry
	dimension --> registry
	instance --> biome
	instance --> dimension
	instance --> entity
	player_list --> client
	inventory --> client
	anvil --> instance
	entity --> block
```

### Issues
- Inventory tests inspect many private implementation details of the inventory module, forcing us to mark things as `pub` and `#[doc(hidden)]`. It would be ideal if the tests only looked at observable behavior.
- Consider moving packets in `valence_core` elsewhere. `Particle` wants to use `BlockState`, but that's defined in `valence_block`, so we can't use it without causing cycles.
- Unsure what exactly should go in `valence::prelude`.
- This could use some more tests of course, but I'm holding off on that until I'm confident this is the direction we want to take things.

## TODOs
- [x] Update examples.
- [x] Update benches.
- [x] Update main README.
- [x] Add short READMEs to crates.
- [x] Test new schedule to ensure behavior is the same. 
- [x] Update tools.
- [x] Copy lints to all crates.
- [x] Fix docs, clippy, etc.


rj00a:
Looks like there have been some performance regressions.  CPU usage and MSPT is unusually high while the server is idling.

I suspect the problem existed before this PR, so I'll just address that later.

qualterz (commented):
There's something weird with weather tests, probably they are falling because of multithreading or another nondeterministic things. I think they can be removed to avoid future CI failures.

I'm saying that because in past I got a lot of CI failures because `app.update()` isn't always proceeded.

rj00a:
Are you saying the weather tests are failing sporadically after 259dbd5649002b46f6587b2d6a4f40932432ce2f? The reason they were failing before is because I removed the clamping behavior, and the tests needed to be updated to reflect that. But everything appears to be working correctly now.

qualterz:
I mean the weather tests before this pull request was failing sporadically, and to fix this issue I just called `app.update()` more than once like here: https://github.com/valence-rs/valence/pull/321/files#diff-3e32698ef1d91df6723f7e45308cc22be3d97247435d04116920548b018190b3R87

rj00a:
> FYI for your diagram, using `mermaid` on github is way easier than graphvis

Nice. I'll put the graph in a README somewhere then.

Issue #325, opened 2023-04-26
Advancements support


CodeDoctorDE:
## Describe the problem related to your feature request.

The ability create own advancements tabs.

## What solution would you like?

An easy to use api to modify these tabs

## What alternative(s) have you considered?

-\

## Additional context

There are already packet structs for this but it seems they are not connected to the api.

rj00a:
Yep, we should probably make a `valence_advancement` crate and feature flag for this.

PR #329, opened 2023-04-30
Advancement api

Jenya705:
## Description

Did an api for advancements.
Issue: https://github.com/valence-rs/valence/issues/325

Each advancement is an entity, it's children is either criteria, either advancement.
Root advancement has no parent.

Also did an event AdvancementTabChange (listens if client changes advancement's tab)

## Test Plan

Use an example "advancements"

### DIFF
--- a/crates/valence_advancement/src/lib.rs
+++ b/crates/valence_advancement/src/lib.rs
@@ -1,5 +1,4 @@
 #[doc = include_str!("../README.md")]
-
 pub mod event;
 
 use std::borrow::Cow;
@@ -142,7 +141,7 @@ impl<'w, 's> UpdateAdvancementCachedBytesQuery<'w, 's> {
                 requirements_p.push(c_identifier.0.as_str());
             }
             pkt.requirements.push(protocol::AdvancementRequirements {
-                requirement: requirements_p
+                requirement: requirements_p,
             });
         }
 
@@ -327,7 +326,7 @@ fn send_advancement_update_packet(
     }
 }
 
-/// Advancement's id. May not be updated. 
+/// Advancement's id. May not be updated.
 #[derive(Component)]
 pub struct Advancement(Ident<Cow<'static, str>>);
 
@@ -387,7 +386,8 @@ impl AdvancementCriteria {
 }
 
 /// Requirements for advancement to be completed.
-/// All columns should be completed, column is completed when any of criteria in this column is completed.
+/// All columns should be completed, column is completed when any of criteria in
+/// this column is completed.
 #[derive(Component, Default)]
 pub struct AdvancementRequirements(pub Vec<Vec<Entity>>);
 
--- a/crates/valence_advancement/src/lib.rs
+++ b/crates/valence_advancement/src/lib.rs
@@ -257,7 +257,6 @@ fn send_advancement_update_packet(
     update_single_query: SingleAdvancementUpdateQuery,
 ) {
     for (mut advancement_client_update, mut client) in client.iter_mut() {
-
         match advancement_client_update.force_tab_update {
             ForceTabUpdate::None => {}
             ForceTabUpdate::First => {
--- a/crates/valence_advancement/src/lib.rs
+++ b/crates/valence_advancement/src/lib.rs
@@ -10,19 +10,20 @@ use bevy_ecs::prelude::{Bundle, Component, Entity};
 use bevy_ecs::query::{Added, Changed, Or, With};
 use bevy_ecs::schedule::{IntoSystemConfig, IntoSystemSetConfig, SystemSet};
 use bevy_ecs::system::{Commands, Query, SystemParam};
-use bevy_ecs::world::Ref;
-#[doc(hidden)]
 pub use bevy_hierarchy;
 use bevy_hierarchy::{Children, Parent};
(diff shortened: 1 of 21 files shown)


CodeDoctorDE:
Thanks, if you use the word `closes #329`, the issue will automatically be closed when merging

crates/valence_advancement/src/lib.rs
```diff
@@ -0,0 +1,426 @@
...
+            Option<&'static Parent>,
+        ),
+    >,
+    advancement_id_query: Query<'w, 's, &'static Advancement>,
+    criteria_query: Query<'w, 's, &'static AdvancementCriteria>,
+    parent_query: Query<'w, 's, &'static Parent>,
+}
+
+#[derive(Debug)]
+pub(crate) struct AdvancementUpdateS2COE<'w, 's, 'a> {
```
dyc3, line 92: What's the difference between this and the packet definitions that already exist in `valence_core`?

Jenya705: - It is easier to implement caching of advancements
- I tried with default packet definitions and i didn't find a way to do that without cloning some things, what is not very well as i think.

crates/valence_advancement/src/lib.rs
```diff
@@ -0,0 +1,439 @@
...
+impl Plugin for AdvancementPlugin {
+    fn build(&self, app: &mut bevy_app::App) {
+        app.configure_set(
+            WriteAdvancementPacketToClientsSet
+                .in_base_set(CoreSet::PostUpdate)
+                .before(FlushPacketsSet),
+        )
+        .add_event::<AdvancementTabChange>()
+        .add_system(
+            init_clients
```
rj00a, line 43: This also needs to be ordered before `RunEventLoopSet` in case `AdvancementClientUpdate` is used in the event loop.

crates/valence_advancement/src/lib.rs
```diff
@@ -0,0 +1,439 @@
...
+impl AdvancementBundle {
+    pub fn new(ident: Ident<Cow<'static, str>>) -> Self {
+        Self {
+            advancement: Advancement::new(ident),
+            cached_bytes: AdvancementCachedBytes(Mutex::new(vec![])),
+        }
+    }
+}
+
+fn init_clients(mut commands: Commands, query: Query<Entity, Added<Client>>) {
```
rj00a, line 67: I would prefer if this were named something more descriptive like `add_advancement_component_to_new_clients` to help when reading the output from `bevy_mod_debugdump`.

crates/valence_advancement/src/lib.rs
```diff
@@ -0,0 +1,439 @@
...
+                .after(SpawnClientsSet)
+                .in_base_set(CoreSet::PreUpdate),
+        )
+        .add_system(handle_advancement_tab_change.in_base_set(CoreSet::PreUpdate))
+        .add_system(send_advancement_update_packet.in_set(WriteAdvancementPacketToClientsSet));
+    }
+}
+
+#[derive(Bundle)]
+pub struct AdvancementBundle {
```
rj00a, line 53: Can we move all the advancement components to this bundle? That way, users don't accidentally forget a component when spawning an advancement. You might have to make the `AdvancementCachedBytes` component public, but that's probably for the best.

crates/valence_advancement/src/lib.rs
```diff
@@ -0,0 +1,439 @@
...
+    pub fn get(&self) -> &Ident<Cow<'static, str>> {
+        &self.0
+    }
+}
+
+#[derive(Component)]
+pub struct AdvancementRequirements(pub Vec<Vec<Entity>>);
+
+#[derive(Component)]
+pub(crate) struct AdvancementCachedBytes(pub(crate) Mutex<Vec<u8>>);
```
rj00a, line 365: This mutex should be avoided somehow. Can you encode the necessary data into the cache before encoding the packet? That way, you won't need `&mut` when encoding the packet.

crates/valence_advancement/src/lib.rs
rj00a: I like that you've taken care to cache packets and avoid clones, but I dislike having to maintain two copies of the same packet. Can you modify the advancement packets in `valence_core` so that they can be used in `valence_advancement`? They'll also need to be both encodable and decodable so they can be used by `packet_inspector`.

In the future I'd like to move all the packets out of `valence_core` which might make this easier.

crates/valence_advancement/src/lib.rs
```diff
@@ -0,0 +1,439 @@
...
+            Option<&'static Parent>,
+        ),
+    >,
+    advancement_id_query: Query<'w, 's, &'static Advancement>,
+    criteria_query: Query<'w, 's, &'static AdvancementCriteria>,
+    parent_query: Query<'w, 's, &'static Parent>,
+}
+
+#[derive(Debug)]
+pub(crate) struct AdvancementUpdateS2COE<'w, 's, 'a> {
```
rj00a, line 96: Not a fan of this name. How about `AdvancementUpdateEncodeS2c`? `packet_name` should be updated to match.

rj00a (approved):
Looks good, thanks.

Issue #299, opened 2023-03-22
Entity Hitboxes


rj00a:
## Describe the problem related to your feature request.

Because of #294, the old entity hitbox calculation code stopped working and has been commented out. We need a new solution that fits the ECS framework.

## What solution would you like?

Accurate entity hitboxes matching what the client sees are a bit tricky because many separate pieces of data from the entity are needed. It depends heavily on the entity's type.

I would like to see the hitbox cached as a component and updated at user-configured points in the schedule. I'm guessing we will need a separate system for each entity type.

## What alternative(s) have you considered?

We could consider entity hitboxes to be out of scope for valence and punt it to a community plugin. However, I expect hitboxes will be needed frequently enough that a first-party solution will be desired.

## Additional Context

https://github.com/valence-rs/valence/blob/main/crates/valence/src/entity/hitbox.rs

PR #330, opened 2023-05-03
Hitboxes

Jenya705:
## Description

Added Hitbox component, a crate for it and systems which update hitboxes.
Issue: https://github.com/valence-rs/valence/issues/299

## Test Plan

Use example "entity_hitbox"


Jenya705:
> A Block hitbox can also be useful to add

They are already added, check BlockState::collision_shapes

crates/valence_entity_hitbox/src/lib.rs
```diff
@@ -0,0 +1,523 @@
...
+pub struct Hitbox(Aabb);
+
+impl Hitbox {
+    #[cfg(not(feature = "hitbox_add"))]
+    pub fn zero() -> Hitbox {
+        Hitbox(Aabb {
+            min: DVec3::ZERO,
+            max: DVec3::ZERO,
+        })
+    }
```
rj00a, line 75: This could be an associated constant instead.
```suggestion
pub const ZERO: Self = Self(Aabb { min: DVec3::ZERO, max: DVec3::ZERO });
```

crates/valence_entity_hitbox/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+
+#[cfg(feature = "hitbox_add")]
+#[derive(SystemSet, Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub struct HitboxComponentAddSet;
+
+pub struct HitboxPlugin;
+
+impl Plugin for HitboxPlugin {
+    fn build(&self, app: &mut App) {
+        app.configure_set(HitboxUpdateSet.in_base_set(CoreSet::PreUpdate))
```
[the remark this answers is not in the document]
rj00a: What's wrong with `PreUpdate`? The way you have it now looks good to me.

crates/valence_entity_hitbox/src/lib.rs
```diff
@@ -0,0 +1,523 @@
...
+type HitboxFilterSingle<T> = Or<T, Added<Hitbox>>;
+
+#[cfg(feature = "hitbox_add")]
+type HitboxFilter<T> = Or<T>;
+
+#[cfg(feature = "hitbox_add")]
+type HitboxFilterSingle<T> = T;
+
+#[derive(Component, Debug)]
+pub struct Hitbox(Aabb);
```
rj00a, line 66: Might be better to derive `Component` on `Aabb` and use that directly instead of this newtype. A side effect of this is that users would be able to mutate the AABB, which might actually be desirable in some circumstances.

crates/valence_entity_hitbox/src/lib.rs
```diff
@@ -0,0 +1,523 @@
...
+    pub(crate) fn centered(&mut self, size: DVec3) {
+        self.0 = Aabb::from_bottom_size(DVec3::ZERO, size);
+    }
+
+    pub fn in_world(&self, pos: DVec3) -> Aabb {
+        Aabb {
+            min: self.0.min + pos,
+            max: self.0.max + pos,
+        }
+    }
```
rj00a, line 90: The name and purpose of this function is not very clear. If you must, I think it would be better to implement `std::ops::Add`.

crates/valence_entity_hitbox/src/lib.rs
```diff
@@ -0,0 +1,523 @@
...
+
+#[cfg(feature = "hitbox_add")]
+#[derive(SystemSet, Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub struct HitboxComponentAddSet;
+
+pub struct HitboxPlugin;
+
+impl Plugin for HitboxPlugin {
+    fn build(&self, app: &mut App) {
+        app.configure_set(HitboxUpdateSet.in_base_set(CoreSet::PreUpdate))
```
rj00a, line 25: This needs `.before(EventLoopSet)`.

crates/valence_entity_hitbox/src/lib.rs
Jenya705: But should they mutate it? Because it will be updated on each tick so it will just confuse some people and because of that bugs can occur

crates/valence_entity_hitbox/src/lib.rs
Jenya705: The hitbox I store is for entities that are located on (0; 0) so in order to get aabb in the world space you need to use that function.

crates/valence_entity_hitbox/src/lib.rs
Jenya705: I didn't find the EventLoopSet

crates/valence_entity/src/hitbox.rs
```diff
@@ -1,294 +1,544 @@
...
+            EntityKind::CAMEL => [1.7, 2.375, 1.7],
+            EntityKind::CHICKEN => [0.4, 0.7, 0.4],
+            EntityKind::DONKEY => [1.5, 1.39648, 1.5],
+            EntityKind::FOX => [0.6, 0.7, 0.6],
+            EntityKind::GOAT => {
+                if pose_query
+                    .get(entity)
+                    .ok()
+                    .map(|v| v.0 == Pose::LongJumping)
+                    .unwrap_or(false)
```
rj00a, line 309: This could use `map_or`

crates/valence_entity/src/hitbox.rs
```diff
@@ -1,294 +1,544 @@
...
+pub struct Hitbox(pub Aabb);
 
-        Aabb {
-            min: center_pos - bounds / 2.0,
-            max: center_pos + bounds / 2.0,
+#[derive(Component, Debug)]
+/// Hitbox, aabb of which is calculated each tick using it's position and
+/// [`Hitbox`]. In order to change size of this hitbox you need to change
+/// [`Hitbox`]
+pub struct InWorldHitbox(Aabb);
```
rj00a, line 84: Why have separate components for a "size" hitbox and an "in world" hitbox? This seems a bit more complicated than it needs to be.

My suggestion is to modify your `update_*` systems to change the hitbox as the `Position` changes. So instead of `Added<Hitbox>`, you query for `Position` and use `Changed<Position>`.

I can see how this setup would let you freely change the hitbox size without changing its position, but I'm not sure we really need that. Apologies if my previous comments were misinterpreted.

Jenya705: How we ensure that the hitbox size is the same as the entity's? It is possible but the only way i found is more complicated as if we had a size of hitbox

rj00a: What's wrong with
```rust
fn update_zombie_hitbox(
    mut query: Query<
        (&mut Hitbox, &Position, &zombie::Baby),
        Or<(Changed<Position>, Changed<zombie::Baby>)>,
    >,
) {
    for (mut hitbox, pos, baby) in query.iter_mut() {
        hitbox.0 = Aabb::from_bottom_size(pos.0, child_hitbox(baby.0, [0.6, 1.95, 0.6].into()));
    }
}
```
Then `Hitbox` takes the place of `InWorldHitbox` and you can remove `InWorldHitbox`.

Jenya705: But if hitbox is changed out of all update hitbox functions?

rj00a: I don't understand?

Jenya705: As i think in your system is the problem of too many needless calculations (like for constant systems and other systems which are depend on `EntityKind`)

We can get rid of a `InWorldHitbox` if we do this kind of system:
```rust
fn update_in_world_hitbox(
    mut hitbox_query: Query<
        (&mut Hitbox, &OldPosition, &Position),
        Or<Changed<Position>>,
    >,
) {
    for (mut hitbox, old_pos, pos) in hitbox_query.iter_mut() {
        hitbox.0 = hitbox.0 - old_pos.0 + pos.0;
    }
}
```
If we want to change a size of hitbox, we do that:
```rust
impl Hitbox {
    fn change_size(&mut self, new_size: Aabb, old_pos: DVec3) {
        self.0 = new_size + old_pos;
    }
}
```

And also we should update all our systems to calculate hitbox from its old_pos, so `centered` function should be changed to like this (and also some systems that don't use centered should add old_pos to something):
```rust
impl Hitbox {
    fn centered(&mut self, size: DVec3, old_pos: DVec3) {
        self.0 = Aabb::from_bottom_size(DVec3::ZERO, size) + old_pos;
    }
}
```

But this solution is also complicated (for some methods we need to use `OldPosition`)

rj00a: > As i think in your system is the problem of too many needless calculations (like for constant systems and other systems which are depend on EntityKind)

Hmm, this looks like a classic case of premature optimization. Honestly I would do the simple approach right now and change it later if you can show that it's an improvement, because right now I'm not yet convinced this is actually faster.

BTW, your other solution involving `OldPosition` might be vulnerable to accumulated floating point errors.

Jenya705: I've been thinking and came to this. With my approach entity's custom hitboxes (for now only marker entity) will be easier to implement, because `InWorldHitbox` depends on `Position` and `Hitbox` when in yours the user will need to implement it's own system to change `Hitbox` to match it's new position.

rj00a: @dyc3 Do you have any opinions on this matter? 😄

dyc3: I think that the current solution here, where we have both `Hitbox` and `InWorldHitbox` types, is the easiest to understand, and allows for the most flexibility for users. I really don't like the `update_in_world_hitbox` system proposed in this thread, it feels like a very easy source of bugs. 

I agree with the premature optimization sentiment. I would feel much more comfortable optimizing this if we had some real numbers and real use cases to reference. Otherwise, the current impl seems perfectly fine to me.

rj00a: Alright, sounds good. Thanks for the input.

The last thing I'd consider changing is the name of the components. Perhaps `InWorldHitbox` should be renamed to `Hitbox` and `Hitbox` to `HitboxShape`, just to avoid confusion and emphasize that the "in world" hitbox is what users are going to be looking for most of the time. Other than that, LGTM.

rj00a (approved):

PR #340, opened 2023-05-23
update aes and cfb8 to 0.8

### LINKED ISSUE #42: Update `aes` and `cfb8` dependencies
`aes` and `cfb8` crates are at `0.8` currently but they are `0.7` in Valence. I tried updating a while back but couldn't figure out how to use the new API (It's very different).

Not much urgency on this one.

dyc3:
- update aes and cfb8 and make it compile
- update some usages to fix lints
- fix it so that it actually works
- small refactor
- fix lint

## Description

This updates the aes and cfb8 dependencies to 0.8

fixes #42

### DIFF
--- a/crates/valence_core/src/packet/decode.rs
+++ b/crates/valence_core/src/packet/decode.rs
@@ -1,5 +1,7 @@
 #[cfg(feature = "encryption")]
-use aes::cipher::{AsyncStreamCipher, NewCipher};
+use aes::cipher::generic_array::GenericArray;
+#[cfg(feature = "encryption")]
+use aes::cipher::{AsyncStreamCipher, BlockDecryptMut, KeyIvInit};
 use anyhow::{bail, ensure};
 use bytes::{Buf, BytesMut};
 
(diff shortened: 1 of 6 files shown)


crates/valence_core/src/packet/decode.rs
```diff
@@ -125,23 +127,34 @@ impl PacketDecoder {
...
+        Ok(())
+    }
+
+    /// Decrypts the provided byte slice in place using the cipher, without consuming the cipher.
+    #[cfg(feature = "encryption")]
+    fn decrypt_bytes(cipher: &mut Cipher, bytes: &mut [u8]) {
+        for chunk in bytes.chunks_mut(Cipher::block_size()) {
+            let gen_arr = GenericArray::from_mut_slice(chunk);
+            cipher.decrypt_block_mut(gen_arr);
+        }
```
rj00a, line 149: Might be able to express this more succinctly with `(en|de)crypt_blocks_inout_mut`, but it's probably fine either way.

rj00a (commented):
Glad to see this fixed :)

crates/valence_core/src/packet/decode.rs
dyc3: ngl I was having a hard time with the docs for these packages because they don't have many good examples. I'll see what I can do.

dyc3: I looked at it a bit more and I'm not convinced that inout doesn't do a copy, so lets leave it as is.

rj00a (approved):

PR #343, opened 2023-05-27
Add cli arg for log level to playground

### LINKED ISSUE #305: Playground: add cli args to set log level

Turning on debug messages for playgrounds requires you to change `main.rs`, which makes your working tree dirty and you could commit the change by mistake.

The playground should be able to accept a command line argument to change the log level

We could also just make it so that the playground runs all log messages enabled by default.

Gingeh:
## Description
Fixes #305 

`cargo run -- --log trace`

Uses `Level`'s `FromStr` impl, which accepts numbers 1-5 or case-insensitive level names. Default level is `Level::DEBUG`

We could use a dedicated argument parsing crate (`clap`?) instead, but idk if we want those extra dependencies.

### DIFF
--- a/tools/playground/src/main.rs
+++ b/tools/playground/src/main.rs
@@ -16,6 +16,9 @@
     clippy::dbg_macro
 )]
 
+use std::str::FromStr;
+use std::env;
+
 use tracing::Level;
 use valence::app::App;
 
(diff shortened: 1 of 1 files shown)


tools/playground/src/main.rs
```diff
@@ -24,9 +27,13 @@ mod extras;
...
 
 fn main() {
-    tracing_subscriber::fmt()
-        .with_max_level(Level::DEBUG)
-        .init();
+    let mut args = env::args().skip(1);
+    let log_level = match (args.next().as_deref(), args.next()) {
+        (Some("-l" | "--log"), Some(level)) => Level::from_str(&level).unwrap_or(Level::DEBUG),
+        _ => Level::DEBUG,
+    };
```
dyc3, line 34: I'd prefer if we went ahead and pulled in clap. We'll probably want to add some additional args for things like connection mode in the future.

Go ahead and create a `struct Args` and use the clap derive api for this.

dyc3 (changes requested):

dyc3 (approved):
Perfect! LGTM.

PR #350, opened 2023-05-30
tag registry

### LINKED ISSUE #349: Make a `TagRegistry` resource in `valence_registry`. initialize with default minecraft tags. Send `SynchronizeTagsS2c` packet on join.
Our extractor creates a `registry_codec.dat` file, which does _not_ contain all the data we need for `TagRegistry`. It only contains these registries:
- `minecraft:chat_type`
- `minecraft:damage_type`
- `minecraft:dimension_type`
- `minecraft:worldgen/biome`
This file is created by the `Codec` extractor, and uses Minecraft's `SerializableRegistries` class.

However, all the registries are listed in the `Registries` class.

dyc3:
- add Tags extractor
- add tags.json to extracted
- send `SynchronizeTagsS2c` packet on join
- fix encode

## Description

Adds a `TagsRegistry` resource that contains all the information needed to build and send `SynchronizeTagsS2c` on join.

closes #349

### DIFF
--- a/crates/valence_core/src/protocol/packet.rs
+++ b/crates/valence_core/src/protocol/packet.rs
@@ -1392,6 +1392,6 @@ pub mod synchronize_tags {
     #[derive(Clone, PartialEq, Eq, Debug, Encode, Decode)]
     pub struct Tag<'a> {
         pub name: Ident<Cow<'a, str>>,
-        pub entries: Vec<VarInt>,
+        pub entries: Vec<u64>,
     }
 }
--- a/crates/valence_core/src/protocol/packet.rs
+++ b/crates/valence_core/src/protocol/packet.rs
@@ -1372,26 +1372,3 @@ pub mod map {
         }
     }
 }
-
-// TODO: Move this to valence_registry?
-pub mod synchronize_tags {
-    use super::*;
-
-    #[derive(Clone, Debug, Encode, Decode, Packet)]
-    #[packet(id = packet_id::SYNCHRONIZE_TAGS_S2C)]
-    pub struct SynchronizeTagsS2c<'a> {
-        pub tags: Vec<TagGroup<'a>>,
-    }
-
-    #[derive(Clone, PartialEq, Eq, Debug, Encode, Decode)]
-    pub struct TagGroup<'a> {
-        pub kind: Ident<Cow<'a, str>>,
-        pub tags: Vec<Tag<'a>>,
-    }
-
-    #[derive(Clone, PartialEq, Eq, Debug, Encode, Decode)]
-    pub struct Tag<'a> {
-        pub name: Ident<Cow<'a, str>>,
-        pub entries: Vec<u64>,
-    }
-}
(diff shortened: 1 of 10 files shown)


crates/valence_registry/src/tags.rs
```diff
@@ -0,0 +1,95 @@
...
+            .expect("tags.json is invalid");
+    tags.registries = registries;
+}
+
+pub fn cache_tags_packet(tags: ResMut<TagsRegistry>) {
+    if tags.is_changed() {
+        let tags = tags.into_inner();
+        let packet = tags.build_synchronize_tags();
+        let mut bytes = vec![];
+        encode_packet(&mut bytes, &packet).expect("failed to encode tags packet");
```
rj00a, line 66: Ah, this needs to use a temporary `PacketWriter` with `.write_packet` instead of using `encode_packet`. You'll have to get the global compression threshold from `Res<Server>`. (sorry it's a bit confusing. `encode_packet` should probably be made private.)

crates/valence_registry/src/tags.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Debug, Deserialize)]
+pub struct TagEntry {
+    pub name: Ident<String>,
+    #[serde(rename = "values")]
+    pub entries: Vec<u64>,
+}
+
+impl TagsRegistry {
+    pub(crate) fn build_synchronize_tags(&self) -> SynchronizeTagsS2c<'_> {
+        SynchronizeTagsS2c {
```
rj00a, line 30: The `SynchronizeTagsS2c` packet definition should be moved to `valence_registry` somewhere. Looks like you can modify the packet to just contain a `Cow<[Registry]>` so you don't need to do this work to build the packet.

crates/valence_registry/src/tags.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub kind: Ident<String>,
+    pub tags: Vec<TagEntry>,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct TagEntry {
+    pub name: Ident<String>,
+    #[serde(rename = "values")]
+    pub entries: Vec<u64>,
+}
```
rj00a, line 26: Would be nice to not need `serde(rename = ...)`. "kind" and "entries" were chosen arbitrarily.

crates/valence_core/src/protocol/packet.rs
```diff
@@ -1392,6 +1392,6 @@ pub mod synchronize_tags {
     #[derive(Clone, PartialEq, Eq, Debug, Encode, Decode)]
     pub struct Tag<'a> {
         pub name: Ident<Cow<'a, str>>,
-        pub entries: Vec<VarInt>,
+        pub entries: Vec<u64>,
```
rj00a, line 1395: Are you sure this is supposed to be a `u64`? wiki.vg says it's a `VarInt`.

crates/valence_registry/src/tags.rs
rj00a, line 1: In modules with systems, I've kept the system functions private so far. Should we deviate from that here?

dyc3: No probably not.

dyc3: I've looked at it again, and I would prefer to keep the systems related to `TagsRegistry` in the same file, so I'll leave it as is for now.

Issue #332, opened 2023-05-06
Brigadier


Jenya705:
## Describe the problem related to your feature request.

There is no command api for valence. I would start with porting brigadier (library that mojang uses to handle commands)

## What solution would you like?

A command object which will contain nodes.
Node will contain:
- `id` - it's id in that command object
- `children` - it's children in that command object.
- `kind` - kind of node (integer, string, executable...)
- `suggestions` - an optional suggestion of node (type of suggestion or executable)

Command object will handle:
- command executions
- command suggestions

## What alternative(s) have you considered?

Just do like in bukkit. All arguments passed into an array of strings.

## Additional context

Brigadier: https://github.com/Mojang/brigadier

dyc3:
We've already discussed this in discord quite a bit, but I can't seem to find the thread right now.

rj00a:
We discussed this a bit on discord in the past, but the way Brigadier relies on mutable callback functions and dynamic typing for the `executes` context doesn't seem like a great fit for Rust or ECS (or any language really). For example:

```java
CommandDispatcher<CommandSourceStack> dispatcher = new CommandDispatcher<>();

dispatcher.register(
    literal("foo")
        .then(
            argument("bar", integer())
                .executes(c -> {
                    System.out.println("Bar is " + getInteger(c, "bar"));
                    return 1;
                })
        )
        .executes(c -> {
            System.out.println("Called foo with no arguments");
            return 1;
        })
);
```

Instead of the above, I would rather see something roughly like this:

```rust
#[derive(Command)]
struct MyFooCommand {
    bar: Option<i32>,
}

fn handle_foo_commands(foos: EventReader<CommandEvent<MyFooCommand>>) {
    for event in foos.iter_mut() {
        match event.command.bar {
            Some(n) => println!("Bar is {n}"),
            None => println!("Called foo with no arguments"),
        }
    }
}
```
In other words, you describe the structure of the command graph as a Rust data type and pattern match on it to do the execution.

You'd register `MyFooCommand` as `"foo"` somewhere (`CommandContext` resource or something) and off you go.

You could also represent cyclic commands this way:
```rust
#[derive(Command)]
struct MyCyclicCommand {
    blah: i32,
    recursion: Option<Box<MyCyclicCommand>>,
}
```

(Disclaimer: I have not actually used Brigadier)

Jenya705:
[...]

PR #351, opened 2023-06-02
Command API

Jenya705:
## Description

Will solve issue: #332 
This is draft, because I want to get feedback on early stage.
- Nodes are entities, that are connected with each other using `NodeFlow`
- Executable nodes have `NodeSystem` component, which contains a bevy's system, which will be executed when the command is executed
- Argument nodes have `NodeParser` component, which contains a boxed trait, which parses reader and puts the parsed object into special struct called `ParseResultsWrite`
- Literal nodes have just `NodeName` on them (no `NodeParser`)
- Root nodes can have `NodeExclude` component on them to exclude nodes that this specific root node doesn't want to use
- Multiple nodes which can be chosen using `EntityNode` component (if there is not such component on an entity then `PrimaryRootNode` will be chosen)
- Now commands have a few methods to manipulate with node's entities ( `spawn_node`, `node`, `spawn_root_node`)

Check out example `command.rs`


crates/valence_command/src/reader.rs
```diff
@@ -0,0 +1,274 @@
...
+        })
+    }
+
+    pub fn read_num<T: FromStr>(&mut self) -> Result<T, <T as FromStr>::Err> {
+        self.read_num_str().parse()
+    }
+
+    pub fn remaining_str(&self) -> &'a str {
+        // SAFETY
+        // cursor is always valid
```
dyc3, line 196: nit: would prefer safety comments to all be on the same line like TODO, eg. `// SAFETY: <msg>`

dyc3 (commented):
Very excited to see progress on this! I haven't looked too hard at this yet, just a couple notes.

crates/valence_command/src/parser.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        impl $crate::parser::ParsingBuild<$crate::parser::ParsingError> for $name {
+            fn build(self) -> $crate::parser::ParsingError {
+                valence_core::text::Text::translate($key, vec![])
+            }
+        }
+    }
+}
+
+pub trait Parsable<'a>: 'a + Sized {
```
rj00a, line 156: It would be more idiomatic to name this trait `Parse`.

rj00a:
Not a full review from me yet, but that's what I've noticed so far.

crates/valence_command/src/argument.rs
```diff
@@ -0,0 +1,1418 @@
...
+/// - color
+/// - dimension
+/// - entity_anchor
+/// - gamemode
+/// - heightmap
+/// - operation
+/// - swizzle (?)
+/// - template mirror
+/// - template rotation
+pub trait CEnum: Sized {
```
rj00a, line 1081: What does the name `CEnum` mean? This isn't clear to me.

Jenya705: C-Like enum. I understand that It is bad naming but I can not think of other one for now

crates/valence_command/src/parser.rs
```diff
@@ -0,0 +1,232 @@
...
+                    (Some(v), Some(o)) => Some((o, v)),
+                    _ => None,
+                }),
+        }
+    }
+}
+
+/// Macro ensures that returned result contains no Error
+#[macro_export]
+macro_rules! parsing_ret_err {
```
rj00a, line 100: I strongly dislike declarative macros (especially those in public APIs) unless it's absolutely necessary. Declarative macros should be tools of last resort. I suggest finding a way to avoid these kinds of macros, but this indicates to me that something about `ParsingResult` is more complicated than it should be.

crates/valence_command/src/argument.rs
rj00a, line 1: There's quite a few uses of `unsafe` in this crate so far and I'm a bit worried about their correctness. I would strongly suggest removing them, even if it means worse performance. Don't try to speed things up using `unsafe`  unless you have benchmarks and proof that this is something that's even worth doing. Every `unsafe` introduced is a potential vulnerability.

Jenya705: I think I removed each line of unsafe code that can be rewritten as safe (in most situations it just was `.get_unchecked()` -> `get().unwrap()`). I didn't remove `MaybeUninit` usage in `[T; C]` because I don't see a way to do this without limiting `T` to be `Default + Copy` or using a vector and limiting `T` to be `Default`.

crates/valence_core/src/game_mode.rs
```diff
@@ -10,3 +10,15 @@ pub enum GameMode {
...
+impl GameMode {
+    /// Converts gamemode to its number representation
+    /// ### Values
+    /// 0. Survival
+    /// 1. Creative
+    /// 2. Adventure
+    /// 3. Spectator
+    pub fn to_index(self) -> usize {
+        self as usize
+    }
```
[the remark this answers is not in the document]
Jenya705: It will be used in `EntitySelector` because I want to store `gamemode` field as an array `[bool; 4]` and `GameMode::to_index` is the index. (I was already doing `EntitySelector` but deleted this code because It was done with another `Parse` trait)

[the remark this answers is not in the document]
dyc3: I dislike this because it makes the order of values in `GameMode` matter, and `EntitySelector` would have to be tweaked if a new game mode is added.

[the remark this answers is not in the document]
rj00a: The order of variants already matters because `GameMode` is `#[derive(Encode, Decode)]`. If you must keep this `to_index` function I would prefer it moved to `valence_command`.

Jenya705:
I changed the whole system again. Now:
- Nodes are entities, that are connected with each other using `NodeFlow`
- Executable nodes have `NodeSystem` component, which contains a bevy's system, which will be executed when the command is executed
- Argument nodes have `NodeParser` component, which contains a boxed trait, which parses reader and puts the parsed object into special struct called `ParseResultsWrite`
- Literal nodes have just `NodeName` on them (no `NodeParser`)
- Root nodes can have `NodeExclude` component on them to exclude nodes that this specific root node don't want to use
- Multiple nodes which can be chosen using `EntityNode` component (if there is not such component on an entity then `PrimaryRootNode` will be chosen)
- Now commands have a few methods to manipulate with node's entities ( `spawn_node`, `node`, `spawn_root_node`)

Check out example `command.rs`

Jenya705:
What I think should also be done:
- [x] Suggestions - the same as execution systems but they will only accept `ReadOnlySystemParam`
- [ ] Better error handles - many of error cases are not handled correctly, they just panic
- [ ] Better safety comments - I didn't write a lot of comments for unsafe blocks
- [ ] Command macro - a command macro, which will automatically insert all required nodes into Root
- [ ] Super root node - a root node, which doesn't need to have a specific Entity. The same as `NodeRoot` but has all not root nodes in it
- [ ] Node name -> Node entity - helper to get a node by it's name
- [ ] Parallel execution - we can do the same what bevy does with their systems

examples/command.rs
```diff
@@ -0,0 +1,96 @@
...
+            child.name(Cow::Borrowed("creative")).executor(
+                |In(arguments): In<CommandArguments>, mut query: Query<&mut GameMode>| {
+                    if let RealCommandExecutor::Player(client) = arguments.1 {
+                        if let Ok(mut game_mode) = query.get_mut(client) {
+                            *game_mode = GameMode::Creative;
+                        }
+                    }
+                },
+            );
+        })
```
dyc3, line 57: Interesting... I'd like to see another example command that implements `/tp x y z` or `/give`. It's not really clear to me how one would go about that.

examples/command.rs
```diff
@@ -81,6 +81,27 @@ fn setup(
                     },
                 );
             });
+        })
+        .with_child(|child| {
+            child.name(Cow::Borrowed("tp")).with_child(|child| {
+                child.name(Cow::Borrowed("tp.x")).parser::<i32>(Default::default()).with_child(|child| {
+                    child.name(Cow::Borrowed("tp.y")).parser::<i32>(Default::default()).with_child(|child| {
+                        child.name(Cow::Borrowed("tp.z")).parser::<i32>(Default::default())
```
dyc3, line 89: Could the gamemode command above be refactored to use a parser for the parameter?

Jenya705: GameMode must implement `Parse` and `RawParseSuggestions` traits. I didn't do it yet because I want to deal with essential parts like suggestions handling and so on and then do all of brigadier's parsers.

JackCrumpLeys:
I know that this is likely completely of the rails compared to the current implementation but cant we just use bevy plugins and events? this way we are tightly coupled with bevy and would make it much nicer to code. I have a few ideas on how this could look here is one (I just typed these up on the textbox so could have errors):
```rs
struct TeleportCommand;

impl Command for TeleportCommand {
    fn command(&self) -> impl Into<String> {
        "tp"
    }
    fn params(&self) -> impl Into<Vec<CommandParameter>> { // with CommandParameter being some kind of enum
        vec![CommandParameter::Position]
        // This might also be vec![CommandParameter::Position(None)] or even vec![CommandParameterMarker::Position]
    }
    // ...
}

// ...
app.add_plugins(CommandHandler::from_command(TeleportCommand)) 
// this should send out events that look something like CommandExecutionEvent<TeleportCommand> 
// this event should contain the params. 
```
I have good ideas on how to make this happen and if you think it is a good idea I could work on the code? again I am really just throughing something out into the blue here but I would like to see an as ergonomic as possible API for this.

hwittenborn:
I haven't looked too much into this PR yet, but would it be possible to make command parsing work via the `clap` crate? I feel like that'd make things a lot simpler and reliable long-term.

Sorry if this PRs already doing that or anything, it was just a thought in my head and I thought I'd go ahead and mention it.

dyc3:
No, the way that minecraft commands work is not compatible with clap. IIRC, this was explicitly discussed on the thread in discord. Minecraft commands can have cyclic references for suggestions, and parsers for stuff like coordinates and entity references (eg, `@p`) would be very hard under clap's model.

crates/valence_command/src/command.rs
```diff
@@ -0,0 +1,97 @@
...
+use valence_core::text::Text;
+
+use crate::exec::CommandExecutionEvent;
+use crate::nodes::EntityNode;
+use crate::parse::ParseResultsRead;
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum RealCommandExecutor {
+    Player(Entity),
+    Console,
```
JackCrumpLeys, line 24: Why would console exist? It should be up to the user of valance to make logic to execute commands though the console. Is this used anywhere in the protocol?

crates/valence_command/src/nodes.rs
```diff
@@ -0,0 +1,420 @@
...
+
+use crate::command::CommandArguments;
+use crate::parse::{Parse, ParseObject, ParseWithData};
+use crate::pkt::{self, RawCommandTreeS2c};
+use crate::suggestions::RawParseSuggestions;
+
+pub(crate) type PCRelationVec = SmallVec<[Entity; 2]>;
+
+#[derive(Component, Debug)]
+pub struct NodeName(pub(crate) Cow<'static, str>);
```
JackCrumpLeys, line 24: Why not just store a `String`?

crates/valence_command/src/nodes.rs
```diff
@@ -0,0 +1,420 @@
...
+use valence_core::__private::VarInt;
+use valence_core::protocol::encode::WritePacket;
+use valence_core::protocol::{Encode, Packet};
+
+use crate::command::CommandArguments;
+use crate::parse::{Parse, ParseObject, ParseWithData};
+use crate::pkt::{self, RawCommandTreeS2c};
+use crate::suggestions::RawParseSuggestions;
+
+pub(crate) type PCRelationVec = SmallVec<[Entity; 2]>;
```
JackCrumpLeys, line 21: Does this ever exceed 2 things? If not why not just use a tuple or array?

JackCrumpLeys (commented):
I had a little skim of the code and reviewed a little.

crates/valence_command/src/command.rs
Jenya705: It is not covered by protocol because it has nothing to do with it, console's command executions handled by server anyway (the same as block's command executions). And I think that console can be a command executor, the same as a block.

crates/valence_command/src/nodes.rs
Jenya705: It is a vec for parents and children relation, so yeah it can exceed 2 elements

JackCrumpLeys (commented):
Some more comments.

crates/valence_command/src/nodes.rs
JackCrumpLeys: Generally when you have a graph like this you want to use edges between the nodes. How do you Tel if an item in this list is parent or child?

Jenya705:
Suggestions are done. @rj00a, can you review current code?

rj00a:
Sorry, I might be a bit slow to review this.

Jenya705:
The basic things are done and the problem with storing nodes as bevy's entities also. Now I just need to do a macro and other's arguments

JackCrumpLeys:
This should be closed or renamed as #446 has been merged.

JackCrumpLeys:
@rj00a this should be closed.

PR #352, opened 2023-06-04
`serde` support for `valence_nbt` + fixes.

rj00a:
## Description

- Reorganize `valence_nbt` and feature flag the different parts. SNBT and binary serialization are each behind their own flags.
- Add optional serde support to `valence_nbt` behind the `serde` flag. This is useful for end users working with `valence_nbt` and allows us to simplify some code in `valence_biome` and `valence_dimension`.

  Note that this includes a `Serializer` and `Deserializer` for `Compound` and _not_ the binary and SNBT formats. In other words, it's not possible to go directly from an arbitrary data format to binary NBT and vice versa, but it _is_ possible to go to and from `Compound` and finish the (de)serialization this way. I consider this an acceptable compromise because writing fast and correct serialization routines for `Compound` becomes more difficult when serde is in the way. Besides, the intermediate `Compound` often needs to be created anyway.
- Fixed unsound uses of `std::mem::transmute` in `valence_nbt` and elsewhere. Using `transmute` to cast between slice types is unsound because slices are `#[repr(Rust)]` and the layouts are not guaranteed. `slice::from_raw_parts` is used as the replacement.


dyc3 (approved):
lgtm

PR #355, opened 2023-06-06
Very rough Inventory Menu prototype

dyc3:
## Description

- refactor handle_click_slot to primarily emit ClickSlot events, and handle those in a new system
- fix cursor item updates not being sent if an inventory is open
- very rough inventory menu behavior
- only allow ClickMode::Click
- emit MenuClickEvents for inventory menu clicks

fixes #307 

This PR makes it possible to do inventory menus. I don't particularly like the approach so far. I think I'll need to do some refactoring of how inventories work before continuing work on this.

<details>

<summary>Playground</summary>

```rust
use valence::client::despawn_disconnected_clients;
use valence::client::hand_swing::HandSwingEvent;
use valence::inventory::menu::{InventoryMenu, InventoryMenuPlugin, MenuClickEvent, OpenMenu};
use valence::network::ConnectionMode;
use valence::prelude::*;

#[allow(unused_imports)]
use crate::extras::*;

const SPAWN_Y: i32 = 64;

pub fn build_app(app: &mut App) {
    app.insert_resource(NetworkSettings {
        connection_mode: ConnectionMode::Offline,
        ..Default::default()
    })
    .add_plugins(DefaultPlugins)
    .add_startup_system(setup)
    .add_system(init_clients)
    .add_system(despawn_disconnected_clients)
    .add_system(toggle_gamemode_on_sneak.in_schedule(EventLoopSchedule))
    .add_plugin(InventoryMenuPlugin::default())
    .add_system(open_menu.in_schedule(EventLoopSchedule))
    .add_system(handle_menu.in_schedule(EventLoopSchedule));
}

fn setup(
    mut commands: Commands,
    server: Res<Server>,
    biomes: Query<&Biome>,
    dimensions: Query<&DimensionType>,
) {
    let mut instance = Instance::new(ident!("overworld"), &dimensions, &biomes, &server);

    for z in -5..5 {
        for x in -5..5 {
            instance.insert_chunk([x, z], Chunk::default());
        }
    }

    for z in -25..25 {
        for x in -25..25 {
            instance.set_block([x, SPAWN_Y, z], BlockState::GRASS_BLOCK);
        }
    }

    commands.spawn(instance);
}

fn init_clients(
    mut clients: Query<(&mut Location, &mut Position), Added<Client>>,
    instances: Query<Entity, With<Instance>>,
) {
    for (mut loc, mut pos) in &mut clients {
        loc.0 = instances.single();
[...]

PR #356, opened 2023-06-07
refactor held item into a new component

dyc3:
## Description

This moves the `held_item_slot` field in `ClientInventoryState` to a new component: `HeldItem`

related: pr #355


rj00a:
I'll merge this but we need to integrate this with the clientbound [set held item packet](https://wiki.vg/Protocol#Set_Held_Item).

PR #358, opened 2023-06-07
Update to Minecraft 1.20

### LINKED ISSUE #357: Update to 1.20
Minecraft 1.20 has released, so we need to update valence to keep tracking the latest version.

dyc3:
- update extractor to 1.20
- update gradle to 8.1
- docs: describe how to update valence to new nc version
- update extractor mod metadata
- update extracted data
- replace 1.19.4 strings with 1.20

## Description

Updates valence to 1.20

closes #357

Not yet complete. I have no idea why its not building right now.


rj00a:
Thanks for working on this.

> I have no idea why its not building right now.

The entity extractor is a bit buggy and the way it handles tracked data is incorrect. The output is nondeterministic which causes some weirdness. If you change `0 air` to `1 air` in the generated `entities.json` it should pass the assertion in the build script for now.

dyc3:
I'm getting this for the `block_entities` example:

```
the trait bound `Value: From<valence::prelude::Uuid>` is not satisfied
```

Seems weird. Did we have this impl before?

rj00a:
> Seems weird. Did we have this impl before?

I think I changed the default feature flags for `valence_nbt` in the top-level Cargo.toml. Try adding the `uuid` feature to it.

rj00a:
Not sure why it's producing an error now though.

dyc3:
I don't have enough context about how the connection handshake works to continue debugging this, but it seems like something changed with packet decoding.

image

rj00a:
[Update 1.20 diff on wiki.vg](https://wiki.vg/index.php?title=Protocol&diff=18250&oldid=18242)

dyc3:
I believe the packet that causes the error is `LoginSuccessS2c`

PR #361, opened 2023-06-11
Registry Redesign

rj00a:
## Description

- Revert the "biomes/dimensions as entities" idea since it caused too many problems without much to show for it.
- Make use of `valence_nbt`'s serde support in `valence_biome` and `valence_dimension`.
- Reduce boilerplate, reorganize `valence_registry` a bit.
- Tweak default biome registry such that `BiomeId::default` always corresponds to "minecraft:plains".
- Add `Option` and unit variant support to `valence_nbt`'s serde impl.


dyc3 (approved):
LGTM

Should we consider moving stuff like blocks and items into registries, kinda like how vanilla does it?

rj00a:
That would be necessary if we plan to support modded. In that case, we should distinguish between static and dynamic registries. I'm assuming static registries should be append-only and get one global variable per default entry.

Issue #342, opened 2023-05-26
Visibility Layers


rj00a:
## Describe the problem related to your feature request.

Sometimes it's necessary to limit what information sets of clients can receive. Here are some example scenarios:
1. You have a game with players and spectators. Spectators should see other spectators and players, but players should only see other players and not the spectators.
2. You have two teams: Red and Blue. Everyone on the same team will "glow" and can be seen through walls. But players on the other team will not glow and cannot be seen through walls.

## What solution would you like?

Create a new component called `LayerMask` (name up for debate)
```rust
#[derive(Component)]
pub struct LayerMask(pub u64);
```

`LayerMask` is simply a bitmask where each bit represents a layer and whether or not it is active. `LayerMask`s are attached to entities and determine what layers the entity is active on.

Next, we have `ClientLayerMask` (name up for debate)
```rust
#[derive(Component)]
pub struct ClientLayerMask(pub u64);
```

This is also a bitmask, but it determines what layers a client can see. `ClientLayerMask`s are attached to clients only.

_Entity_ visibility is not the only thing we want to control with layers. For instance, if an entity is meant to emit a particle effect, the particles should probably follow the same visibility rules as the entity. So a hypothetical API for emitting particles (or sounds, block events, etc.) should look something like this:

```rust
pub struct PlayParticleEvent {
    pub instance: Entity,
    pub particle: Particle,
    pub position: DVec3,
    pub layers: LayerMask, // <---
}
```

### Implementation Details

We want to continue to cache packets in per-chunk and per-instance buffers. Therefore, we need one packet buffer per layer wherever we previously had just one.

```rust
Vec<u8> // Old
[Vec<u8>; 64] // New
```

(Instead of a `[T; 64]`, we could use `SmallVec<[T; 4]>`. The smallvec would grow on demand. But maybe there's an even better way...)

When clients go to read the packet buffers, they use their `ClientLayerMask` to determine which buffers to read from.

### Questions
- `u64` limits the number of unique layers to `64`. Is that enough for everyone?
- Should we integrate layers into spatial queries like collision and raycasts? A hierarchical spatial index could store layer masks at each node representing the union of its children. This can be used to cull subtrees when querying.
- What should the default value of `LayerMask` and `ClientLayerMask` be? `1` seems reasonable.
- The solution described doesn't fully account for scenario (2). Because "glowing" is part of the entity's metadata, you'd need to create two identical entities and put each on a separate layer to get the desired behavior. Can we fix that?
[...]

PR #362, opened 2023-06-11
Visibility layer

Bafbi:
## Ref
#342 

## Description

> I use the word 'mob' to reference minecraft entities and client for the player.

Implementation of a layer system for mob, the client can have a `ClientLayerMask` component which determine for what layer he can see mobs. The mobs them can have a `Layer` component which determine on what layer they are.

## Implementation

- [x] Client see only update for mask layer
- [ ] Spawn or Sync mob status when arriving on a new layer
- [ ] Despawn mob when quitting a layer

## Potential Upgrade

- Not have one buffer per layer
- Having the possibility to have infinite numbers of layers
- Extend layers to blocks ?

### DIFF
--- a/crates/valence_client/src/lib.rs
+++ b/crates/valence_client/src/lib.rs
@@ -644,14 +644,8 @@ pub struct IsDebug(pub bool);
 #[derive(Component, Copy, Clone, PartialEq, Eq, Default, Debug)]
 pub struct IsFlat(pub bool);
 
-#[derive(Component, Debug)]
+#[derive(Component, Debug, Clone, PartialEq, Eq, Default)]
 pub struct ClientLayerMask(pub u64);
-
-impl Default for ClientLayerMask {
-    fn default() -> Self {
-        Self(0)
-    }
-}
 
 impl ClientLayerMask {
     pub fn set(&mut self, layer: u8, enabled: bool) {
@@ -660,6 +654,10 @@ impl ClientLayerMask {
         } else {
             self.0 &= !(1 << layer);
         }
+    }
+
+    pub fn toggle(&mut self, layer: u8) {
+        self.0 ^= 1 << layer;
     }
 
     pub fn get(&self, layer: u8) -> bool {
@@ -909,14 +907,24 @@ fn read_data_in_old_view(
         &OldPosition,
         &OldViewDistance,
         Option<&PacketByteRange>,
-        Option<&ClientLayerMask>
+        Option<&ClientLayerMask>,
     )>,
     instances: Query<&Instance>,
     entities: Query<(EntityInitQuery, &OldPosition)>,
     entity_ids: Query<&EntityId>,
 ) {
     clients.par_iter_mut().for_each_mut(
-        |(mut client, mut remove_buf, loc, old_loc, pos, old_pos, old_view_dist, byte_range, client_layer_mask)| {
+        |(
+            mut client,
+            mut remove_buf,
+            loc,
+            old_loc,
+            pos,
+            old_pos,
+            old_view_dist,
+            byte_range,
+            client_layer_mask,
+        )| {
             let Ok(instance) = instances.get(old_loc.get()) else {
                 return;
             };
@@ -968,13 +976,11 @@ fn read_data_in_old_view(
                         }
                     }
 
-                    // Using the layer mask we gonna send the right ranges of bytes to the client from the layer packet buffer.
+                    // Using the layer mask we gonna send the right layers buffers of bytes to the
+                    // client
                     if let Some(client_layer_mask) = client_layer_mask {
-                        // Send all the ranges that the client is interested in.
                         client_layer_mask.get_all().iter().for_each(|layer| {
-                            if let Some(byte_range) = &cell.layers_byte_range.get(layer) {
-                                client.write_packet_bytes(&cell.layers_packet_buf[byte_range.start..byte_range.end]);
-                            }
+                            client.write_packet_bytes(&cell.layers_packet_buf[*layer as usize][..]);
                         });
                     }
 
--- a/crates/valence_client/src/lib.rs
+++ b/crates/valence_client/src/lib.rs
@@ -644,6 +644,40 @@ pub struct IsDebug(pub bool);
 #[derive(Component, Copy, Clone, PartialEq, Eq, Default, Debug)]
 pub struct IsFlat(pub bool);
 
+#[derive(Component, Debug)]
+pub struct ClientLayerMask(pub u64);
+
+impl Default for ClientLayerMask {
+    fn default() -> Self {
+        Self(0)
+    }
+}
+
+impl ClientLayerMask {
+    pub fn set(&mut self, layer: u8, enabled: bool) {
+        if enabled {
+            self.0 |= 1 << layer;
+        } else {
+            self.0 &= !(1 << layer);
+        }
+    }
+
+    pub fn get(&self, layer: u8) -> bool {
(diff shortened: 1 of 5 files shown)


Bafbi:
https://github.com/valence-rs/valence/assets/26836027/ce2119f3-897e-4f2e-b44d-f1658fc558ab

I do have some probleme with the sync of the entity when I toggle the layer

dyc3 (changes requested):
I'd like to see some unit tests for this.

rj00a:
Currently working on a redesign of Instances and Chunks to fix a number of issues. I'd like to finish my redesign before taking a closer look at this, since it may have a large impact on the approach taken here. It should be finished in the next day or two.

crates/valence_client/src/layer.rs
dyc3, line 1: Just spitballing: Could we make the `u8` in `ClientLayerSet` and `Layer` generic `where T: Hash`? It would allow for more than 256 layers, and multiple layer sets.

Bafbi: sure, can do

Bafbi: I don't think it's a good idea, we should use a u32 possibly but not a generic even if it's possible like what append if someone use multiple type of layer it become a mess

dyc3: Fair enough.

tachibanayui:
> Currently working on a redesign of Instances and Chunks to fix a number of issues. I'd like to finish my redesign before taking a closer look at this, since it may have a large impact on the approach taken here. It should be finished in the next day or two.

Visibility Layers will impact a lot of features IMO. I think we should take our time and make it right the first time or it gonna be expensive to fix later.

crates/valence_client/src/layer.rs
```diff
@@ -0,0 +1,76 @@
...
+/// The entity that are on layer 1 and 2 will be sent to the client.
+/// as well as the entity that don't have a `Layer` component.
+#[derive(Component, Default, Debug)]
+pub struct ClientLayerSet {
+    pub layers: HashSet<LayerType>,
+    old_layers: HashSet<LayerType>,
+}
+
+impl ClientLayerSet {
+    pub fn new(layers: Vec<LayerType>) -> Self {
```
tachibanayui, line 28: Should `layers` be of type `impl Iterator<Output = LayerType>` to avoid heap allocation?

crates/valence_client/src/lib.rs
```diff
@@ -934,13 +960,33 @@ fn read_data_in_old_view(
...
+                                    if let Some(layer) = layer {
+                                        if client_layer_set.contains(&layer.0) {
+                                            remove_buf.push(entity_id.get());
+                                        }
+                                    } else {
+                                        remove_buf.push(entity_id.get());
+                                    }
+                                } else {
+                                    remove_buf.push(entity_id.get());
+                                }
```
tachibanayui, line 975: Nitpicking: group ifs `if let (Some(client_layer_set), Some(layer)) = (client_layer_set, layer)` so that we can remove the duplicate branch like you did earlier

crates/valence_client/src/lib.rs
```diff
@@ -960,6 +1006,62 @@ fn read_data_in_old_view(
...
+
+            // Send entity spawn packets for entities that are in an entered layer and
+            // already in the client's view.
+            for (entity, &old_pos, layer) in entities.iter() {
+                if view.contains(old_pos.chunk_pos()) {
+                    if let (Some(client_layer_set), Some(layer)) = (client_layer_set, layer) {
+                        if client_layer_set
+                            .added()
+                            .collect::<HashSet<&LayerType>>()
+                            .contains(&layer.0)
```
tachibanayui, line 1034: Maybe instead of collecting it into a `HashSet`, we use `find` to save memory allocation and need to read through all items returned by `added()`?
```rust
if client_layer_set
    .added()
    .find(|x| **x == layer.0)
    .is_some()
```

crates/valence_client/src/lib.rs
```diff
@@ -960,6 +1006,62 @@ fn read_data_in_old_view(
...
+            for (entity_init_item, &old_pos, layer) in entities.iter() {
+                if view.contains(old_pos.chunk_pos()) {
+                    if let (Some(client_layer_set), Some(layer)) = (client_layer_set, layer) {
+                        // if client_layer_mask.0.ne(&old_layer_mask.0) {debug!("client_layer_mask:
+                        // {:?}, old_client_layer_mask: {:?}, layer: {:?}", client_layer_mask,
+                        // old_client_layer_mask, layer);}
+                        if client_layer_set
+                            .removed()
+                            .collect::<HashSet<&LayerType>>()
+                            .contains(&layer.0)
```
tachibanayui, line 1053: Maybe instead of collecting it into a `HashSet`, we use `find` to save memory allocation and need to read through all items returned by `added()`?
```rust
if client_layer_set
    .removed()
    .find(|x| **x == layer.0)
    .is_some()
```

crates/valence_client/src/lib.rs
```diff
@@ -960,6 +1006,62 @@ fn read_data_in_old_view(
...
 
+fn update_layer_view(
+    mut clients: Query<(
+        &mut Client,
+        &mut EntityRemoveBuf,
+        &OldPosition,
+        &OldViewDistance,
+        Option<&ClientLayerSet>,
+    )>,
+    entities: Query<(EntityInitQuery, &OldPosition, Option<&Layer>)>,
```
tachibanayui, line 1017: Why do we need to query for `Option<&Layer>` and `Option<&ClientLayerSet>` if we don't handle the `None` variant?

crates/valence_instance/src/lib.rs
```diff
@@ -447,6 +470,12 @@ pub struct PartitionCell {
     /// A cache of packets to send to all clients that are in view of this cell
     /// at the end of the tick.
     pub packet_buf: Vec<u8>,
+    /// A cache for packets that are layer specifics.
+    #[doc(hidden)]
+    pub layers_packet_buf: Vec<Vec<u8>>,
+    /// Mapping from layer to index in `layers_packet_buf`.
+    #[doc(hidden)]
+    pub layers_packet_buf_indices: FxHashMap<LayerType, usize>,
```
tachibanayui, line 478: Why do we need a Map between LayerType and a Vec of packet buffer instead of something like `FxHashMap<LayerType, Vec<u8>>`?

crates/valence_core/src/layer.rs
```diff
@@ -0,0 +1,11 @@
...
+
+/// The type use to represent a layer.
+/// u32 is used to allow for +4 billion layers.
+pub type LayerType = u32; // could be define behind a feature flag
+
+/// Used in conjunction with the `ClientLayerSet` component.  
+/// Example: you have `Layer(0)` on a cow, it will only be visible to clients
+/// with the access to layer 0 in their `ClientLayerSet`.
+#[derive(Component, Copy, Clone, PartialEq, Eq, Debug)]
+pub struct Layer(pub LayerType);
```
tachibanayui, line 11: Should an Entity be on multiple layers? It could be useful for interactions between different entities such as collisions, ai behaviors, hit-scanning

tachibanayui (commented):
Nice work on implementing Visibility Layers. I have a few suggestions and design choices to consider.

rj00a:
Finished the chunk/instance redesign in #402. Let me know if you have any issues.

rj00a:
Here's my take on visibility layers: #424.

PR #369, opened 2023-06-16
new packet inspector

### LINKED ISSUE #346: Update `tools/packet_inspector`
#335 changed the packet API, but `tools/packet_inspector` has not yet been updated in response.  It is currently excluded in `Cargo.toml` and should be added back to the project.

While we're at it, the packet inspector ought to be simplified a bit. The non-gui mode is not super useful and the "save packet histoy to file" feature causes issues when Minecraft updates happen (because all the packet IDs get shifted around). Overall, I think it could be rewritten to be a bit simpler while preserving its current behavior.

AviiNL:
## Description

... changes... wel, lets see... uuhhmmm, i rewrote the entire packet inspector :)

(also dont mind the snarkyness of that sentence, im tired)

closes #346


dyc3 (approved):
LGTM, good work.

PR #370, opened 2023-06-17
world time implementation

tachibanayui:
## Description
Implementation of world time for instance. Support sets the time explicitly, advances the time and adjusts the frequency that the server sends `WorldTimeUpdateS2c` packet to clients.

<details>
<summary>Playground:</summary>

```rust
use valence::client::hand_swing::HandSwingEvent;
use valence::client::message::SendMessage;
use valence::inventory::HeldItem;
use valence::prelude::*;
use valence::world_time::{ChangeTrackingTimeBroadcast, LinearTimeTicking, WorldTime};

#[allow(unused_imports)]
use crate::extras::*;

const SPAWN_Y: i32 = 64;

        ..Default::default()
    })
    .add_plugins(DefaultPlugins)
    .add_startup_system(setup)
    .add_system(init_clients)
    .add_system(handle_adjust_speed)
    .add_system(handle_display_time)
    .add_system(toggle_gamemode_on_sneak.in_schedule(EventLoopSchedule));
}

fn setup(
    mut commands: Commands,
    server: Res<Server>,
    biomes: Res<BiomeRegistry>,
) {
    let mut instance = Instance::new(ident!("overworld"), &dimensions, &biomes, &server);

    for z in -5..5 {
        for x in -5..5 {
        }
    }

    for z in -25..25 {
        for x in -25..25 {
        }
    }

    commands.spawn(instance).insert((
        WorldTime::default(),
        ChangeTrackingTimeBroadcast,
        LinearTimeTicking { speed: 1 },
    ));
}

fn init_clients(
    mut clients: Query<(&mut Client, &mut Location, &mut Position, &mut Inventory), Added<Client>>,
) {
    for (mut client, mut loc, mut pos, mut inv) in &mut clients {
        loc.0 = instances.single();
        pos.set([0.5, SPAWN_Y as f64 + 1.0, 0.5]);

        client.send_chat_message("Let's control time!");
        client.send_chat_message("- Left click the clock to make time run faster");
        client.send_chat_message("- Left click the trident to make time run slower");

        inv.set_slot(36, Some(ItemStack::new(ItemKind::Trident, 1, None)));
        inv.set_slot(37, Some(ItemStack::new(ItemKind::Clock, 1, None)));
    }
}

fn handle_adjust_speed(
    mut instances: Query<&mut LinearTimeTicking, With<Instance>>,
    clients: Query<(&Inventory, &HeldItem)>,
    mut events: EventReader<HandSwingEvent>,
) {
    for e in events.iter() {
        let Ok((inv, inv_state)) = clients.get(e.client) else {
            continue;
        };

        let mut ltt = instances.single_mut();

        let slot = inv_state.slot();
        let Some(is) = inv.slot(slot) else {
            continue;
        };

        match is.item {
            ItemKind::Clock => ltt.speed += 1,
            ItemKind::Trident => ltt.speed -= 1,
            _ => (),
        }
    }
}

fn handle_display_time(
    mut clients: Query<&mut Client>,
    instances: Query<(&WorldTime, &LinearTimeTicking), With<Instance>>,
) {
    for mut client in clients.iter_mut() {
        let (time, ltt) = instances.single();

        client.send_action_bar_message(format!(
            "Time: {} / {} per tick",
            time.time_of_day, ltt.speed
        ));
    }
}

// Add more systems here!

``` 
</details>

example: `cargo run --package valence --example time_travel`
tests: `cargo test --package valence --lib -- tests::world_time`

**Related**
part of #210


crates/valence_world_time/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+//!         time.time_of_day
+//!             .set_current_day_time(DayPhase::Night.into());
+//!     }
+//! }
+//! ```
+//!
+//! ## Advacing the world time
+//! Time of day and world age can be ticked individually using
+//! [`LinearTimeTicking`] and [`LinearWorldAging`] respectively.
+//! Note: If these component don't meet your requirements
```
dyc3, line 42:
```suggestion
//! If these components don't meet your requirements
```

crates/valence_world_time/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+//! (eg: you need time increment follow a sine wave ~~for some reason~~)
+//! You can create your own ticking component by modify the respective
+//! fields
```
dyc3, line 45: nit
```suggestion
//! (eg: you need time increment follow a sine wave ~~for some reason~~),
//! you can tick the time yourself by modifying the respective
//! fields on [`WorldTime`].
```

dyc3 (changes requested):
Looks good. Just a few nitpicks.

crates/valence/examples/time_travel.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    commands.spawn(instance).insert((
+        WorldTime {
+            time_of_day: -1,
+            ..Default::default()
+        },
+        ChangeTrackingTimeBroadcast,
+        LinearTimeTicking { speed: 1 },
+    ));
```
rj00a, line 50:
```suggestion
    commands.spawn((
        instance,
        WorldTime {
            time_of_day: -1,
            ..Default::default()
        },
        ChangeTrackingTimeBroadcast,
        LinearTimeTicking { speed: 1 },
    ));
```

crates/valence_world_time/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+            (handle_linear_time_ticking, handle_linear_world_aging).in_set(CalculateWorldTimeSet),
+        )
+        .add_system(send_time_to_joined_players.before(FlushPacketsSet));
+    }
+}
+
+/// Base component for storing world time information
+#[derive(Component, Default)]
+pub struct WorldTime {
+    /// The amount of time in game tick the current world has passed
```
rj00a, line 142:
```suggestion
    /// The total number of game ticks that have passed in the current world.
```

crates/valence_world_time/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+        let client_ticking = self.client_time_ticking();
+        self.time_of_day = self.time_of_day.abs().wrapping_add(amount);
+        if self.time_of_day < 0 {
+            self.time_of_day = self.time_of_day + i64::MAX + 1;
+        }
+
+        self.set_client_time_ticking(client_ticking);
+    }
+
+    /// If the client advance world time without server update
```
rj00a, line 165:
```suggestion
    /// If the client advances world time locally without server updates.
```

crates/valence_world_time/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    pub fn client_time_ticking(&self) -> bool {
+        !self.time_of_day.is_negative()
+    }
+
+    /// Set will the client advance world time without server update
```
rj00a, line 170:
```suggestion
    /// Sets if the client advances world time locally without server updates.
```

crates/valence_world_time/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    /// Get the time part of `time_of_day`
+    pub fn current_day_time(&self) -> i64 {
+        self.time_of_day % DAY_LENGTH
+    }
+
+    /// Set the time part of `time_of_day`
+    /// Use the [`DayPhase`] enum to easily handle common time
+    /// of day events without the need to look up information in the wiki.
```
rj00a, line 186:
```suggestion
    /// Set the time part of `time_of_day`.
    ///
    /// Use the [`DayPhase`] enum to easily handle common time
    /// of day events without the need to look up information in the wiki.
```

crates/valence_world_time/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+            interval.last_broadcast = currect_tick;
+        }
+    }
+}
+
+fn handle_linear_time_ticking(
+    mut instances: Query<(&mut WorldTime, &LinearTimeTicking), With<Instance>>,
+) {
+    for (mut time, ltr) in instances.iter_mut() {
+        if ltr.speed != 0 {
```
rj00a, line 363:
```suggestion
    for (mut time, ltt) in instances.iter_mut() {
        if ltt.speed != 0 {
```

crates/valence_world_time/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+) {
+    for (mut ins, time, mut interval) in instances.iter_mut() {
+        let currect_tick = server.current_tick();
+        if currect_tick - interval.last_broadcast >= interval.broadcast_rate {
+            ins.write_packet(&WorldTimeUpdateS2c {
+                time_of_day: time.time_of_day,
+                world_age: time.world_age,
+            });
+
+            interval.last_broadcast = currect_tick;
```
rj00a, line 354:
```suggestion
        let current_tick = server.current_tick();
        if current_tick - interval.last_broadcast >= interval.broadcast_rate {
            ins.write_packet(&WorldTimeUpdateS2c {
                time_of_day: time.time_of_day,
                world_age: time.world_age,
            });

            interval.last_broadcast = current_tick;
```

crates/valence/examples/time_travel.rs
```diff
@@ -0,0 +1,108 @@
...
+    for z in -25..25 {
+        for x in -25..25 {
+            instance.set_block([x, SPAWN_Y, z], BlockState::GRASS_BLOCK);
+        }
+    }
+
+    commands.spawn((
+        instance,
+        WorldTime {
+            time_of_day: -1,
```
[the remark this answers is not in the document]
tachibanayui: Vanilla server will set time_of_day to `-1` if you use `/gamerule doDaylightCycle false` and `/time set 0`

crates/valence_world_time/src/lib.rs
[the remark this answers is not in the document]
tachibanayui: I may have the wrong idea about bundles but I am not sure if bundling them is a good idea because most of the time they should be picked individually depending on the requirements, and I haven't found a good default for the bundle. Also, vanilla server only sends update time packets on an interval, even if you do `/time set day`.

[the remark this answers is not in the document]
rj00a: Personally I expect the defaults to be:
- world and day time is incremented every tick, starting at zero.
- Packets are broadcast once every second.
- Any manual changes to `WorldTime` are broadcast immediately and everything continues to work.

I think users would expect it to work that way too, and that's what `WorldTimeBundle::default()` should express.

> Also, vanilla server only sends update time packets on an interval, even if you do /time set day.

If anything that seems like an oversight on Mojang's part. I think users would appreciate having their changes be applied immediately. I can't think of a situation where you wouldn't want that behavior.

crates/valence/examples/time_travel.rs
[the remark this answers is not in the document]
rj00a: Maybe you could make that more explicit here by using `set_client_time_ticking`. But it looks like `set_client_time_ticking` is wrong when the time of day is zero because `-0.abs()` is still `0`.

crates/valence_world_time/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+        if self.time_of_day < 0 {
+            self.time_of_day = self.time_of_day + i64::MAX + 1;
+        }
+
+        self.set_client_time_ticking(client_ticking);
+    }
+
+    /// If the client advance world time without server update
+    pub fn client_time_ticking(&self) -> bool {
+        !self.time_of_day.is_negative()
```
rj00a, line 167: I think `self.time_of_day >= 0` would be easier to understand here.

crates/valence/examples/time_travel.rs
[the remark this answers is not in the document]
tachibanayui: Yeah, currently there is no way around that I think. Mojang devs probably knew it, so they use -1 as a substitute

crates/valence_world_time/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    /// Use the [`DayPhase`] enum to easily handle common time
+    /// of day events without the need to look up information in the wiki.
+    pub fn set_current_day_time(&mut self, time: i64) {
+        self.time_of_day = self.day() * DAY_LENGTH + time % DAY_LENGTH;
+    }
+
+    /// Get the current day part of `time_of_day`
+    pub fn day(&self) -> i64 {
+        self.time_of_day / DAY_LENGTH
+    }
```
rj00a, line 194: `self.time_of_day % DAY_LENGTH` and `self.time_of_day / DAY_LENGTH` return negative values when `self.time_of_day` is negative. Was that intentional?

rj00a: Some precise documentation on what exactly is returned here would be nice.

tachibanayui: No, that's an oversight

rj00a (commented):
One problem I see here is that the `LinearTimeTicking` and `LinearWorldAging` components do not mix well with the `ChangeTrackingTimeBroadcast` component, since it constantly triggers change detection and spams the client with world time packets.

tachibanayui:
> One problem I see here is that the `LinearTimeTicking` and `LinearWorldAging` components do not mix well with the `ChangeTrackingTimeBroadcast` component, since it constantly triggers change detection and spams the client with world time packets. This is problematic for `WorldTimeBundle`.

I might have to redesign the API a bit. I will create a Broadcaster kind of a mix of the 2 current broadcasters.
```rust
struct ScheduledBroadcaster {
    pub next_tick: i64;
    pub interval: u64;
}
```

And a prebuilt query to provide some API sugarily. I want to separate `WorldTime` and `ScheduledBroadcaster` because the user can choose the level of abstraction they want
```rust
struct TimeControllingQuery {
    pub time: &mut 'static WorldTime,
    pub schedule: &mut 'static ScheduledBroadcaster 
}

impl TimeControllingQuery {
    pub set_time(&mut self, time: i64) { ... }
    pub set_time_defered(&mut self, time: i64) { ... }
}
```

to change the time we can do:
```rust
let mut query: TimeControllingQuery = ...;
// Update immediately
query.set_time(10);

// Scheduled
for i in 11..20 {
    query.set_time_defered(i);
}
```

What do you think about this API?

Issue #337, opened 2023-05-23
Boss Bar API


rj00a:
## Describe the problem related to your feature request.

There is no API for [boss bars](https://minecraft.fandom.com/wiki/Commands/bossbar).

## What solution would you like?

Create a new crate called `valence_boss_bar`. Every boss bar is an entity with the following components:
- `BossBarTitle`
- `BossBarHealth`
- `BossBarStyle` (contains color and division)
- `BossBarFlags`
- `Uuid` (defined in `valence_entity` currently)
- `BossBarViewers`. Contains the set of client entities viewing this boss bar. When the boss bar is despawned with the `Despawned` component, the boss bar is removed from all viewers.

Change detection on the components is used to send packets to the boss bar viewers.

Also might be good to use another component containing a `Vec<u8>` to cache boss bar packets before sending them to viewers.

## What alternative(s) have you considered?

When entity relations in `bevy_ecs` become a thing, the `BossBarViewers` component could probably be replaced with a `boss bar -> client` relation for each viewer.

## Additional context

See the `BossBarS2c` packet.

tarunsamanta2k20:
@rj00a I am interested to work on this issue. But I need some help.

dyc3:
Here some some references/pointers to get you started:

- The relevant packet is documented here: https://wiki.vg/Protocol#Boss_Bar
- The packet is already defined here: https://github.com/valence-rs/valence/blob/main/crates/valence_core/src/packet/s2c/play/boss_bar.rs

All you need to do is make the components and implement the systems to actually send the packets, preferably with unit tests.

JackCrumpLeys:
This should be closed.

PR #371, opened 2023-06-17
Boss Bar API

pepperoni21:
## Description

Hey, I started working on this issue https://github.com/valence-rs/valence/issues/337.

It's the first time I'm trying to contribute to this project, and I'm not yet very familiar with Bevy ECS. I tried to do what I could but now I have some questions.

### So far:
- I created a crate "**valence_boss_bar**" and I moved this packet there https://github.com/valence-rs/valence/blob/2ed5a8840da244d8f1f91f1df43750664952be46/crates/valence_core/src/protocol/packet.rs#L612
- I initialized the plugin and added a system that removes the players when a boss bar is despawned.
- I created/moved the boss bar components (BossBarTitle, BossBarHealth, BossBarStyle, BossBarFlags and BossBarViewers)
- I created a test file which is not done yet.

### Here are the questions I have:
- I tried to store Client objects directly inside the BossBarViewers struct but I realized it was a bad idea, now I just store bevy entities, is it how I should do?
- How should I manage to add/remove viewers or update titles/health/style in a BossBar? I implemented methods for the BossBar struct to make these actions but I can't figure out how to send packets to the clients because I don't know how to turn my vector of bevy entities into Client objects.
- I basically forked my test file from the world_border's one, is it how I should spawn a BossBar?
```rust
app.world.entity_mut(instance_ent).insert(
            BossBar::new(Text::text("Test"),
                BossBarColor::Red,
                BossBarDivision::SixNotches,
                BossBarFlags::new()));
```

### What (I think) is left to do
- Finding a way to send packets to viewers
- Documenting the code
- Make the unit tests

### DIFF
--- a/crates/valence/src/tests/boss_bar.rs
+++ b/crates/valence/src/tests/boss_bar.rs
@@ -1,0 +1,48 @@
+use bevy_app::App;
+use bevy_ecs::entity::Entity;
+use valence_boss_bar::components::{BossBar, BossBarColor, BossBarDivision, BossBarFlags};
+use valence_core::text::Text;
+use valence_instance::Instance;
+
+use super::{MockClientHelper, scenario_single_client, create_mock_client};
+
+/*#[test]
+fn test_intialize_on_join() {
+    let mut app = App::new();
+    let (_, instance_ent) = prepare(&mut app);
+
+    let (mut client, mut client_helper) = create_mock_client();
+    let client_ent = app.world.spawn(client).id();
+
+}*/
+
+fn prepare(app: &mut App) -> (MockClientHelper, Entity) {
+    let (_, mut client_helper) = scenario_single_client(app);
+
+    // Process a tick to get past the "on join" logic.
+    app.update();
+    client_helper.clear_sent();
+
+    // Get the instance entity.
+    let instance_ent = app
+        .world
+        .iter_entities()
+        .find(|e| e.contains::<Instance>())
+        .expect("could not find instance")
+        .id();
+
+    // Insert a boss bar to the instance.
+    app.world
+        .entity_mut(instance_ent)
+        .insert(BossBar::new(Text::text(""),
+                                    BossBarColor::Red,
+                                    BossBarDivision::SixNotches,
+                                    BossBarFlags::new()));
+    
+    for _ in 0..2 {
+        app.update();
+    }
+
+    client_helper.clear_sent();
+    (client_helper, instance_ent)
+}
--- a/crates/valence_boss_bar/Cargo.toml
+++ b/crates/valence_boss_bar/Cargo.toml
@@ -1,10 +1,11 @@
 [package]
 name = "valence_boss_bar"
 description = "Boss bar API for Valence"
-version = "0.1.0"
-edition = "2021"
 readme = "README.md"
 keywords = ["minecraft", "bossbar", "api"]
+documentation.workspace = true
+version.workspace = true
+edition.workspace = true
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
--- a/crates/valence_boss_bar/Cargo.toml
+++ b/crates/valence_boss_bar/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "valence_boss_bar"
+description = "Boss bar API for Valence"
+version = "0.1.0"
+edition = "2021"
+readme = "README.md"
+keywords = ["minecraft", "bossbar", "api"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+valence_core.workspace = true
+valence_network.workspace = true
+valence_entity.workspace = true
+uuid.workspace = true
+bitfield-struct.workspace = true
+bevy_app.workspace = true
+bevy_ecs.workspace = true
--- a/crates/valence_boss_bar/Cargo.toml
+++ b/crates/valence_boss_bar/Cargo.toml
@@ -12,6 +12,7 @@ keywords = ["minecraft", "bossbar", "api"]
 valence_core.workspace = true
 valence_network.workspace = true
 valence_entity.workspace = true
+valence_client.workspace = true
 uuid.workspace = true
 bitfield-struct.workspace = true
 bevy_app.workspace = true
--- a/crates/valence_boss_bar/src/components.rs
+++ b/crates/valence_boss_bar/src/components.rs
@@ -1,0 +1,101 @@
+use bevy_ecs::prelude::{Component, Entity};
+use valence_core::{uuid::UniqueId, text::Text, protocol::{Encode, Decode}};
+use bitfield_struct::bitfield;
+
+#[derive(Component)]
+pub struct BossBar {
+    pub id: UniqueId,
+    pub title: BossBarTitle,
+    pub health: BossBarHealth,
+    pub style: BossBarStyle,
+    pub flags: BossBarFlags,
+    pub viewers: BossBarViewers,
+}
+
+impl BossBar {
+
+    pub fn new(title: Text, color: BossBarColor, division: BossBarDivision, flags: BossBarFlags) -> BossBar {
+        BossBar {
+            id: UniqueId::default(),
+            title: BossBarTitle(title),
+            health: BossBarHealth(1.0),
+            style: BossBarStyle {
+                color,
+                division,
+            },
+            flags,
+            viewers: BossBarViewers(Vec::new()),
+        }
+    }
+
+    pub fn update_health(&mut self, health: f32) {
+        self.health.0 = health;
+    }
+
+    pub fn update_title(&mut self, title: Text) {
+        self.title.0 = title;
+    }
+
+    pub fn update_style(&mut self, color: BossBarColor, division: BossBarDivision) {
+        self.style.color = color;
+        self.style.division = division;
+    }
+
+    pub fn update_flags(&mut self, flags: BossBarFlags) {
+        self.flags = flags;
+    }
+
+    pub fn add_client(&mut self, entity: Entity) {
+        self.viewers.0.push(entity);
+    }
+
+    pub fn remove_client(&mut self, entity: Entity) {
+        self.viewers.0.retain(|e| *e != entity);
+    }
+
+}
+
+#[derive(Component, Clone)]
+pub struct BossBarTitle(pub Text);
+
+#[derive(Component)]
+pub struct BossBarHealth(pub f32);
+
+#[derive(Component)]
+pub struct BossBarStyle {
+    pub color: BossBarColor,
+    pub division: BossBarDivision,
+}
+
+#[derive(Component, Copy, Clone, PartialEq, Eq, Debug, Encode, Decode)]
+pub enum BossBarColor {
+    Pink,
+    Blue,
+    Red,
+    Green,
+    Yellow,
+    Purple,
+    White,
+}
+
+#[derive(Component, Copy, Clone, PartialEq, Eq, Debug, Encode, Decode)]
+pub enum BossBarDivision {
+    NoDivision,
+    SixNotches,
+    TenNotches,
+    TwelveNotches,
+    TwentyNotches,
+}
+
+#[bitfield(u8)]
+#[derive(Component, PartialEq, Eq, Encode, Decode)]
+pub struct BossBarFlags {
+    pub darken_sky: bool,
+    pub dragon_bar: bool,
+    pub create_fog: bool,
+    #[bits(5)]
+    _pad: u8,
+}
+
+#[derive(Component)]
+pub struct BossBarViewers(pub Vec<Entity>);
--- a/crates/valence_boss_bar/src/lib.rs
+++ b/crates/valence_boss_bar/src/lib.rs
@@ -1,4 +1,5 @@
 use bevy_ecs::prelude::Component;
+use valence_client::Client;
 use valence_core::{protocol::{Decode, Encode}, text::Text, uuid::UniqueId};
 use bitfield_struct::bitfield;
 
@@ -7,12 +8,17 @@ pub mod packet;
 pub struct BossBar {
     pub id: UniqueId,
     pub title: BossBarTitle,
+    pub health: BossBarHealth,
     pub style: BossBarStyle,
     pub flags: BossBarFlags,
+    pub viewers: BossBarViewers,
 }
 
 #[derive(Component)]
 pub struct BossBarTitle(pub Text);
+
+#[derive(Component)]
+pub struct BossBarHealth(pub f32);
 
 #[derive(Component)]
 pub struct BossBarStyle {
@@ -51,4 +57,4 @@ pub struct BossBarFlags {
 }
 
 #[derive(Component)]
-pub struct BossBarViewers(pub Vec<u128>);
+pub struct BossBarViewers(pub Vec<Client>);
--- a/crates/valence_boss_bar/src/lib.rs
+++ b/crates/valence_boss_bar/src/lib.rs
@@ -1,60 +1,29 @@
-use bevy_ecs::prelude::Component;
+use bevy_app::Plugin;
+use bevy_ecs::{system::Query, query::Added};
+use components::{BossBar, BossBarViewers};
 use valence_client::Client;
-use valence_core::{protocol::{Decode, Encode}, text::Text, uuid::UniqueId};
-use bitfield_struct::bitfield;
+use valence_core::{despawn::Despawned, protocol::encode::WritePacket};
 
 pub mod packet;
+pub mod components;
 
-pub struct BossBar {
-    pub id: UniqueId,
-    pub title: BossBarTitle,
-    pub health: BossBarHealth,
-    pub style: BossBarStyle,
-    pub flags: BossBarFlags,
-    pub viewers: BossBarViewers,
+pub struct BossBarPlugin;
+
+impl Plugin for BossBarPlugin {
+
+    fn build(&self, app: &mut bevy_app::App) {
+        app
+        .add_system(remove_despawned_boss_bars_from_viewers);
+    }
+
 }
 
-#[derive(Component)]
-pub struct BossBarTitle(pub Text);
-
-#[derive(Component)]
-pub struct BossBarHealth(pub f32);
-
-#[derive(Component)]
-pub struct BossBarStyle {
-    pub color: BossBarColor,
-    pub division: BossBarDivision,
+fn remove_despawned_boss_bars_from_viewers(mut bossbar_viewers: Query<(&BossBar, &mut BossBarViewers), Added<Despawned>>, mut clients: Query<&mut Client>) {
+    for (boss_bar, mut viewers) in bossbar_viewers.iter_mut() {
+        let viewers = &mut viewers.0;
+        for viewer in viewers.iter_mut() {
+            let mut client = clients.get_mut(*viewer).unwrap();
+            client.write_packet(&boss_bar.generate_remove_packet());
+        }
+    }
 }
-
-#[derive(Component, Copy, Clone, PartialEq, Eq, Debug, Encode, Decode)]
-pub enum BossBarColor {
-    Pink,
-    Blue,
-    Red,
-    Green,
-    Yellow,
-    Purple,
-    White,
-}
-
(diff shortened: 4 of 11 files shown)


crates/valence_boss_bar/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+
+fn remove_despawned_boss_bars_from_viewers(mut bossbar_viewers: Query<(&BossBar, &mut BossBarViewers), Added<Despawned>>, mut clients: Query<&mut Client>) {
+    for (boss_bar, mut viewers) in bossbar_viewers.iter_mut() {
+        let viewers = &mut viewers.0;
+        for viewer in viewers.iter_mut() {
+            let mut client = clients.get_mut(*viewer).unwrap();
+            client.write_packet(&boss_bar.generate_remove_packet());
+        }
+    }
+}
```
Bafbi, line 29: In the same line, that `BossBar` should be a Bundle, here you query the `BossBar` component that contain the `BassBarViewers` and query the `BassBarViewers` a second time.
If you want to have impl on your query you can use `WordQuery` like here : https://github.com/valence-rs/valence/blob/2ed5a8840da244d8f1f91f1df43750664952be46/crates/valence_client/src/lib.rs#L812-L873

Bafbi (changes requested):

Bafbi:
> I tried to store Client objects directly inside the BossBarViewers struct but I realized it was a bad idea, now I just store bevy entities, is it how I should do?
Yes it's like in a db you put the id of what you want to ref.

> How should I manage to add/remove viewers or update titles/health/style in a BossBar? I implemented methods for the BossBar struct to make these actions but I can't figure out how to send packets to the clients because I don't know how to turn my vector of bevy entities into Client objects.
When you use `write_packet` on the client, it's being put on a buffer to be send.

> I basically forked my test file from the world_border's one, is it how I should spawn a BossBar?
No, BossBar are not link to the instance, but you should spawn a new bevy_entity
```rust
commands.spawn(BossBarBundle::new(...));
```

dyc3:
Regarding your questions

> I tried to store Client objects directly inside the BossBarViewers struct but I realized it was a bad idea, now I just store bevy entities, is it how I should do?

Yes, that's how you should do it. Otherwise, you would be attempting to violate rust's aliasing rules.

>How should I manage to add/remove viewers or update titles/health/style in a BossBar? I implemented methods for the BossBar struct to make these actions but I can't figure out how to send packets to the clients because I don't know how to turn my vector of bevy entities into Client objects.

You need to query for clients separately, and then you can use the `Entity` to look up the `Client`.

```rust
fn foo(clients: Query<Client>) {
    // Then you can look up the client using the entity
    clients.get(entity)
```

> I basically forked my test file from the world_border's one, is it how I should spawn a BossBar?

For unit tests, operating directly on `app.world` is fine, but otherwise use bevy's `Commands`. You'll probably want to create a new entity instead of using an existing one.

pepperoni21:
Hey thanks a lot for your help!
> 
> You need to query for clients separately, and then you can use the `Entity` to look up the `Client`.
> 
> ```rust
> fn foo(clients: Query<Client>) {
>     // Then you can look up the client using the entity
>     clients.get(entity)
> ```
So I have to do that in a system? But isn't a system supposed to be executed every tick? Or I should use something like an event?
>
> For unit tests, operating directly on `app.world` is fine, but otherwise use bevy's `Commands`. You'll probably want to create a new entity instead of using an existing one.

Fine I will do that!

crates/valence_boss_bar/src/components.rs
```diff
@@ -0,0 +1,76 @@
...
+pub struct BossBarFlags {
+    pub darken_sky: bool,
+    pub dragon_bar: bool,
+    pub create_fog: bool,
+    #[bits(5)]
+    _pad: u8,
+}
+
+#[derive(Component)]
+pub struct BossBarViewers(pub Vec<Entity>);
```
dyc3, line 76: I'm somewhat on the fence about having this component, because we're very close to having visibility layers. #362

Bafbi:
> Hey thanks for your help! The problem I have is that I don't know how to modify something in the Bundle. I could simply make a method to change a property but how am I supposed to send the packets then? Because since it's not a system I can't query anything right?

> So I have to do that in a system? But isn't a system supposed to be executed every tick? Or I should use something like an event?

You want to send the packet when the bossbar is created or that an component change for that you can use [changeDetection](https://bevy-cheatbook.github.io/programming/change-detection.html) 
```rust
fn foo(clients: Query<Client>, bossbar: Query<(UniqueId, ...), Added<BossBarBundle>>) { // not sure if you can put a bundle here, but if you can't use a component that uniquely identify a bossbar
    // you generates packets for each bossbar
    // Then you can look up each client of the bossbar to send them like you did before
}
```

Bafbi:
> Hey thanks for your help! The problem I have is that I don't know how to modify something in the Bundle. I could simply make a method to change a property but how am I supposed to send the packets then? Because since it's not a system I can't query anything right?

you don't make method to modify the component.
When someone use your plugin they can make system to modify what they want :
// Someone server code
```rust
fn update_bossbar(mob: Query<&Hp, &BossBarLink, Change<Hp>>, bossbar: Query<&mut BossBarHealth>) {
    for (hp, link) in mob {
        if let Ok(bossbar_health) = bossbar.get(link) {
            bossbat_health -= hp;
        }
    }
}
```

crates/valence_boss_bar/src/components.rs
pepperoni21: You mean the visibility layers could replace this? If they can then I will try to integrate that once your PR is merged

pepperoni21:
> > So I have to do that in a system? But isn't a system supposed to be executed every tick? Or I should use something like an event?
> 
> You want to send the packet when the bossbar is created or that an component change for that you can use [changeDetection](https://bevy-cheatbook.github.io/programming/change-detection.html)
> 
> ```rust
> fn foo(clients: Query<Client>, bossbar: Query<(UniqueId, ...), Added<BossBarBundle>>) { // not sure if you can put a bundle here, but if you can't use a component that uniquely identify a bossbar
>     // you generates packets for each bossbar
>     // Then you can look up each client of the bossbar to send them like you did before
> }
> ```

That's exactly what I was looking for, thanks

crates/valence_boss_bar/src/components.rs
Bafbi: Agreed

crates/valence/src/tests/boss_bar.rs
```diff
@@ -1,47 +1,132 @@
...
+    boss_bar.current_viewers.push(client_ent);
+
+    app.update();
+
+    app.world.entity_mut(instance_ent).insert(Despawned);
+
+    app.update();
+
+    let frames = client_helper.collect_sent();
+    frames.assert_count::<BossBarS2c>(2);
```
pepperoni21, line 37: I don't know if there's a way to view the actual content of the packets. I'm just checking the count now.

crates/valence_boss_bar/src/components.rs
```diff
@@ -73,4 +73,18 @@ pub struct BossBarFlags {
 }
 
 #[derive(Component)]
-pub struct BossBarViewers(pub Vec<Entity>);
\ No newline at end of file
+pub struct BossBarViewers {
+    pub current_viewers: Vec<Entity>,
```
pepperoni21, line 77: I keep track of the previous viewers to know which one were added/removed when this component is changed.

crates/valence_boss_bar/src/lib.rs
```diff
@@ -14,16 +14,115 @@ impl Plugin for BossBarPlugin {
...
+        .add_system(handle_boss_bar_title_update)
+        .add_system(handle_boss_bar_health_update)
+        .add_system(handle_boss_bar_style_update)
+        .add_system(handle_boss_bar_flags_update)
+        .add_system(handle_boss_bar_viewers_update);
     }
 
 }
 
+fn handle_boss_bar_title_update(mut boss_bars: Query<(&UniqueId, &BossBarTitle, &mut BossBarViewers), Changed<BossBarTitle>>, mut clients: Query<&mut Client>) {
```
pepperoni21, line 27: I'm sure that's not the right way to make a query but I could not find how to query the bundle itself.

tachibanayui: I don't think there's a way to query bundles

crates/valence_boss_bar/src/lib.rs
```diff
@@ -1,34 +1,191 @@
...
-        for viewer in viewers.0.iter_mut() {
+        for viewer in viewers.current_viewers.iter_mut() {
             let mut client = clients.get_mut(*viewer).unwrap();
             client.write_packet(&BossBarS2c {
                 id: id.0,
                 action: BossBarAction::Remove,
             });
         }
     }
-}
```
Bafbi: You don't remove the bossbar from the world

crates/valence_boss_bar/src/components.rs
```diff
@@ -72,5 +82,13 @@ pub struct BossBarFlags {
...
+/// The viewers of a boss bar.
+#[derive(Component, Default)]
+pub struct BossBarViewers {
+    /// The current viewers of the boss bar. It is the list that should be
+    /// updated.
+    pub current_viewers: Vec<Entity>,
+    /// The viewers of the last tick in order to determine which viewers have
+    /// been added and removed.
+    pub last_viewers: Vec<Entity>,
+}
```
Bafbi, line 94: Maybe use the same naming of other place in the project :
- viewers
- old_viewers

Bafbi (changes requested):

crates/valence_boss_bar/src/lib.rs
pepperoni21: Isn't there a system that automatically removes entities from the world when they get the Despawned component? Otherwise I will do it, thanks.

crates/valence_boss_bar/src/components.rs
pepperoni21: I will, thanks!

crates/valence_boss_bar/src/lib.rs
Bafbi: Ho I didn't think of that, You are right it is done automatically
https://github.com/valence-rs/valence/blob/d36998dc9cba25aba4b9287db41ce69911708775/crates/valence_core/src/despawn.rs#L19-L26

rj00a: Yes, but you need to detect when the boss bar gains the `Despawned` component and send the appropriate packets to despawn the bar for all viewers.

crates/valence_boss_bar/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "valence_boss_bar"
+description = "Boss bar API for Valence"
+readme = "README.md"
+keywords = ["minecraft", "bossbar", "api"]
+documentation.workspace = true
+version.workspace = true
+edition.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
rj00a, line 11:
```suggestion
```

crates/valence_boss_bar/src/components.rs
```diff
@@ -0,0 +1,94 @@
...
+
+/// The viewers of a boss bar.
+#[derive(Component, Default)]
+pub struct BossBarViewers {
+    /// The current viewers of the boss bar. It is the list that should be
+    /// updated.
+    pub viewers: Vec<Entity>,
+    /// The viewers of the last tick in order to determine which viewers have
+    /// been added and removed.
+    pub old_viewers: Vec<Entity>,
```
rj00a, line 93: Access to these fields needs to be protected in some way to prevent misuse. Also consider using `BTreeSet` since the set of clients viewing a boss bar should not have duplicates.

crates/valence_boss_bar/src/lib.rs
rj00a, line 1: nit: no need for the `handle_` prefix on all the systems.

crates/valence_boss_bar/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+use bevy_ecs::removal_detection::RemovedComponents;
+use bevy_ecs::system::Query;
+use components::{BossBarFlags, BossBarHealth, BossBarStyle, BossBarTitle, BossBarViewers};
+use packet::{BossBarAction, BossBarS2c};
+use valence_client::Client;
+use valence_core::despawn::Despawned;
+use valence_core::protocol::encode::WritePacket;
+use valence_core::uuid::UniqueId;
+
+pub mod components;
```
rj00a, line 14: Consider making this module private and re-exporting everything publicly, since the only public item in the root of the crate is `BossBarPlugin`.

```suggestion
mod components;
pub use components::*;
```

crates/valence_boss_bar/src/lib.rs
rj00a, line 1: crate needs more lints.
```rust
#![allow(clippy::type_complexity)]
#![deny(
    rustdoc::broken_intra_doc_links,
    rustdoc::private_intra_doc_links,
    rustdoc::missing_crate_level_docs,
    rustdoc::invalid_codeblock_attributes,
    rustdoc::invalid_rust_codeblocks,
    rustdoc::bare_urls,
    rustdoc::invalid_html_tags
)]
#![warn(
    trivial_casts,
    trivial_numeric_casts,
    unused_lifetimes,
    unused_import_braces,
    unreachable_pub,
    clippy::dbg_macro
)]
```

crates/valence_boss_bar/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+
+/// System that sends a bossbar update title packet to all viewers of a boss bar
+/// that has had its title updated.
+fn handle_boss_bar_title_update(
+    mut boss_bars: Query<(&UniqueId, &BossBarTitle, &mut BossBarViewers), Changed<BossBarTitle>>,
+    mut clients: Query<&mut Client>,
+) {
+    for (id, title, mut boss_bar_viewers) in boss_bars.iter_mut() {
+        for viewer in boss_bar_viewers.viewers.iter_mut() {
+            let mut client = clients.get_mut(*viewer).unwrap();
```
rj00a, line 39: Don't `unwrap`. (issue repeated in other systems)

crates/valence_boss_bar/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+            .add_system(handle_boss_bar_viewers_update)
+            .add_system(handle_boss_bar_despawn)
+            .add_system(handle_client_disconnection);
+    }
+}
+
+/// System that sends a bossbar update title packet to all viewers of a boss bar
+/// that has had its title updated.
+fn handle_boss_bar_title_update(
+    mut boss_bars: Query<(&UniqueId, &BossBarTitle, &mut BossBarViewers), Changed<BossBarTitle>>,
```
rj00a, line 34: Does not need to be mutable. (issue repeated in other systems)

```suggestion
    mut boss_bars: Query<(&UniqueId, &BossBarTitle, &BossBarViewers), Changed<BossBarTitle>>,
```

crates/valence_boss_bar/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+        for removed_viewer in removed_viewers {
+            if let Ok(mut client) = clients.get_mut(removed_viewer) {
+                client.write_packet(&BossBarS2c {
+                    id: id.0,
+                    action: BossBarAction::Remove,
+                });
+            }
+        }
+
+        boss_bar_viewers.old_viewers = boss_bar_viewers.viewers.clone();
```
rj00a, line 157: There's some O(n²) work happening here that could be avoided with sets. Also the allocation of temporary `added_viewers` and `removed_viewers` vecs shouldn't be necessary.

crates/valence_boss_bar/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/// System that sends a bossbar remove packet to all viewers of a boss bar that
+/// has been despawned.
+fn handle_boss_bar_despawn(
+    mut boss_bars: Query<(&UniqueId, &mut BossBarViewers), Added<Despawned>>,
+    mut clients: Query<&mut Client>,
+) {
+    for boss_bar in boss_bars.iter_mut() {
+        let (id, mut viewers) = boss_bar;
```
rj00a, line 168:
```suggestion
    for (id, mut viewers) in boss_bars.iter_mut() {
```

crates/valence_boss_bar/src/packet.rs
```diff
@@ -0,0 +1,28 @@
...
+        color: BossBarColor,
+        division: BossBarDivision,
+        flags: BossBarFlags,
+    },
+    Remove,
+    UpdateHealth(f32),
+    UpdateTitle(Text),
+    UpdateStyle(BossBarColor, BossBarDivision),
+    UpdateFlags(BossBarFlags),
+}
```
rj00a, line 28: We can avoid cloning `Text` when encoding if we use `Cow<Text>` in the packet.

```suggestion
pub struct BossBarS2c<'a> {
    pub id: Uuid,
    pub action: BossBarAction<'a>,
}

#[derive(Clone, PartialEq, Debug, Encode, Decode)]
pub enum BossBarAction<'a> {
    Add {
        title: Cow<'a, Text>,
        health: f32,
        color: BossBarColor,
        division: BossBarDivision,
        flags: BossBarFlags,
    },
    Remove,
    UpdateHealth(f32),
    UpdateTitle(Cow<'a, Text>),
    UpdateStyle(BossBarColor, BossBarDivision),
    UpdateFlags(BossBarFlags),
}
```

rj00a (changes requested):

crates/valence_boss_bar/src/components.rs
rj00a, line 93: This needs to be private.

rj00a (commented):
Graph in `crates/README.md` needs an update.

crates/valence_boss_bar/src/components.rs
pepperoni21: How am I supposed to access them then? Should I use methods?

rj00a: You could make it `pub(crate)`

pepperoni21: If I make `viewers` `pub(crate)` we won't be able to add/remove players to a boss bar outside of the crate, but I can do it for `old_viewers`.

rj00a: Yep I meant `old_viewers` only.

crates/valence_boss_bar/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+
+/// System that removes a client from the viewers of its boss bars when it
+/// disconnects.
+fn client_disconnection(
+    disconnected_clients: Query<Entity, (With<Client>, Added<Despawned>)>,
+    mut boss_bars_viewers: Query<&mut BossBarViewers>,
+) {
+    for entity in disconnected_clients.iter() {
+        for mut boss_bar_viewers in boss_bars_viewers.iter_mut() {
+            boss_bar_viewers.viewers.retain(|viewer| *viewer != entity);
```
rj00a, line 206:
```suggestion
            boss_bar_viewers.viewers.remove(&entity);
```

crates/valence_boss_bar/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+    fn build(&self, app: &mut bevy_app::App) {
+        app.add_systems(
+            (
+                boss_bar_title_update,
+                boss_bar_health_update,
+                boss_bar_style_update,
+                boss_bar_flags_update,
+                boss_bar_viewers_update,
+                boss_bar_despawn,
+                client_disconnection,
```
rj00a, line 52: `client_disconnection` should be ordered before `boss_bar_viewers_update` so the updates are done on the same tick.

pepperoni21: Oh I see, I added that `client_disconnection.before(boss_bar_viewers_update)`, so sure that's how I should do though

rj00a: That's correct. Just need to `cargo fmt` and we can merge this.

rj00a (approved):

rj00a (changes requested):
`valence_boss_bar` needs to be exported as `boss_bar` in the main `valence` crate. Also, please add an example to `examples/`.

rj00a (approved):

rj00a:
Looks good, thanks.

PR #377, opened 2023-06-19
Valence tick

AviiNL:
## Description

Adapted the bevy_time implementation to be used with Ticks as usize instead of Time as Instant/Duration, allows you to run systems with a fixed tick interval.

Quick note: This does (still) use `CoreSchedule::FixedUpdate` and I am not sure that it should, I looked briefly in how to add a custom Schedule thing, but that went way over my head.
What I am missing, (also from bevy itself) is how to make multiple FixedTick to run separate systems, I don't think it's possible, I would prefer to assign a system to a schedule, where you can have systems run at different intervals without needing to resort to something like https://github.com/valence-rs/valence/pull/377/files#diff-50197256120b6495759f6a7d5ff7276b9716bc0deddc48b62e174c0762e4ac00R68
But that might just be a bevy generic limitation from what I can tell.

### DIFF
--- a/crates/valence/examples/fixed_tick.rs
+++ b/crates/valence/examples/fixed_tick.rs
@@ -1,0 +1,74 @@
+#![allow(clippy::type_complexity)]
+
+use valence::prelude::*;
+use valence_client::message::SendMessage;
+
+const SPAWN_Y: i32 = 64;
+
+pub fn main() {
+    tracing_subscriber::fmt().init();
+
+    App::new()
+        .add_plugins(DefaultPlugins)
+        .add_plugin(TickSystem)
+        .add_startup_system(setup)
+        .add_system(init_clients)
+        .add_system(despawn_disconnected_clients)
+        .add_system(run_every_20_ticks.in_schedule(CoreSchedule::FixedUpdate))
+        .add_system(manual_tick_interval)
+        .insert_resource(FixedTick::new(20))
+        .run();
+}
+
+fn setup(
+    mut commands: Commands,
+    server: Res<Server>,
+    dimensions: Res<DimensionTypeRegistry>,
+    biomes: Res<BiomeRegistry>,
+) {
+    let mut instance = Instance::new(ident!("overworld"), &dimensions, &biomes, &server);
+
+    for z in -5..5 {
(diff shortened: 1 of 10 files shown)


crates/valence/examples/fixed_tick.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    App::new()
+        .add_plugins(DefaultPlugins)
+        .add_plugin(TickSystem)
+        .add_startup_system(setup)
+        .add_system(init_clients)
+        .add_system(despawn_disconnected_clients)
+        .add_system(run_every_20_ticks.in_schedule(CoreSchedule::FixedUpdate))
+        .add_system(manual_tick_interval)
+        .insert_resource(FixedTick::new(20))
```
dyc3, line 19: Does this mean this only allows for one timer?

AviiNL: Functionally it works the same as bevy_timer, so yeah, i think so, which is why I added the other example, or want to run it on a seperate scheduler, or maybe even make it some form of dynamic scheduler if thats possible?

Bafbi:
Could be useful for #440

dyc3:
@Bafbi I don't think so. The behavior we'd want for rate limiting is more dynamic than just "send x packets every n ticks"

Bafbi:
> @Bafbi I don't think so. The behavior we'd want for rate limiting is more dynamic than just "send x packets every n ticks"

@dyc3  What do you mean? we would just not be able to the delay per entity (which is not really an indispensable feature) but would get the behavior wanted .
I would like to know what do you mean by *dynamic* ?

dyc3:
This is getting off topic for this PR, I'll respond in #440

dyc3:
Regardless in this PR's current state, it's only possible to have one `FixedTick`. Ideally, we'd have something that allows for more than 1 time interval to run systems at.

rj00a:
With only one `FixedTick` I don't see this having much use. Perhaps what you're looking for is something analogous to bevy's [`Timer`](https://docs.rs/bevy_time/latest/bevy_time/struct.Timer.html)?

PR #388, opened 2023-06-20
Broadcast to lan

### LINKED ISSUE #376: LAN support

LAN support allows easy broadcasting to your own device and other devices instead of having to manually put in the chosen port.

Valence should (definitely not high priority) add LAN broadcasting support ([wiki.vg](https://wiki.vg/Server_List_Ping#Ping_via_LAN_.28Open_to_LAN_in_Singleplayer.29))

Joining manually.

AviiNL:
## Description

~~I've added a `lan_broadcast_motd` which takes an `Option<String>` to the `NetworkSettings` to enable (or disable if `None`) the Broadcast to LAN functionality, which sends a UDP packet over the broadcast address `224.0.2.60:4445` to let LAN clients know the server exists.~~

~~I'm not too sure about the name of the field, nor the fact that it takes a String, would much rather see &str or something, but borrows and lifetimes still elude me a bunch.~~

Closes #376 

```rs
// Playground

#![allow(clippy::type_complexity)]

use valence::network::{async_trait, BroadcastToLan, ConnectionMode};
use valence::prelude::*;

#[allow(unused_imports)]
use crate::extras::*;

        callbacks: MyCallbacks.into(),
        ..Default::default()
    })
    .add_plugins(DefaultPlugins)
    .run();
}

struct MyCallbacks;

#[async_trait]
impl NetworkCallbacks for MyCallbacks {
    async fn broadcast_to_lan(&self, _shared: &SharedNetworkState) -> BroadcastToLan {
        BroadcastToLan::Enabled("Hello Valence!")
    }
}
```


rj00a (changes requested):
Doesn't seem to be working. Also there needs to be a way to enable and disable broadcasting after the server has started, probably via `SharedNetworkState`.

crates/valence_network/src/lib.rs
```diff
@@ -553,3 +579,39 @@ pub struct PlayerSampleEntry {
...
+        };
+
+        if socket.connect(server_addr).is_err() {
+            return;
+        }
+
+        let message = format!("[MOTD]{motd}[/MOTD][AD]{port}[/AD]");
+        if socket.send(message.as_bytes()).is_err() {
+            return;
+        }
```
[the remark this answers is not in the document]
AviiNL: Should it though, if there are no clients listening, wouldn't we want to ignore a "failed to connect" ? Although, since it's broadcast, shouldn't the error arm technically be unreachable?

crates/valence_network/src/lib.rs
```diff
@@ -334,6 +343,12 @@ pub trait NetworkCallbacks: Send + Sync + 'static {
         }
     }
 
+    async fn broadcast_to_lan(&self, shared: &SharedNetworkState) -> BroadcastToLan {
```
rj00a, line 346: This needs documentation.

crates/valence_network/src/lib.rs
```diff
@@ -549,3 +576,31 @@ pub struct PlayerSampleEntry {
...
     pub id: Uuid,
 }
+
+async fn do_broadcast_to_lan_loop(shared: SharedNetworkState) {
+    let port = shared.0.address.port();
+
+    // connect to server_addr on Udp to send a text packet
+    // if this ever fails, you have bigger problems to worry about, since we're
+    // asing the OS for a port
+    let socket = tokio::net::UdpSocket::bind("0.0.0.0:0").await.unwrap();
```
rj00a, line 586: Even in this scenario I think it's a bad idea to panic. Better to log an error message and return.

crates/valence_network/src/lib.rs
```diff
@@ -538,6 +553,18 @@ pub enum ServerListPing<'a> {
...
+/// The result of the Broadcast To Lan [callback].
+///
+/// [callback]: NetworkCallbacks::broadcast_to_lan
+#[derive(Clone, Default, Debug)]
+pub enum BroadcastToLan<'a> {
+    /// Disabled Broadcast To Lan.
+    #[default]
+    Disabled,
+    /// Send packet to broadcast to LAN every 1.5 seconds with specified MOTD.
+    Enabled(&'a str),
```
rj00a, line 565: I think using `&str` here isn't going to work if users wanted to create a message dynamically in the body of `broadcast_to_lan`, since the lifetime would end before returning. So it should be a `Cow<str>` instead.

crates/valence_network/src/lib.rs
```diff
@@ -549,3 +576,31 @@ pub struct PlayerSampleEntry {
     /// The player UUID.
     pub id: Uuid,
 }
+
+async fn do_broadcast_to_lan_loop(shared: SharedNetworkState) {
+    let port = shared.0.address.port();
+
+    // connect to server_addr on Udp to send a text packet
+    // if this ever fails, you have bigger problems to worry about, since we're
+    // asing the OS for a port
```
rj00a, line 585:
```suggestion
    // asking the OS for a port
```

crates/valence_network/src/lib.rs
rj00a, line 586: Consider importing `UdpSocket`, `Duration`, `tokio::time`,

rj00a (changes requested):

rj00a:
Would be nice to add lan broadcasting to the `server_list_ping` example.

rj00a:
Still doesn't show up in the list of LAN servers for me unfortunately. Haven't had time to investigate yet.

rj00a (approved):

Issue #380, opened 2023-06-19
Expand the `Packet` trait


rj00a:
## Describe the problem related to your feature request.

The `Packet` trait contains the packet's `id` and `name`, but it does not include the packet's `side` and `state`. This means bugs like sending a serverbound packet to clients, or sending a play packet during login cannot be caught.

## What solution would you like?

Expand the `Packet` trait to look like this:

```rust
pub trait Packet: std::fmt::Debug {
    /// The name of this packet for debugging purposes.
    const NAME: &'static str;
    /// The leading VarInt ID of this packet.
    const ID: i32;
    const SIDE: PacketSide;
    const STATE: PacketState;

    ...
}

#[derive(Copy, Clone, PartialEq, Eq)]
pub enum PacketSide {
    /// Server -> Client
    Clientbound,
    /// Client -> Server
    Serverbound,
}

#[derive(Copy, Clone, PartialEq, Eq)]
pub enum PacketState {
    Handshaking,
    Status,
    Login,
    Play,
}
```

Then the `Packet` derive macro will be used like this:

```rust
#[derive(Packet, Debug, Encode, Decode)]
#[packet(name = "ExamplePacketS2c", id = packet_id::EXAMPLE_PACKET_S2C, side = PacketSide::Clientbound, state = PacketState::Play)]
struct ExamplePacketS2c {
    foo: i32,
    bar: f32,
}
```

Specifying all those arguments to the `packet` helper attribute is really cumbersome, so I propose using defaults for the arguments when they are not supplied:
- `name`: The name of the type being defined (this already happens, but there's currently no way to override the inferred name).
- `id`: `valence_core::protocol::packet_id::PACKET_NAME` where `PACKET_NAME` is the shouty snake case name of the type being defined.
- `side`: Chooses either `valence_core::protocol::PacketSide::Clientbound` or `valence_core::protocol::PacketSide::Serverbound` depending on if the packet name ends in `C2s` or `S2c` (case insensitive).
- `state`: Always defaults to `valence_core::protocol::PacketState::Play` since that is the most common.

## What alternative(s) have you considered?

We could use the trait system to catch bad `side` and `state` values at compile time instead of runtime by making `SIDE` and `STATE` associated types instead of constants, with `PacketSide` and `PacketState` trait bounds. However, this a pain in the ass.

rj00a:
In that case there shouldn't be a default and an error should be raised if `side` is not supplied.

PR #391, opened 2023-06-21
Expand packet trait and derive macro

tachibanayui:
## Description
Implement changes described in #380


rj00a (commented):
The packets in `valence_network` should be updated with the correct `state` since they are now defaulting to the wrong value.

crates/valence_core_macros/src/packet.rs
```diff
@@ -55,6 +85,34 @@ fn parse_packet_helper_attr(attrs: &[Attribute]) -> Result<Option<PacketAttr>> {
...
+                    let Some(first) = path.segments.first() else {
+                        return Err(meta.error("state path should have length equals 2"));
+                    };
+
+                    if first.ident != "PacketState" {
+                        return Err(meta.error("state must starts with `PacketState`"));
+                    }
+
+                    res.state = Some(path);
+                    Ok(())
```
rj00a, line 115: Might be better to treat these arguments as expressions rather than paths or literals for more flexibility.

PR #394, opened 2023-06-21
Version lock nightly in CI

AviiNL:
## Description

Version lock nightly to yesterday for fmt, it seems to fail right now, not sure why, testing if this fixes the fmt job again..

Okay, this seems to fix todays issue of nightly not wanting to install on the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
         platform: [ubuntu-latest]
         style: [default]
         rust:
-          - nightly
+          - nightly-2023-06-20
         include:
           - style: default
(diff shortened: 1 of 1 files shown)


rj00a:
Just reran CI in #388 and everything seems to be working now, so I'll close this.

AviiNL:
> Just reran CI in #388 and everything seems to be working now, so I'll close this.

It was very on-and-off this morning, not sure if that's going to be something that might happen more often when rustup cant find the new daily build if it's still processing, so I did this to better be safe than sorry and avoid issues like this in the future.

rj00a:
Good point.

PR #406, opened 2023-07-06
Add server list legacy ping handler

### LINKED ISSUE #282: Handle legacy server list pings

pre-Netty server list pings are not currently handled.

https://github.com/valence-rs/valence/blob/cd7dd8cc4cd6fe7765f82f0e2c6bdcf3dac887a1/crates/valence/src/server/connect.rs#L102-L104

Modern clients can still send the ping in some circumstances so we should deal with it.

This might be the source of an issue I've seen where the client will be timed out waiting for a response.

## Describe the solution you would like.

Respond to the status ping in the expected way. The `AsyncCallbacks` trait might need adjustment to account for this.

## Additional context

https://wiki.vg/Server_List_Ping#1.6

PonasKovas:
## Description

Closes #282 

Any questions, insights, comments or criticism welcome.

### DIFF
--- a/crates/valence_core/src/text.rs
+++ b/crates/valence_core/src/text.rs
@@ -556,15 +556,67 @@ impl Text {
         }
     }
 
-    /// Attempts to convert the [`Text`] object to a plain string with the [legacy formatting (`§` and format codes)](https://wiki.vg/Chat#Old_system)
+    /// Converts the [`Text`] object to a plain string with the [legacy formatting (`§` and format codes)](https://wiki.vg/Chat#Old_system)
     ///
-    /// Fails if the [`Text`] object contains anything other than text, color
-    /// and styles, or if any of the colors are not on [the legacy color list](https://wiki.vg/Chat#Colors).
-    pub fn try_into_legacy(&self) -> Option<String> {
-        fn try_into_legacy_inner(this: &Text, mut result: String) -> Option<String> {
-            if let Some(color) = this.0.color {
-                result.push('§');
-                let code = match color {
+    /// Removes everything that can't be represented with a `§` and a modifier.
+    /// Any colors not on the [the legacy color list](https://wiki.vg/Chat#Colors) will be replaced with their closest equivalent.
+    pub fn to_legacy_lossy(&self) -> String {
+        // For keeping track of the currently active modifiers
+        #[derive(Default, Clone)]
+        struct Modifiers {
+            obfuscated: Option<bool>,
+            bold: Option<bool>,
+            strikethrough: Option<bool>,
+            underlined: Option<bool>,
+            italic: Option<bool>,
+            color: Option<char>,
+        }
+
+        impl Modifiers {
+            // Writes all active modifiers to a String as `§<mod>`
+            fn write(&self, output: &mut String) {
+                if let Some(color) = self.color {
+                    output.push('§');
+                    output.push(color);
+                }
+                if let Some(true) = self.obfuscated {
+                    output.push_str("§k");
+                }
+                if let Some(true) = self.bold {
+                    output.push_str("§l");
+                }
+                if let Some(true) = self.strikethrough {
+                    output.push_str("§m");
+                }
+                if let Some(true) = self.underlined {
+                    output.push_str("§n");
+                }
+                if let Some(true) = self.italic {
+                    output.push_str("§o");
+                }
+            }
+            // Merges 2 Modifiers. The result is what you would get if you applied them both
+            // sequentially.
+            fn add(&self, other: &Self) -> Self {
+                Self {
+                    obfuscated: other.obfuscated.or(self.obfuscated),
+                    bold: other.bold.or(self.bold),
+                    strikethrough: other.strikethrough.or(self.strikethrough),
+                    underlined: other.underlined.or(self.underlined),
+                    italic: other.italic.or(self.italic),
+                    color: other.color.or(self.color),
+                }
+            }
+        }
+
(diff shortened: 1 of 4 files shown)


crates/valence_core/src/text.rs
```diff
@@ -555,6 +555,65 @@ impl Text {
             TextContent::StorageNbt { nbt, .. } => nbt.is_empty(),
         }
     }
+
+    /// Attempts to convert the [`Text`] object to a plain string with the [legacy formatting (`§` and format codes)](https://wiki.vg/Chat#Old_system)
+    ///
+    /// Fails if the [`Text`] object contains anything other than text, color
+    /// and styles, or if any of the colors are not on [the legacy color list](https://wiki.vg/Chat#Colors).
+    pub fn try_into_legacy(&self) -> Option<String> {
```
PonasKovas, line 563: Oops, just now noticed that this is not correct, it doesn't remove formatting (if for example `obfuscated` was set to `Some(false)`). Could probably use a few tests too.

Will take care of this later when I have time. In the meantime waiting for any other comments

PonasKovas: I don't like this whole thing honestly. I quickly hacked it together until got something that works, but it definitely can be improved. If a color that's not of the 16 valid colors is encountered, instead of failing it could replace it with the closest valid color, for example

crates/valence_network/src/legacy_ping.rs
```diff
@@ -0,0 +1,305 @@
...
+pub struct ServerListLegacyPingResponse {
+    protocol: i32,
+    version: String,
+    online_players: i32,
+    max_players: i32,
+    description: String,
+}
+
+#[derive(PartialEq, Debug)]
+enum PingFormat {
```
rj00a, line 51: Could be nice to have this information passed to `server_list_legacy_ping`.

crates/valence_network/src/legacy_ping.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    Ok(true)
+}
+
+async fn read_payload(stream: &mut TcpStream) -> io::Result<ServerListLegacyPingPayload> {
+    // consume the first 29 useless bytes of this amazing protocol
+    stream.read_exact(&mut [0u8; 29]).await?;
+
+    let protocol = stream.read_u8().await? as i32;
+    let hostname_len = stream.read_u16().await? * 2;
```
rj00a, line 178: This multiplication could overflow and panic in debug mode. Also might want to think about bounding this to some reasonable limit.

crates/valence_network/src/legacy_ping.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    if num < 0 {
+        (num_abs.log10() + 2.0) as usize // because minus sign
+    } else {
+        (num_abs.log10() + 1.0) as usize
+    }
+}
+
+// Removes all `§` and their modifiers, if any
+fn remove_formatting(input: &str) -> String {
```
rj00a, line 290: Should there be a function somewhere to remove `\0` in the case where that's used as the separator?

crates/valence_network/src/lib.rs
```diff
@@ -347,6 +349,40 @@ pub trait NetworkCallbacks: Send + Sync + 'static {
...
+    /// This function is called from within a tokio runtime.
+    ///
+    /// # Default Implementation
+    ///
+    /// [`server_list_ping`][Self::server_list_ping] re-used.
+    async fn server_list_legacy_ping(
+        &self,
+        shared: &SharedNetworkState,
+        remote_addr: SocketAddr,
+        payload: Option<ServerListLegacyPingPayload>,
```
rj00a, line 364: Docs should explain what it means when this is `None`.

crates/valence_core/src/text.rs
rj00a, line 563: "into" usually implies the argument is consumed. Perhaps `try_to_legacy` or `to_legacy`?

rj00a: Considering this is a "best effort" sort of thing I'm OK with the hackyness of it. But some additional improvements and unit tests are welcome :)

rj00a (commented):
Looks good overall. Thanks for the detailed comments.

crates/valence_network/src/legacy_ping.rs
PonasKovas: `\0` are removed by the constructor of `ServerListLegacyPingResponse` (line [242](https://github.com/valence-rs/valence/pull/406/files/a25542f0b951d95b1c917afb37e557aeb94b72b7#diff-52d9e3c43798d1b5996c379d4a3bd84c436c6b876211a08e482271651b00cb84R242) and [262](https://github.com/valence-rs/valence/pull/406/files/a25542f0b951d95b1c917afb37e557aeb94b72b7#diff-52d9e3c43798d1b5996c379d4a3bd84c436c6b876211a08e482271651b00cb84R262))

PonasKovas:
Thanks for the review. Should be ready for merging now.

crates/valence_network/src/connect.rs
```diff
@@ -79,7 +79,15 @@ async fn handle_connection(
         error!("failed to set TCP_NODELAY: {e}");
     }
 
-    match try_handle_legacy_ping(&shared, &mut stream, remote_addr).await {
+    let timeout = Duration::from_secs(5);
+
+    match tokio::time::timeout(
```
rj00a, line 84: This made me realize there should be a single timeout where `handle_connection` is called instead of individual timeouts for every packet send/recv via `PacketIo`. that'll have to be fixed later.

PR #407, opened 2023-07-10
Building example code de-duplication

PonasKovas:
Just a small PR fixing something I came across while reading around.

# Objective

- Simplify the example

# Solution

- Code de-duplication


dyc3 (approved):
Looks good.

PR #416, opened 2023-07-16
Add '.' to valid usernames to support GeyserMC

irishgreencitrus:
# Objective

Allow GeyserMC players to connect to a Valence server if GeyserMC is installed on the proxy and the Valence server is behind it (there is no way to support GeyserMC directly on Valence and there probably never will be)
# Solution

- GeyserMC uses a '.' prefix to distinguish Bedrock players
- Allow '.' in the valid names
- Bedrock players can now connect

### DIFF
--- a/crates/valence_client/src/lib.rs
+++ b/crates/valence_client/src/lib.rs
@@ -478,8 +478,12 @@ impl fmt::Display for Username {
 /// Returns whether or not the given string is a valid Minecraft username.
 ///
 /// A valid username is 3 to 16 characters long with only ASCII alphanumeric
-/// characters. The username must match the regex `^[a-zA-Z0-9_]{3,16}$` to be
+/// characters. The username must match the regex `^[a-zA-Z0-9_\.]{3,16}$` to be
 /// considered valid.
+///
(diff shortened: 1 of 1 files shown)


rj00a:
What happens if someone with a 16 character username joins the server? I'm guessing GeyserMC would prepend the `.` character and push the username over the 16 character max, causing `is_valid_username` to return false.

I also see that the `LoginHelloC2SPacket` from vanilla MC has a 16 character limit on the `name` field. Does geyser do something to get around this?

rj00a:
Regardless, I think we should just delete the `is_valid_username` function here. The client doesn't have any problem handling usernames with unusual characters, and proper username validation will happen in online mode anyway.

irishgreencitrus:
> Regardless, I think we should just delete the `is_valid_username` function here. The client doesn't have any problem handling usernames with unusual characters, and proper username validation will happen in online mode by the auth server anyway.

I think that's a good shout. Valence also seems to have no problems with invalid characters in usernames, so this limitation seems artificial. It can also be re-enabled by the user in a bevy system if desired.

rj00a:
Closing in favor of #420

PR #418, opened 2023-07-18
Allow clippy::suspicious_else_formatting in generated valence_entity source file

PonasKovas:
# Objective

- Remove all of the useless clippy warnings from`valence_entity` build.rs generated file.

Screenshot_20230718_133009

# Solution

- `#[allow(clippy::suspicious_else_formatting)]`


rj00a:
Guessing this only happens when the `rustfmt` invocation from the build script fails to run.

PR #424, opened 2023-07-26
Visibility Layers

### LINKED ISSUE #342: Visibility Layers

rj00a:
# Objective

The objective of this PR is to solve the following problems with a holistic redesign of instances.

Closes #342, #362, but deviates greatly from the solution described in the issue.

- If an entity is within the view distance of a client, packets for that entity are sent to the client unconditionally. However, there is often a need to limit the visibility of entities to certain clients depending on the situation. Some examples:
  - You have a minigame with spectators and players. Spectators can see both players and spectators, but players can only see players and _not_ spectators.
  - An entity or set of entities is used as a makeshift GUI, only visible to the client using the GUI.
  - The server has many separate "arenas" that have identical chunk data, and you wish to share the chunk data across arenas to save memory.
  - You want to make a player truly invisible and potentially have some other entity take its place.

  It is possible to work around the problem by using invisibility effects or sending packets manually. But these are hacky solutions that are incomplete, inefficient, and inconvenient.
- Updating clients involves looking up every chunk in the client's view distance every tick for every client. This wastes CPU time and doesn't scale well with larger view distances[^1].
- Sometimes we want to broadcast packets to all clients matching some condition. Conditions could include...
  - If a chunk position is within the client's view.
  - If the client is within a certain radius of some position.
  - If the client is not some specific client (self exclusion).
  
  It isn't really possible to add all of these conditions in an efficient way using the current design.

# Solution

Split the existing `Instance` component into two new components: `ChunkLayer` and `EntityLayer`. Chunk layers contain all of the chunks in a world along with some global configuration, like the world's dimension. Entity layers contain Minecraft entities. A `LayerBundle` containing both is provided for convenience. Both `ChunkLayer` and `EntityLayer` implement the common `Layer` trait.

The key idea is this: Clients can only view one chunk layer[^2], but can view any number of entity layers. These are the `VisibleChunkLayer` and `VisibleEntityLayers` components respectively. The client will receive entity packets from only the layers it can see. Clients can add and remove entity layers at any time to spawn and despawn the entities in the layer.

Every layer contains a "message buffer" for broadcasting information to all viewers. Every message contains a "key" and a payload of bytes (its meaning depends on the message). Clients walk through the list of messages and use this to update themselves. 

There are a few things done to make this faster:
- Message processing is parallelized over all clients.
- The messages are sorted by their "key" and then deduplicated by merging payloads together. For instance, messages sending packet data on the condition that a certain chunk position is in view will be merged together if the chunk position is the same. Now there's only one memcpy instead of two.
- Messages are split into two categories: "global" and "local". Local messages are those that have some spatial condition to them. Local messages are put in a bounding volume hierarchy so that large swaths of messages do not need to be examined by clients.

# TODOs
- [x] More unit tests and fix bugs.
- [x] Add missing docs.
- [x] Reimplement `valence_anvil`.
- [x] Reimplement `weather` module.
- [x] Reimplement `valence_world_border`.

[^1]: At least the work is completely parallelized ¯\\\_(ツ)\_/¯.
[^2]: Viewing multiple chunk layers has some technical problems and I don't see it as a design worth pursuing.


dyc3:
I'm working on a capture the flag example to try out the API. I'll post it as a PR when it's done.

Bafbi:
I'v made a simple example for though who want : 

## Playground

<details>
<summary>code</summary>

```rust
use std::f64::consts::TAU;

use glam::{DQuat, EulerRot};
use valence::log::LogPlugin;
use valence::prelude::*;

#[allow(unused_imports)]
use crate::extras::*;

type SpherePartBundle = valence::entity::cow::CowEntityBundle;

const SPHERE_CENTER: DVec3 = DVec3::new(0.5, SPAWN_POS.y as f64 + 2.0, 0.5);
const SPHERE_AMOUNT: usize = 200;
const SPHERE_MIN_RADIUS: f64 = 6.0;
const SPHERE_MAX_RADIUS: f64 = 12.0;
const SPHERE_FREQ: f64 = 0.5;

const SPAWN_POS: BlockPos = BlockPos::new(0, 100, -16);

/// Marker component for entities that are part of the sphere.
#[derive(Component)]
struct SpherePart;

#[derive(Component)]
struct Main;

#[derive(Component)]
struct LayerSwitcherState(bool);

#[derive(Component)]
struct LayerSwitcherPos(BlockPos);

#[derive(Bundle)]
struct LayerSwitcherBundle {
    state: LayerSwitcherState,
    pos: LayerSwitcherPos,
    layer: EntityLayerId,
}

impl LayerSwitcherBundle {
    fn new(layer: EntityLayerId, pos: BlockPos) -> Self {
        Self {
            state: LayerSwitcherState(false),
            pos: LayerSwitcherPos(pos),
            layer,
        }
    }
}

const LAYER_SWITCHERS_POS: [BlockPos; 4] = [
    BlockPos::new(-1, SPAWN_Y, -1),
    BlockPos::new(1, SPAWN_Y, -1),
    BlockPos::new(-1, SPAWN_Y, 1),
    BlockPos::new(1, SPAWN_Y, 1),
];

const SPAWN_Y: i32 = 64;

        ..Default::default()
    })
    .add_systems(Startup, setup)
    .add_systems(EventLoopUpdate, (update_sphere, update_visible_layers))
    .run();
}

fn setup(mut commands: Commands, server: Res<Server>) {
    commands.spawn((EntityLayer::new(&server), Main));

    // Spawn 4 EntityLayers, one for each quadrant and get there IDs.
    let layer_ids = (0..4)
        .map(|_| commands.spawn(EntityLayer::new(&server)).id())
        .collect::<Vec<_>>();

    // Setup layers switchers.
    // create a bundle layer switcher with there positions.
    let layer_switchers = [
        LayerSwitcherBundle::new(EntityLayerId(layer_ids[0]), LAYER_SWITCHERS_POS[0]),
        LayerSwitcherBundle::new(EntityLayerId(layer_ids[1]), LAYER_SWITCHERS_POS[1]),
        LayerSwitcherBundle::new(EntityLayerId(layer_ids[2]), LAYER_SWITCHERS_POS[2]),
        LayerSwitcherBundle::new(EntityLayerId(layer_ids[3]), LAYER_SWITCHERS_POS[3]),
    ];

    // Spawn the layer switchers.
    commands.spawn_batch(layer_switchers);

    commands.spawn_batch([0; SPHERE_AMOUNT].iter().enumerate().map(move |(i, _)| {
        (
            SpherePartBundle {
                layer: EntityLayerId(layer_ids[i % 4]),
                ..Default::default()
            },
            SpherePart,
        )
    }));
}

fn init_clients(
    mut commands: Commands,
    server: Res<Server>,
    biomes: Res<BiomeRegistry>,
    mut clients: Query<
        (
            &mut EntityLayerId,
            &mut VisibleChunkLayer,
            &mut VisibleEntityLayers,
            &mut Position,
            &mut GameMode,
        ),
        Added<Client>,
    >,
    main_entity_layer: Query<Entity, (With<EntityLayer>, With<Main>)>,
) {
    for (
        mut layer_id,
        mut visible_chunk_layer,
        mut visible_entity_layers,
        mut pos,
        mut game_mode,
    ) in &mut clients
    {
        let entity_layer = main_entity_layer.single();

        let mut chunk_layer = ChunkLayer::new(ident!("overworld"), &dimensions, &biomes, &server);
        init_chunk_layer(&mut chunk_layer);
        let chunk_layer_id = commands.spawn(chunk_layer).id();

        layer_id.0 = entity_layer;
        visible_chunk_layer.0 = chunk_layer_id;
        visible_entity_layers.0.insert(entity_layer);
        *game_mode = GameMode::Creative;
    }
}

fn init_chunk_layer(chunk_layer: &mut ChunkLayer) {
    for z in -5..5 {
        for x in -5..5 {
            chunk_layer.insert_chunk([x, z], UnloadedChunk::new());
        }
    }

    for z in -25..25 {
        for x in -25..25 {
            chunk_layer.set_block([x, SPAWN_Y, z], BlockState::GRASS_BLOCK);
        }
    }

    // set_block the redstone lamps.
    for pos in LAYER_SWITCHERS_POS.iter() {
        chunk_layer.set_block(*pos, BlockState::REDSTONE_LAMP);
    }
}

// Add more systems here!

fn update_visible_layers(
    mut dig_events: EventReader<DiggingEvent>,
    mut clients: Query<(&mut VisibleEntityLayers, &VisibleChunkLayer)>,
    mut layer_switchers: Query<(&mut LayerSwitcherState, &EntityLayerId, &LayerSwitcherPos)>,
    mut chunk_layers: Query<&mut ChunkLayer>,
) {
    for dig_event in dig_events.iter() {
        for (mut state, layer, pos) in layer_switchers.iter_mut() {
            if pos.0 == dig_event.position {
                state.0 = !state.0;
                if let Ok((mut layers, chunk_layer_id)) = clients.get_mut(dig_event.client) {
                    if let Ok(mut chunk_layer) = chunk_layers.get_mut(chunk_layer_id.0) {
                        chunk_layer.set_block(
                            pos.0,
                            BlockState::REDSTONE_LAMP.set(
                                PropName::Lit,
                                if state.0 {
                                    PropValue::True
                                } else {
                                    PropValue::False
                                },
                            ),
                        );
                    }

                    if state.0 {
                        layers.0.insert(layer.0);
                    } else {
                        layers.0.remove(&layer.0);
                    }
                }
            }
        }
    }
}

fn update_sphere(
    settings: Res<CoreSettings>,
    server: Res<Server>,
    mut parts: Query<(&mut Position, &mut Look, &mut HeadYaw), With<SpherePart>>,
) {
    let time = server.current_tick() as f64 / settings.tick_rate.get() as f64;

    let rot_angles = DVec3::new(0.2, 0.4, 0.6) * SPHERE_FREQ * time * TAU % TAU;
    let rot = DQuat::from_euler(EulerRot::XYZ, rot_angles.x, rot_angles.y, rot_angles.z);

    let radius = lerp(
        SPHERE_MIN_RADIUS,
        SPHERE_MAX_RADIUS,
        ((time * SPHERE_FREQ * TAU).sin() + 1.0) / 2.0,
    );

    for ((mut pos, mut look, mut head_yaw), p) in
        parts.iter_mut().zip(fibonacci_spiral(SPHERE_AMOUNT))
    {
        debug_assert!(p.is_normalized());

        let dir = rot * p;

        pos.0 = SPHERE_CENTER + dir * radius;
        look.set_vec(dir.as_vec3());
        head_yaw.0 = look.yaw;
    }
}

/// Distributes N points on the surface of a unit sphere.
fn fibonacci_spiral(n: usize) -> impl Iterator<Item = DVec3> {
    let golden_ratio = (1.0 + 5_f64.sqrt()) / 2.0;

    (0..n).map(move |i| {
        // Map to unit square
        let x = i as f64 / golden_ratio % 1.0;
        let y = i as f64 / n as f64;

        // Map from unit square to unit sphere.
        let theta = x * TAU;
        let phi = (1.0 - 2.0 * y).acos();
        DVec3::new(theta.cos() * phi.sin(), theta.sin() * phi.sin(), phi.cos())
    })
}

fn lerp(a: f64, b: f64, t: f64) -> f64 {
    a * (1.0 - t) + b * t
}
```

rj00a:
@Bafbi I'm unable to reproduce the desync you're seeing using your example. I've repeatedly hit the lamps in different ways and everything seems OK so far. What steps did you do to arrive at the screenshot?

Bafbi:
@rj00a ngl, I have no idea, I couldn't reproduce it either but I'm sure it was with this code.

Bafbi:
Something I would also really like to have is the possibility to put directly the `Entity` of a client  in the `EntityLayerId` component, that would mean that only this client can see this entity. This would be much more practical than to having to manage a layer for each client.

dyc3:
I put what I have so far for the `ctf` exmaple in #426. 

Currently, there is a bug where new players on a team are unable to see existing players. I don't think I'm doing anything too weird in the team joining system, so I'm inclined to think it could be a bug in this PR.

Also, it would be nice if we could make cloning entities to other layers a bit easier. Something where it clones the entity's appearance as it currently is, and then only updates position, rotation, etc automatically.

rj00a:
> Something I would also really like to have is the possibility to put directly the Entity of a client in the EntityLayerId component

If you insert an `EntityLayer` component on the client itself, you will get exactly the behavior you're looking for.

rj00a:
> Also, it would be nice if we could make cloning entities to other layers a bit easier. Something where it clones the entity's appearance as it currently is, and then only updates position, rotation, etc automatically.

The flecs ECS library has a feature that lets you inherit components from other entities. It uses a special relationship the query engine understands, called `IsA`. Components from the child entity take precedence over the parent entity. https://ajmmertens.medium.com/a-roadmap-to-entity-relationships-5b1d11ebb4eb#cc02

The bevy devs have plans to add entity relationships to bevy_ecs so we might be getting this feature in the near future. Once it's in, I don't think there's anything special we need to do on our end to make this work.

dyc3:
The visibility bug that's happening in the ctf example is happening because the packets to create the new entity with the same uuid arrive before the destroy packets for the old ones.

image

rj00a:
> The visibility bug that's happening in the ctf example is happening because the packets to create the new entity with the same uuid arrive before the destroy packets for the old ones.

Should be fixed now. This also fixed the issue where cows were disappearing in the cow sphere example 😃

dyc3:
Just rebased the ctf example, I can confirm it's fixed.

dyc3 (commented):
Overall, I think the design is good, and that the API surface doesn't seem to have any major gaps.

Once the TODOs are fixed I can review this with a finer comb.

rj00a:
Alright, this should be ready to go.

@dyc3 If you'd like to merge the ctf example we can do that now.

rj00a:
Hmm mysterious CI failures.

crates/valence_client/src/lib.rs
```diff
@@ -774,234 +667,308 @@ fn update_chunk_load_dist(
...
+                dist_squared as u32 <= radius_squared
+            }
+
+            // Chunk layer messages
+            if let Ok(chunk_layer) = chunk_layers.get(old_visible_chunk_layer.get()) {
+                let messages = chunk_layer.messages();
+                let bytes = messages.bytes();
+
+                // Global messages
+                for (msg, range) in messages.iter_global() {
```
dyc3, line 720: Would you mind explaining the difference between global and local messages?

crates/valence_layer/README.md
```diff
@@ -0,0 +1 @@
+# TODO
```
dyc3, line 1: fix todo

crates/valence_layer/src/message.rs
```diff
@@ -0,0 +1,276 @@
...
+impl<G, L> Messages<G, L>
+where
+    G: Copy + Ord,
+    L: Copy + Ord + GetChunkPos,
+{
+    pub(crate) fn new() -> Self {
+        Self::default()
+    }
+
+    pub(crate) fn send_global<E>(
```
dyc3, line 27: This and `send_local` would be a good place to explain the difference between global and local messages

crates/valence_world_border/src/lib.rs
```diff
@@ -1,59 +1,4 @@
...
-//! [`WorldBorderDiameter::get()`]. Note: If you want to modify the
-//! diameter size, do not modify the value directly! Use
-//! [`SetWorldBorderSizeEvent`] instead.
-//!
-//! ## Access other world border properties.
-//! Access to the rest of the world border properties is fairly straightforward
-//! by querying their respective component. [`WorldBorderBundle`] contains
-//! references for all properties of the world border and their respective
-//! component
+#![doc = include_str!("../README.md")]
```
dyc3, line 1: These docs need to come back in the README

crates/valence_world_border/src/lib.rs
```diff
@@ -94,316 +34,215 @@ pub const DEFAULT_DIAMETER: f64 = (DEFAULT_PORTAL_LIMIT * 2) as f64;
...
-            &WorldBorderWarnTime,
-            &WorldBorderWarnBlocks,
-            &WorldBorderDiameter,
-            &WorldBorderPortalTpBoundary,
-            Option<&MovingWorldBorder>,
-        ),
-        With<Instance>,
-    >,
+fn init_world_border_for_new_clients(
+    mut clients: Query<(&mut Client, &VisibleChunkLayer), Changed<VisibleChunkLayer>>,
```
dyc3, line 129: does this make world borders act per chunk layer?

src/tests/weather.rs
dyc3, line 1: It would be nice to not lose these tests. Either file an issue for it, or fix it here.

src/tests/world_border.rs
dyc3, line 1: It would be nice to not lose these tests. Either file an issue for it, or fix it here.

dyc3 (changes requested):
Looks good.

dyc3:
The CI failures might be related to `Testing many_players`. Maybe it's getting OOM killed?

crates/valence_world_border/src/lib.rs
rj00a: Yep. I'll explain that in the docs. I didn't think it made much sense to act on entity layers since clients can view >1 entity layers.

dyc3 (approved):

PR #432, opened 2023-08-02
Configure cargo to use `-Ctarget-cpu=native`

rj00a:
# Objective

Tell rustc to use native CPU features when building the project. Especially useful for the benchmarks.

# Solution

Add a `.cargo/config.toml` with the following contents:

```toml
[build]
rustflags = ["-Ctarget-cpu=native"]
```

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+rustflags = ["-Ctarget-cpu=native"]


dyc3:
This makes it so that built binaries can possibly be not portable.

rj00a:
True, but I didn't think it would matter since the config only applies to those building valence from the repo AFAIK. But I can still see that coming back to bite us.

Issue #438, opened 2023-08-04
Handle Player Abilities


Bafbi:
## Describe the problem related to your feature request.

**Player Abilities** refer to this packet [PlayerAbilitiesS2c](https://wiki.vg/Protocol#Player_Abilities).

Player abilities are client side variable, that indicate some abilities of the client (can_fly, fly_speed, is_invulnerable, ...).
It can be use to activate fly in survival...

## What solution would you like?

Adding the `PlayerAbilitiesFlags` struct as a component on the ClientBundle and send packet on change.

## What alternative(s) have you considered?

Currently we can just send directly the packet, but it have a limitation which is that, changing the gamemode also change **Player Abilities** and so we can't do both at the same time without complexity.
__demo__
```rust
*creator_state = Gamemode::Adventure
client.write_packet(&PlayerAbilitiesS2c {
    flags: PlayerAbilitiesFlags::new()
        .with_allow_flying(true)
        .with_flying(true),
    flying_speed: 0.05,
    fov_modifier: 0.1,
});
```
Doing that result in not be able to fly because the `PlayerAbilitiesS2c` is send before `GameStateChangeS2c` that change the gamemode which update the abilities.

![image](https://github.com/valence-rs/valence/assets/26836027/a701e66f-c012-475c-8ffa-80c4e8799cea)

PR #439, opened 2023-08-04
Add abilities components on the player

Bafbi:
# Objective

see #438 

# Solution

The client is updated when a change is detected on one of the player abilities component.

### DIFF
--- a/crates/valence_client/src/abilities.rs
+++ b/crates/valence_client/src/abilities.rs
@@ -1,0 +1,62 @@
+pub use valence_packet::packets::play::player_abilities_s2c::PlayerAbilitiesFlags;
+use valence_packet::packets::play::PlayerAbilitiesS2c;
+
+use super::*;
+
+/// [`Component`] that stores the player's flying speed ability.
+///
+/// [`Default`] value: `0.05`.
+#[derive(Component)]
+pub struct FlyingSpeed(pub f32);
+
+impl Default for FlyingSpeed {
+    fn default() -> Self {
+        Self(0.05)
+    }
+}
+
+/// [`Component`] that stores the player's field of view modifier ability.
+/// The lower the value, the higher the field of view.
+///
+/// [`Default`] value: `0.1`.
+#[derive(Component)]
+pub struct FovModifier(pub f32);
+
+impl Default for FovModifier {
+    fn default() -> Self {
+        Self(0.1)
+    }
+}
+
+pub(super) fn build(app: &mut App) {
+    app.add_systems(
+        PostUpdate,
+        update_player_abilities
+            .in_set(UpdateClientsSet)
+            .after(update_game_mode),
+    );
+}
+
+fn update_player_abilities(
+    mut clients_query: Query<
+        (
+            &mut Client,
+            &PlayerAbilitiesFlags,
+            &FlyingSpeed,
+            &FovModifier,
+        ),
+        Or<(
+            Changed<PlayerAbilitiesFlags>,
+            Changed<FlyingSpeed>,
+            Changed<FovModifier>,
+        )>,
+    >,
+) {
+    for (mut client, flags, flying_speed, fov_modifier) in clients_query.iter_mut() {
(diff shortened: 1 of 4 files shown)


rj00a (changes requested):
If changing game modes changes abilities on the client, then we should be updating the ability values here to keep things in sync.

There's also a [serverbound player abilities flag](https://wiki.vg/Protocol#Player_Abilities) that we should handle by updating the `flying` flag.

crates/valence_client/src/abilities.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+fn update_server_player_abilities(
+    mut packet_events: EventReader<PacketEvent>,
+    mut client_query: Query<&mut PlayerAbilitiesFlags>,
+) {
+    for packets in packet_events.iter() {
+        if let Some(pkt) = packets.decode::<UpdatePlayerAbilitiesC2s>() {
+            if let Ok(mut flags) = client_query.get_mut(packets.client) {
+                flags.set_flying(UpdatePlayerAbilitiesC2s::StartFlying.eq(&pkt));
```
rj00a, line 101:
```suggestion
                flags.set_flying(UpdatePlayerAbilitiesC2s::StartFlying == pkt);
```

crates/valence_client/src/abilities.rs
```diff
@@ -0,0 +1,106 @@
...
+                flags.bypass_change_detection();
```
rj00a, line 102: You have to use the reference returned by this function to bypass change detection. The previous line is triggering change detection as usual.

I would be careful using this function in general because it can lead to some surprising behavior for end users. However, I think it's fine in this case because the change detection isn't granular enough to be useful.

examples/cow_sphere.rs
```diff
@@ -142,3 +149,16 @@ fn fibonacci_spiral(n: usize) -> impl Iterator<Item = DVec3> {
...
+    mut clients: Query<(&mut Client, &PlayerAbilitiesFlags), Changed<PlayerAbilitiesFlags>>,
+) {
+    for (mut client, abilities_flags) in clients.iter_mut() {
+        if abilities_flags.flying() {
+            client.send_action_bar_message("You are flying!".into_text().color(Color::GREEN));
+        } else {
+            client.send_action_bar_message("You are not flying!".into_text().color(Color::RED));
+        }
+    }
+}
```
rj00a, line 164: At this point I think it would be best to create a dedicated example for abilities.

crates/valence_client/src/abilities.rs
rj00a, line 101: This should check that the client is allowed to start flying before modifying to prevent cheating.

rj00a (changes requested):

crates/valence_client/src/abilities.rs
Bafbi: I don't think cheat detection should be handle here, but in a separate plugin that handle all of them, now to do it, I firstly want to make a better event system.
```
(valence_event) --------------> (valence_check) -------------> (user_code) --------------------> (valence_client)
[PlayerStartFlyingEvent)      [event.set_cancel(...)]           [if event.is_cancel {...}]          [flags.set_flying(...)]
```

examples/cow_sphere.rs
Bafbi: I don't really know because in reality, `cow_sphere` wasn't showing much and I think `Abilities` are not this important, but they have demonstration in some example so that if someone ask how to use it we can just redirect them.

crates/valence_client/src/abilities.rs
Bafbi: First the change detection is usefull cause I use it in the  `cow_sphere` but why is this working if change detection was not bypass ?

examples/cow_sphere.rs
rj00a: That's fine. In the future I would like the examples to be more complex so we don't have to keep making examples for every tiny feature we add.

rj00a (approved):

PR #445, opened 2023-08-05
Add CI documentation publisher

### LINKED ISSUE #423: Publish nightly docs
It doesn't look like there's docs for the current version on `main`, but I think they'd be quite helpful for anyone who has valence as a Git dependency from this repository. Things like `DefaultPlugins` don't show up on docs.rs (I'm assuming because they weren't present in the 0.1 release), but it'd be nice to figure out what those do without digging into the actual source code.

hwittenborn:
# Objective
Fixes #423.

# Solution
Docs are automatically published to GitHub pages on every commit to `main`.


hwittenborn:
Once this gets merged, nightly docs should automatically be available at https://valence.github.io/valence (GitHub Pages will have to be enabled in the repo settings via GitHub Actions though). Hosting on a custom domain (i.e. `docs.valence.rs`) should work without any changes in the event it's needed too though.

hwittenborn:
Also noting that I checked that this works in my fork, you can view the docs at https://github.hwittenborn.io/valence.

I also wasn't sure which crates should be documented, but adding `-p {pkg}` arguments to `cargo doc` will fix the issue just fine.

hwittenborn:
I noticed all the [`valence-*`](https://hwittenborn.github.io/valence/valence/#reexports) crates are re-exported at the root, but the current implementation doesn't allow viewing the docs for them since those other crates aren't having their docs exposed.

Are the `valence-*` crates meant to be able to accessed directly?

dyc3:
The subcrates and reexports need to be accessible via the root valence docs. The `valence` crate is intended to appear as one crate from the outside.

hwittenborn:
Sounds good, that's not a problem at all.

I'd like to note that there isn't really a concept of the "root" of the docs with rustdoc, it's just a collection of all the crates that you can navigate between. Since there wasn't a root I decided I'd make the `valence` crate the landing page, since it'd be the main crate people interact with.

hwittenborn:
I'm marking this as a draft until I can review the functionality in my repo's deployed site.

hwittenborn:
Should [the README](https://github.com/valence-rs/valence#adding-valence-as-a-dependency) be updated to the new docs location as well? Considering it talks about using valence as a Git dependency there, I think it'd be good to also link to the git-based documentation.

hwittenborn:
I'm still working out why some crates aren't being linked in the root of the `valence` crate's re-exports, after that's done this should be pretty much ready to go.

dyc3:
Updating the readme would be good. 

Don't worry about urgency, I'll review it when you mark it as ready to review.

hwittenborn:
I'm figuring out some documentation link issues before I get this finished up (see https://github.com/rust-lang/rust/issues/114675), and then this should be almost ready to go.

rj00a:
I wouldn't mind merging what we have here and fixing the doc issue later, since it looks like it's going to be a while before that's resolved.

PR #446, opened 2023-08-06
Alternate solution to command api.

JackCrumpLeys:
# Objective

I want to add commands ergonomically, unfortunately @Jenya705 's implementation did not look to promising for my purposes so I am working on an alternative command api.

# Solution

lib-users should be able to make commands quickly and easily.

Some objectives in the order I plan to implement them:

- [x] Sync commands to players based on command visibility scope.
- [x] Derive Macro
- [x] Execute and handle commands.
- [x] Parse commands with a lot of lib-user friendly options.
- [ ] ~~Handle server-side suggestions.~~ (this has been defered to the future)


crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+    fn arg_from_string(string: String) -> Result<Self::Result, CommandArgParseError> {
+        Ok(if string == "@a" {
+            PlayerSelector::All
+        } else if string == "@s" {
+            PlayerSelector::SelfPlayer
+        } else if string == "@p" {
+            PlayerSelector::Nearest
+        } else {
+            PlayerSelector::Single(string)
+        })
```
rj00a, line 197: nit: would prefer a `match` on the string.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+
+impl<T> CommandArg for PossiblyRelative<T>
+where
+    T: CommandArg,
+{
+    type Result = PossiblyRelative<T::Result>;
+
+    fn arg_from_string(string: String) -> Result<Self::Result, CommandArgParseError> {
+        if string.starts_with('~') {
+            let possibly_value = string.trim_start_matches('~').to_string();
```
rj00a, line 258: Looks like you should use `strip_prefix` here.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+    fn display() -> Parser {
+        Parser::Vec3
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Default)]
+pub struct Vec2 {
+    x: PossiblyRelative<f32>,
+    y: PossiblyRelative<f32>,
+}
```
rj00a, line 367: Are the vec types supposed to use `f32` instead of `f64`?

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+    fn display() -> Parser {
+        Parser::Entity {
+            only_players: true,
+            single: true,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PossiblyRelative<T> {
```
rj00a, line 233: nit: Would prefer the name `AbsoluteOrRelative` (first pick) or `MaybeRelative`.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+    fn display() -> Parser {
+        Parser::Color
+    }
+}
+
+// TODO: json chat component proper
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct JsonChatComponent;
+
+impl CommandArg for JsonChatComponent {
```
rj00a, line 500: Looks like you could reuse the `valence_server::Text` type.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+        Parser::Message
+    }
+}
+
+// TODO: nbt proper
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct Nbt;
+
+impl CommandArg for Nbt {
```
rj00a, line 532: Could use the `valence_server::nbt::Compound` type. We have a (somewhat incomplete) SNBT encoders/decoders there.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+            allow_multiple: false,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct Swizzle {
+    x: bool,
+    y: bool,
+    z: bool,
```
rj00a, line 689: should these be public?

crates/valence_command/src/arg_parser.rs
rj00a, line 1: This file is getting quite large. I suggest placing each command arg in a private module file and re-exporting them all at the parent. Also consider renaming `arg_parser.rs` to `arg.rs`.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+    Aqua,
+    Red,
+    LightPurple,
+    Yellow,
+    #[default]
+    White,
+    Reset,
+}
+
+impl CommandArg for ChatColor {
```
rj00a, line 460: We have the `NamedColor` type already in `valence_text`.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+    }
+
+    fn display() -> Parser {
+        Parser::BlockPredicate
+    }
+}
+
+// TODO: item stack proper
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct ItemStack;
```
rj00a, line 424: We have an existing `ItemStack` type.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+
+    fn display() -> Parser {
+        Parser::ItemSlot
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct ResourceLocation;
+
+impl CommandArg for ResourceLocation {
```
rj00a, line 757: `valence_ident::Ident` for this.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum Dimension {
+    #[default]
+    Overworld,
+    Nether,
+    End,
+}
```
rj00a, line 878: Are these three dimensions hard coded, or does the client read from the dimension registry? If it's the ladder then it should probably be an arbitrary string.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum GameMode {
+    #[default]
+    Survival,
+    Creative,
+    Adventure,
+    Spectator,
+}
+
+impl CommandArg for GameMode {
```
rj00a, line 911: We have a `GameMode` type already.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+
+    fn display() -> Parser {
+        Parser::Time
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct Uuid;
+
+impl CommandArg for Uuid {
```
rj00a, line 960: `uuid::Uuid`.

crates/valence_command/src/command_graph.rs
rj00a, line 1: Could rename this to `graph.rs` since "command" should be assumed.

crates/valence_command/src/command_scopes.rs
rj00a, line 1: Could also rename this to `scopes.rs`.

crates/valence_command/src/manager.rs
```diff
@@ -0,0 +1,87 @@
...
+use bevy_ecs::query::Changed;
+use valence_server::client::{Client, SpawnClientsSet};
+use valence_server::protocol::packets::play::CommandTreeS2c;
+use valence_server::protocol::WritePacket;
+
+use crate::command_graph::CommandGraph;
+use crate::command_scopes::CommandScopes;
+use crate::{CommandRegistry, CommandScopeRegistry};
+
+pub struct CommandManagerPlugin;
```
rj00a, line 13: Is the main plugin for the whole lib? I would prefer the name `CommandPlugin` for consistency.

rj00a (commented):
Mostly skimmed through it for now but it's looking good so far. Let me know when you're ready for a full review.

crates/valence_command/src/arg_parser.rs
```diff
@@ -0,0 +1,970 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ArgLen {
+    Infinite, // can only be the last argument
+    Exact(u32),
+    Within(char), // Example "man this is cool" would be 4 args without this distinction
+    WithinExplicit(char, char), // [man this is cooler] would be 4 arg without this distinction
+}
+
+pub trait CommandArg: Default {
+    type Result: Default;
```
rj00a, line 19: This is a bit suspicious to me. What's the reason we're not just assuming `Result = Self` and using newtype wrappers where appropriate? I imagine that would be important for the derive macro.

JackCrumpLeys: Many of the arg parses simply use strings. For example their are 3 variants of strings: single word, quoted word and greedy string.

crates/valence_command/src/arg_parser.rs
JackCrumpLeys: That's the difference between a vec and a dvec

crates/valence_command/src/command_scopes.rs
```diff
@@ -0,0 +1,324 @@
...
+    ///
+    /// registry.link("valence:admin", "valence:command");
+    ///
+    /// assert!(registry.grants("valence:admin", "valence:command"));
+    /// ```
+    pub fn link(&mut self, scope: &str, other: &str) {
+        let scope_idx = *self
+            .string_to_node
+            .get(scope)
+            .expect(format!("scope {} does not exist", scope).as_str());
```
rj00a, line 311:
```suggestion
            .unwrap_or_else(|| panic!("scope \"{scope}\" does not exist"));
```

crates/valence_command/src/command_scopes.rs
```diff
@@ -0,0 +1,324 @@
...
+    ///
+    /// registry.add_scope("valence:command");
+    /// registry.add_scope("valence:command:tp");
+    /// registry.add_scope("valence:admin");
+    ///
+    /// registry.link("valence:admin", "valence:command");
+    ///
+    /// assert!(registry.grants("valence:admin", "valence:command"));
+    /// ```
+    pub fn link(&mut self, scope: &str, other: &str) {
```
rj00a, line 307: Should document that this can panic.

crates/valence_command/src/command_scopes.rs
```diff
@@ -0,0 +1,324 @@
...
+            Dot::with_config(&self.graph, &[dot::Config::EdgeNoLabel])
+        )?;
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub struct Scope {
+    pub name: String,
+}
```
rj00a, line 71:
```suggestion
pub struct Scope(pub String);
```

JackCrumpLeys: Ditto, I prefer to keep it expanded as it keeps code more readable then using .0 all the time.

examples/command.rs
```diff
@@ -0,0 +1,308 @@
...
+                )
+            });
+    }
+}
+
+pub fn main() {
+    App::new()
+        .add_plugins((
+            DefaultPlugins,
+            CommandHandler::from_command(TeleportCommand),
```
rj00a, line 106: Don't think each command should be a separate plugin. This would clutter the list of loaded plugins pretty quickly.

What I'd like to see is an `add_command` method on `App` similar to `add_event`. You could do that via an `AddCommand` trait impl'd on `App` and maybe `World` too.

crates/valence_command/src/handler.rs
rj00a, line 1: Don't use `println`. If you need to log something, use the macros in `tracing`.

crates/valence_command/src/handler.rs
```diff
@@ -0,0 +1,329 @@
...
+    fn build(&self, app: &mut App) {
+        println!("Registering command: {}", Self::command_name());
+
+        app.add_event::<CommandTypingEvent<T>>()
+            .add_event::<CommandExecutionEvent<T>>()
+            .insert_resource(CommandResource::new(self.command.clone()))
+            .add_systems(Update, command_event_system::<T>)
+            .add_systems(PostStartup, command_startup_system::<T>);
+
+        println!("Registered command: {}", Self::command_name());
```
rj00a, line 80: Bevy already logs when plugins are loaded. (if this is just a temp thing that's fine)

examples/command.rs
```diff
@@ -0,0 +1,308 @@
...
+
+pub enum TeleportResult {
+    ExecutorToLocation(arg_parser::Vec3),
+    ExecutorToTarget(String),
+    TargetToTarget((String, String)),
+    TargetToLocation((String, arg_parser::Vec3)),
+}
+
+#[derive(Resource, Clone)]
+struct TeleportCommand;
```
rj00a, line 21: I don't think this struct needs to exist. Unless I'm missing something, all the methods get run once, so it seems that it would be better to make a function like

```rust
impl App {
    pub fn add_command<T>(&mut self, name: impl Into<String>, build: impl FnOnce(&mut CommandGraphBuilder<T>)) -> &mut Self;
}
```

Or to make this friendly for the derive macro, rename the above to `add_command_raw` and add

```rust
fn add_command<C: Command>(&mut self) -> &mut Self {
    self.add_command_raw(C::name(), C::assemble_graph)
}
```

`Command` would be implemented on `TeleportResult` in this example.

examples/command.rs
JackCrumpLeys: What would the ad command function do aside from adding systems and/or plugins.

crates/valence_command/src/handler.rs
JackCrumpLeys: Be aware that this is an unclean draft branch. I just have println for my own debugging.

examples/command.rs
JackCrumpLeys: I'd like the command resource to be capable of storing data for various reasons. I also want to be able to reassemble the graph at runtime by modifying the command resource. Remember I'm not done yet so I have not started to work on the final implementation.

rj00a: Right, I'm just trying the avoid the situation mentioned on Discord where the derive macro has to generate a secondary type that the user needs to know about.

crates/valence_command/src/command_scopes.rs
```diff
@@ -0,0 +1,324 @@
...
+        &self.name
+    }
+}
+
+/// Command scope Component for players. this is a list of scopes that a player
+/// has. if a player has a scope, they can use any command that requires
+/// that scope.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Component, Default)]
+pub struct CommandScopes {
+    pub scopes: Vec<Scope>,
```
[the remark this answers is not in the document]
JackCrumpLeys: Why?

typos.toml
```diff
@@ -2,4 +2,4 @@
 extend-exclude = ["*.svg", "*.json"]
 
 [default]
-extend-ignore-re = ['\d+ths', 'CC BY-NC-ND']
+extend-ignore-re = ['\d+ths', 'CC BY-NC-ND', "trUe", "fAlse"]
```
dyc3, line 5: Is there a way we can ignore these per line instead of the entire project?

examples/command.rs
```diff
@@ -0,0 +1,433 @@
...
+use valence_command::{parsers, Command, CommandApp, CommandScopeRegistry, ModifierValue};
+use valence_command_derive::Command;
+use valence_server::op_level::OpLevel;
+
+const SPAWN_Y: i32 = 64;
+
+#[derive(Command, Debug, Clone)]
+#[paths("teleport", "tp")]
+#[scopes("valence:command:teleport")]
+enum Teleport {
```
dyc3, line 23: In the PR description, would you mind elaborating a little on the derive macro and how it works? Specifically, what's `paths` and `scopes`? I might have forgotten some context, and it would probably be useful to add to the book anyway.

JackCrumpLeys: no problem

typos.toml
JackCrumpLeys: I have no clue. I will do some research

JackCrumpLeys: no I don't think I can. take a look at the [config ref](https://github.com/crate-ci/typos/blob/master/docs/reference.md):

AviiNL: `extend-exclude` to exclude the file instead of the pattern?

dyc3: Ah. Not ideal, but it's not worth the effort to block on this.

JackCrumpLeys: > `extend-exclude` to exclude the file instead of the pattern?

blocking a whole file is not ideal I'd think.

crates/valence_protocol/src/packets/play/command_tree_s2c.rs
```diff
@@ -86,10 +86,10 @@ pub enum Parser<'a> {
     Dimension,
     GameMode,
     Time,
-    ResourceOrTag { registry: Ident<Cow<'a, str>> },
```
adryzz, line 89: is there a specific need for the switch from references and `Cow`s to owned strings?

adryzz (commented):
looks very good, especially the documentation for the graph structure

crates/valence_protocol/src/packets/play/command_tree_s2c.rs
JackCrumpLeys: The rest of my code uses `String`s and they are easier to work with. Basically &str good for packets not for centrally used data structures.

crates/valence_command/src/parsers/colour.rs
dyc3, line 1: nit: would prefer `color.rs` instead of `colour.rs` to remain consistent

crates/valence_command_derive/README.md
```diff
@@ -0,0 +1,104 @@
...
+        a: Option<i32>,
+        b: Option<QuotableString>,
+        c: Option<Vec2Parser>,
+        d: Option<Vec3Parser>,
+        e: Option<GreedyString>,
+    },
+}
+```
+
+## attributes
```
dyc3, line 74: nit
```suggestion
## Attributes
```

crates/valence_command_derive/README.md
```diff
@@ -0,0 +1,104 @@
+Simplify the creation of Valence commands with a derive macro.
+
+## Usage
+
+```rust
+#[derive(Command, Debug, Clone)]
+#[paths("teleport", "tp")]
+#[scopes("valence:command:teleport")]
+enum Teleport {
```
dyc3, line 9: I still think that `paths` is not a great name for this. What if we renamed it to `aliases`, `command` or `literal`?

crates/valence_command_derive/README.md
```diff
@@ -0,0 +1,104 @@
...
+}
+```
+
+## attributes
+
+### `#[paths(...)]` or `#[paths = "..."]`
+
+The `#[paths(...)]` or `#[paths = "..."]` attribute is used to specify the different paths that can be used to invoke
+the command. The paths are specified as string literals, where any arguments are enclosed in curly braces `{}`.
+The arguments are then mapped to fields in the command enum variant.
```
dyc3, line 80: If we are going to keep the name paths though, this needs to do a better job at explaining that Minecraft commands are parsed according to a graph structure. This is could be the first thing that someone looks at if they want to implement commands, and the context would be helpful for understanding the API.

Perhaps a mermaid diagram for the teleport command would be a useful visual.

crates/valence_command/src/modifier_value.rs
```diff
@@ -0,0 +1,219 @@
...
+    String(String),
+
+    Unit,
+    Option(Option<Box<ModifierValue>>),
+    Seq(Vec<ModifierValue>),
+    Map(BTreeMap<ModifierValue, ModifierValue>),
+}
+
+#[allow(clippy::unit_hash)]
+impl Hash for ModifierValue {
```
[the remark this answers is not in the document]
JackCrumpLeys: float is a bit picky about that stuff. (the reason we use `OrderedFloat`)

crates/valence_command/src/parsers/colour.rs
JackCrumpLeys: haha you filthy americans :rofl:.

will do!

crates/valence_command_derive/README.md
JackCrumpLeys: aliases doesn't reflect the fact it is a path, command is a bit to similar to the macro name for me but it could work and literal is a bit too out there.

dyc3: Yeah, I suppose it would be better to just elaborate more in the docs.

crates/valence_command/src/parsers/color.rs
```diff
@@ -0,0 +1,55 @@
...
+use valence_text::Color;
+
+use super::Parser;
+use crate::parsers::{CommandArg, CommandArgParseError, ParseInput};
+
+impl CommandArg for Color {
+    fn parse_arg(input: &mut ParseInput) -> Result<Self, CommandArgParseError> {
+        input.skip_whitespace();
+        if input.match_next("black") {
+            Ok(Self::Named(NamedColor::Black))
```
rj00a, line 11:
```suggestion
            Ok(Self::BLACK)
```

and so on

crates/valence_command/src/parsers/column_pos.rs
```diff
@@ -0,0 +1,64 @@
+use super::Parser;
+use crate::parsers::{AbsoluteOrRelative, CommandArg, CommandArgParseError, ParseInput};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct ColumnPos {
+    x: AbsoluteOrRelative<i32>,
+    y: AbsoluteOrRelative<i32>,
+    z: AbsoluteOrRelative<i32>,
```
rj00a, line 8: Make these fields public?

crates/valence_command/src/parsers/mod.rs
rj00a, line 1: We're not using `mod.rs` style modules in valence, so the other style should be used for consistency.

crates/valence_command/src/parsers/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+    pub fn new(input: impl Into<String>) -> Self {
+        Self {
+            input: input.into(),
+            cursor: 0,
+        }
+    }
+
+    pub fn set_cursor(&mut self, cursor: usize) {
+        self.cursor = cursor;
+    }
```
rj00a, line 58: cursor is `pub`, no need for method.

examples/command.rs
```diff
@@ -0,0 +1,433 @@
...
+use valence_command::scopes::CommandScopes;
+use valence_command::{parsers, Command, CommandApp, CommandScopeRegistry, ModifierValue};
+use valence_command_derive::Command;
+use valence_server::op_level::OpLevel;
+
+const SPAWN_Y: i32 = 64;
+
+#[derive(Command, Debug, Clone)]
+#[paths("teleport", "tp")]
+#[scopes("valence:command:teleport")]
```
rj00a, line 22: I think we should use `.` as the scope separator instead of `:`
- More familiar to users of luckperms
- Avoids confusion with namespaced identifiers.

crates/valence_command_derive/src/lib.rs
```diff
@@ -0,0 +1,402 @@
+use proc_macro::TokenStream;
+use proc_macro2::{Ident, TokenTree};
+use quote::{format_ident, quote};
+use syn::{parse_macro_input, Attribute, Data, DeriveInput, Expr, Fields, Meta};
+
+#[proc_macro_derive(Command, attributes(command, scopes, paths))]
+pub fn derive_command(a_input: TokenStream) -> TokenStream {
```
rj00a, line 7: Panicking should not be used as an error reporting mechanism in proc macros. It's better to bubble up the error with a proper span and use syn's `into_compile_error`. See https://docs.rs/syn/latest/syn/parse/struct.Error.html

> The correct way to report errors back to the compiler from a procedural macro is by emitting an appropriately spanned invocation of [compile_error!](https://doc.rust-lang.org/nightly/core/macro.compile_error.html) in the generated code. This produces a better diagnostic message than simply panicking the macro.

crates/valence_command_derive/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "valence_command_derive"
```
rj00a, line 2:
```suggestion
name = "valence_command_macros"
```

For consistency and in case we want to add more than just derives in the future.

crates/valence_command/src/command_scopes.rs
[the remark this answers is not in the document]
rj00a: Does it make sense for a player to have an identical scope more than once? If not, then this should be a set.

crates/valence_command/src/scopes.rs
```diff
@@ -0,0 +1,281 @@
...
+use bevy_ecs::system::Resource;
+use petgraph::dot;
+use petgraph::dot::Dot;
+use petgraph::prelude::*;
+
+/// Command scope Component for players. this is a list of scopes that a player
+/// has. if a player has a scope, they can use any command that requires
+/// that scope.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Component, Default)]
+pub struct CommandScopes(pub Vec<String>);
```
rj00a, line 53: This should derive `Deref`, `DerefMut` since that has become the established convention.

crates/valence_command/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+ordered-float.workspace = true
+petgraph.workspace = true
+thiserror.workspace = true
+tracing.workspace = true
+
+valence_server = { workspace = true, optional = true }
+valence_text = { workspace = true, optional = true }
+
+[features]
+valence = ["dep:valence_server", "dep:valence_text"]
```
rj00a, line 25: I'm not so sure we should have a `valence` feature flag. The lib makes some deep assumptions about its own use as a bevy plugin for a server, so realistically I don't see it being used elsewhere. The use case is just missing for me.

Feature flags are also annoying to test and easy to screw up, so I like avoiding them if we don't need them.

crates/valence_command/src/parsers/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+            self.cursor = self.input.len();
+        }
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum CommandArgParseError {
+    // these should be player facing and not disclose internal information
+    #[error("invalid argument, expected {0} got {1}")] // e.g. "integer" number
+    InvalidArgument(String, String),
```
rj00a, line 167: I would prefer named fields to make the distinction between the strings clearer. Something like
```suggestion
    InvalidArgument {
        expected: String,
        got: String,
    },
```

crates/valence_command/src/scopes.rs
JackCrumpLeys: if this is common maybe a derive macro would be useful for this.

rj00a: There is. You should use the macro from `derive_more` or `bevy_derive`.

crates/valence_command_derive/src/lib.rs
rj00a: The fix isn't quite what I was looking for because there aren't any precise error spans. I suggest creating a helper for `derive_command` that returns `syn::Result`. You can see how that's done that [here](https://github.com/valence-rs/valence/blob/c3d112df657218c07bf922ee2227db3f3f24748c/crates/valence_protocol_macros/src/decode.rs#L8).

crates/valence_command/src/parsers.rs
```diff
@@ -0,0 +1,269 @@
...
+        self.cursor += 1;
+    }
+
+    pub fn pop(&mut self) -> Option<char> {
+        let c = self.peek();
+        if c.is_some() {
+            self.advance();
+        }
+        c
+    }
```
rj00a, line 82: There's some buggy behavior here. A character is not always one byte wide due to UTF-8 encoding, so `advance` is incorrect. Also, `chars().nth(n)` is an `O(n)` operation.

crates/valence_command/src/parsers.rs
```diff
@@ -0,0 +1,269 @@
...
+    fn parse_arg(input: &mut ParseInput) -> Result<Self, CommandArgParseError>;
+    /// what will the client be sent
+    fn display() -> Parser;
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ParseInput {
+    pub input: String,
+    pub cursor: usize,
+}
```
rj00a, line 51: I would prefer if `ParseInput` looked like

```rust
pub struct ParseInput<'a>(pub &'a str);
```
or maybe even
```rust
pub trait ParseInput { ... }
impl ParseInput for &'_ str { ... }
```

This way, we can just shrink the string from the front in order to advance the input. This would eliminate the 1 byte = 1 char bugs mentioned below, and also prevents users from moving the cursor backwards.

crates/valence_command/src/parsers.rs
```diff
@@ -0,0 +1,269 @@
...
+
+impl ParseInput {
```
rj00a, line 53: There's some buggy behavior here. A character is not always one byte wide due to UTF-8 encoding. The code below sometimes assumes `cursor` is a count of bytes, while other times it's used as a count of chars. Also, `chars().nth(n)` has to walk the entire length of the string to get to `n`.

crates/valence_command/src/handler.rs
```diff
@@ -0,0 +1,123 @@
...
+    mut events: EventWriter<CommandResultEvent<T>>,
+    command: ResMut<CommandResource<T>>,
+) where
+    T: Command + Send + Sync + Debug,
+{
+    for command_event in commands_executed.iter() {
+        if command.executables.contains_key(&command_event.node) {
+            let result = command.executables.get(&command_event.node).unwrap()(
+                &mut ParseInput::new(&command_event.command),
+            );
```
rj00a, line 115:
```suggestion
        if let Some(executable) = command.executables.get(&command_event.node) {
            let result = executable(&mut ParseInput::new(&command_event.command));
```

crates/valence_command/src/scopes.rs
```diff
@@ -0,0 +1,304 @@
...
+    /// add a scope to this component
+    pub fn add(&mut self, scope: &str) {
+        self.0.insert(scope.into());
+    }
+
+    /// remove a scope from this component
+    pub fn remove(&mut self, scope: &str) {
+        // let scope = scope.into();
+        self.0.retain(|p| p != scope);
+    }
```
rj00a, line 99:
```suggestion
```

examples/command.rs
```diff
@@ -0,0 +1,642 @@
...
+enum Gamemode {
+    #[paths("survival {target?}", "{/} gms {target?}")]
+    Survival { target: Option<EntitySelector> },
+    #[paths("creative {target?}", "{/} gmc {target?}")]
+    Creative { target: Option<EntitySelector> },
+    #[paths("adventure {target?}", "{/} gma {target?}")]
+    Adventure { target: Option<EntitySelector> },
+    #[paths("spectator {target?}", "{/} gmspec {target?}")]
+    Spectator { target: Option<EntitySelector> },
+}
```
rj00a, line 53: I tried to make a basic version of the gamemode command and wrote this.

```rust
#[derive(Command, Debug, Clone)]
#[paths("gamemode {gamemode} {target?}")]
struct GameModeCommand2 {
    gamemode: GameMode,
    target: Option<EntitySelector>,
}
```

But it doesn't compile because only `enum` is allowed. So I had to write this instead.

```rust
#[derive(Command, Debug, Clone)]
#[paths("gamemode")]
enum GameModeCommand {
    #[paths("{gamemode} {target?}")]
    GameMode {
        gamemode: GameMode,
        target: Option<EntitySelector>,
    }
}
```

Can we change the macro to support `struct` commands?

examples/command.rs
```diff
@@ -0,0 +1,642 @@
...
+    TargetToLocation {
+        target: EntitySelector,
+        location: Vec3Parser,
+    },
+}
+
+#[derive(Command, Debug, Clone)]
+#[paths("gamemode", "gm")]
+#[scopes("valence.command.gamemode")]
+enum Gamemode {
```
rj00a, line 44: Not sure this is the best command to use as an example because people like me will be wondering why we're not just using the normal `GameMode` parser.

examples/command.rs
rj00a, line 1: Can we suffix the commands with `Command`? I think that would be a nice convention to stick to.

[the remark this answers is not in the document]
JackCrumpLeys: It is up to whomever handles raw strings to send warnings (the client will show red text with the current parsing). Trying to guess what node a client is talking about with a broken or incomplete command is a bad idea.

examples/command.rs
JackCrumpLeys: I like how the current version shows multiple paths or redirections. we could add a comment if we must.

examples/command.rs
```diff
@@ -0,0 +1,642 @@
...
+    // this is technically unreachable)
+    #[paths = "a {a} {b} b {c?}"]
+    B {
+        a: Vec3Parser,
+        b: GreedyString,
+        c: Option<String>,
+    },
+    // greedy string optional arg
+    #[paths = "a {a} b {b?}"]
+    C { a: String, b: Option<GreedyString> },
```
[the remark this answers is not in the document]
JackCrumpLeys: fixed

examples/command.rs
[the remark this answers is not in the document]
rj00a: But how do I get access to the error that occurred so I can send an error in the chat or whatever else? It looks like the error is thrown away.

https://github.com/valence-rs/valence/pull/446/files#diff-494de62f0e3a6fa84f8d7e1d109af30f30d37abba15f0271dff866d8512c20c1R502-R503

crates/valence_command/src/parsers.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    pub fn advance_to_next_non_whitespace(&mut self) {
+        while let Some(c) = self.peek() {
+            if c.is_whitespace() {
+                self.advance();
+            } else {
+                break;
+            }
+        }
+    }
```
rj00a, line 136: this is exactly the same as `skip_whitespace`

```suggestion
```

crates/valence_command/src/parsers.rs
```diff
@@ -0,0 +1,302 @@
...
+            error!("peek_n(0) called, don't do that.");
+            return None; // never peek 0 chars
+        }
+        if n > self.input.chars().count() {
+            return Some(self.input);
+        }
+        Some(&self.input[..=self.input.char_indices().nth(n - 1)?.0])
+    }
+
+    pub fn peek_word(&self) -> String {
```
rj00a, line 77: This could return a slice.

crates/valence_command/src/parsers.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+///
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ParseInput<'a> {
+    input: &'a str,
+    traversed: usize,
+}
+
+impl<'a> ParseInput<'a> {
```
rj00a, line 55: The `advance*` functions don't need to exist because we have `pop*` which do basically the same thing.

crates/valence_command/src/parsers.rs
```diff
@@ -0,0 +1,302 @@
...
+    fn parse_arg(input: &mut ParseInput) -> Result<Self, CommandArgParseError>;
+    /// what will the client be sent
+    fn display() -> Parser;
+}
+
+///
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ParseInput<'a> {
+    input: &'a str,
+    traversed: usize,
```
rj00a, line 52: I'm a bit skeptical that this field needs to exist. Can't you take a diff of the input before and after parsing to determine how many characters were advanced?

examples/command.rs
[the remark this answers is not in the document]
JackCrumpLeys: The thing is when trying to match a node an error is assumed to mean that the command doesn't match the given node. For example if we have two executables at /foo and one is a number arg and the other is a string arg /foo bar should not trigger "expected number got bar".

crates/valence_command/src/parsers.rs
JackCrumpLeys: That is much more costly.

JackCrumpLeys: I mean we COULD use pointers (original str pointer with length of new str pointer - old str pointer)

examples/command.rs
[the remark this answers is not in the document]
rj00a: I had a look at the source for `GameModeArgumentType` and it throws an exception when the gamemode fails to parse.

```java
private static final DynamicCommandExceptionType INVALID_GAME_MODE_EXCEPTION = new DynamicCommandExceptionType(gameMode -> Text.translatable("argument.gamemode.invalid", gameMode));
```

`argument.gamemode.invalid` translates to the message the user sees when they type `/gamemode aklwjdlkawlhd`.

So clearly the error is bubbling up somehow, but I don't know what the semantics are exactly. Might need to dig into the brigadier source.

crates/valence_command/src/parsers.rs
```diff
@@ -0,0 +1,327 @@
...
+        let mut far_ptr = self.0.as_ptr();
+        let mut near_ptr = other.as_ptr();
+
+        if near_ptr > far_ptr {
+            std::mem::swap(&mut near_ptr, &mut far_ptr);
+        }
+        let new_len = far_ptr.offset_from(near_ptr) as usize;
+
+        std::str::from_utf8_unchecked(std::slice::from_raw_parts(near_ptr, new_len))
+    }
```
rj00a, line 189: Not a fan of the unsafe code here. I'm also not confident its use is entirely sound.

JackCrumpLeys: It is either this or storing some kind of cursor or traversed chars I'm afraid. The only unsafely is `far_ptr.offset_from` and `from_utf8_unchecked` as long as both slices are from the same string it should be completely fine.

examples/command.rs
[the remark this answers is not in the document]
JackCrumpLeys: We would have to make some assumptions 1. no other nodes match 2. no better fit for error reporting. This applies fine to the small set of uncomplicated vinnila commands but anything more complex and dynamic will cause errors in the future. OTOH We could label exclusive args somehow (the only arg reachable at a certain path) and bubble those errors up.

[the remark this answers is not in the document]
rj00a: Maybe vanilla is throwing away all but the last exception? Just a guess, but I'd like to get some confirmation so we can have more confidence in a decision.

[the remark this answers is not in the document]
JackCrumpLeys: How about we hash args based on the path taken to get to them and if the hash is unique then we report errors when the last arg does not match. This only solves the problem for unique beaches (most cases). This would not be too hard to implement, I can probably do it in about 30m to an hour tomorrow.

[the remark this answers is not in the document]
rj00a: But what is brigadier doing to resolve this situation? I doubt it's anything like that.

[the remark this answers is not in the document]
JackCrumpLeys: I have not read brigadier's code and I am not well versed on java so I don't particularly want to track it down.

PR #457, opened 2023-08-08
Tool for downloading the vanilla server for use with the packet inspector.

rj00a:
# Objective

I've created a simple Rust script to download the vanilla server from Mojang and configure it for use with the packet inspector. This makes it easier for folks to examine vanilla server behavior using the packet inspector without configuring a bunch of things first.

# Solution

Created `tools/download_vanilla_server`.

## Problem

I've tried to configure the server to generate a superflat world, but it doesn't seem to be working. The server keeps overwriting `WorldGenSettings` in the `level.dat` and `level-type=flat` isn't working.

I'm also not sure if we're allowed to agree to the EULA on the user's behalf, but to be safe I've not done that.


dyc3:
I wrote up some instructions to do this with docker in the packet inspector readme a whole ago https://github.com/valence-rs/valence/tree/main/tools/packet_inspector#quick-start-with-vanilla-server-via-docker

Does they not work anymore?

rj00a:
Ok now that I'm looking at it more closely, there are a lot of environment variables we can give `docker run` to achieve the same result so I'll just close this.

```bash
docker run -e EULA=TRUE -e ONLINE_MODE=false -e ANNOUNCE_PLAYER_ACHIEVEMENTS=false -e GENERATE_STRUCTURES=false -e SPAWN_ANIMALS=false -e SPAWN_MONSTERS=false -e SPAWN_NPCS=false -e SPAWN_PROTECTION=0 -e VIEW_DISTANCE=16 -e MODE=creative -e LEVEL_TYPE=flat -e RCON_CMDS_STARTUP="gamerule doWeatherCycle false" -d -p 25565:25565 --name mc itzg/minecraft-server
```

PR #461, opened 2023-08-11
Fix `update_selected_slot_c2s`

Bafbi:
# Objective

This packet wasn't right anymore because I had change it thinking it was odd that the S2c and C2s don't use the same primitive size, but it create warning so I reverse it.

# Solution

Change u8 to u16 so the packet can be decode without warning.


Bafbi:
Are we sure that the 'update_slot_s2c' is not also a u16?

PR #462, opened 2023-08-11
Custom skin example

### LINKED ISSUE #460: Custom skins for NPC Player entities

It's common to have custom NPCs on minigame servers with custom skins.

There should be an optional component that you can attach to a Player entity to change it's texture.

None

## Additional context

Apparently, this is possible in Bukkit by setting the `textures` property on `GameProfile`. I haven't found a good specific code example yet other than [this forum post](https://www.spigotmc.org/threads/set-custom-skins-to-players.431292/?__cf_chl_tk=Xzmp8JTDE3xb6jHwSa4bNShwT3BX.kEKTlC2BVIH90s-1691589807-0-gaNycGzNC-U).

dyc3:
- bare minimum to do custom skins on player entities
- add a `CustomSkin` component

# Objective

This PR aims to make it a little bit easier to create player entities with custom skins. This is common for things like NPCs.

closes #460

# Solution

- Added a new example.
- Ideally, we could provide the `CustomSkin` component inside valence itself, maybe in a new `valence_util` crate?

### DIFF
--- a/crates/valence_server/src/client.rs
+++ b/crates/valence_server/src/client.rs
@@ -444,6 +444,32 @@ impl Properties {
     pub fn textures_mut(&mut self) -> Option<&mut Property> {
         self.0.iter_mut().find(|p| p.name == "textures")
     }
+
+    /// Returns the value of the "textures" property. It's a base64-encoded
+    /// JSON string that contains the skin and cape URLs.
+    pub fn skin(&self) -> Option<&str> {
+        self.textures().map(|p| p.value.as_str())
+    }
+
+    /// Sets the value of the "textures" property, or adds it if it does not
+    /// exist. Can be used for custom skins on player entities.
+    ///
+    /// `signature` is the Yggdrasil signature for the texture data. It is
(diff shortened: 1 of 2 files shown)


rj00a:
Instead of a new component, perhaps we should use a custom world query to avoid duplicating data.

dyc3:
@rj00a Would you mind elaborating on your idea there? I might have some spare time soon to fix this up.

rj00a:
I was thinking something like this:

```rust
#[derive(WorldQuery)]
pub struct SkinQuery {
    pub properties: &'static mut Properties,
}

impl SkinQueryItem<'_> {
    pub fn skin(&self) {}
    pub fn cape(&self) {}
    pub fn set_skin(&mut self) {}
    pub fn set_cape(&mut self) {}
}

impl SkinQueryReadOnlyItem<'_> {
    pub fn skin(&self) {}
    pub fn cape(&self) {}
}
```

rj00a:
wait, we can just make those methods on `Properties`.

rj00a (approved):

PR #463, opened 2023-08-11
Update pkg meta

rj00a:
# Objective

- Set up Valence so it's ready to publish on crates.io (an alpha version for all crates has already been published)

# Solution

- Changed some package names.
- Added more info for crates.io
- Added symlink spaghetti for `extracted/` because data outside the package root can't be included directly.


dyc3 (approved):

crates/valence_ident/Cargo.toml
```diff
@@ -1,5 +1,7 @@
 [package]
 name = "valence_ident"
+description = "Parse and create Minecraft's resource locations"
```
Bafbi, line 3:
```suggestion
description = "Parse and create Minecraft's namespaces"
```
I'm maybe wrong but this terminology seems more common around users. (I personally never had ear of resource location before and yes I know on the wiki it's resource location)

dyc3: If we're nitpicking, it would probably make more sense to call them "resource identifiers", similar to the URI acronym.

PR #478, opened 2023-08-18
Add ahash feature to valence_nbt

Earthcomputer:
# Objective

- Add the option to use the `ahash` hasher in `valence_nbt`. This hasher is faster than the std hasher, at the cost of being less resistant to hash DOS attacks.

# Solution

- Added an `ahash` feature to the `valence_nbt` crate to allow the user to optionally select this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,7 @@ license = "MIT"
 
 [workspace.dependencies]
 aes = "0.8.2"
+ahash = "0.8.3"
 anyhow = { version = "1.0.70", features = ["backtrace"] }
 approx = "0.5.1"
 arrayvec = "0.7.2"
--- a/crates/valence_nbt/Cargo.toml
+++ b/crates/valence_nbt/Cargo.toml
@@ -10,6 +10,7 @@ edition.workspace = true
 repository.workspace = true
 
 [features]
+ahash = ["dep:ahash"]
 binary = ["dep:byteorder", "dep:cesu8"]
 snbt = []
 # When enabled, the order of fields in compounds are preserved.
@@ -17,6 +18,7 @@ preserve_order = ["dep:indexmap"]
 serde = ["dep:serde", "dep:thiserror", "indexmap?/serde"]
 
 [dependencies]
+ahash = { workspace = true, optional = true }
 byteorder = { workspace = true, optional = true }
 cesu8 = { workspace = true, optional = true }
 indexmap = { workspace = true, optional = true }
--- a/crates/valence_nbt/src/compound.rs
+++ b/crates/valence_nbt/src/compound.rs
@@ -20,8 +20,11 @@ pub struct Compound {
 #[cfg(not(feature = "preserve_order"))]
 type Map = std::collections::BTreeMap<String, Value>;
 
-#[cfg(feature = "preserve_order")]
+#[cfg(all(feature = "preserve_order", not(feature = "ahash")))]
 type Map = indexmap::IndexMap<String, Value>;
+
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -97,6 +97,7 @@ license = "MIT"
 
 [workspace.dependencies]
 aes = "0.8.2"
+ahash = "0.8.3"
```
dyc3, line 100: Shouldn't this be marked as optional?

Earthcomputer: No idea how workspace dependencies work, should it?

dyc3: Ah, I just looked and we don't do this for all of the adjacent dependencies, so this is probably fine.

dyc3 (approved):
Once those jobs are passing I'm fine to merge this.

rj00a:
I think we should leave the choice of hasher as an implementation detail and rename the feature to `fast_hash`.

dyc3:
I disagree with renaming it to `fast_hash` because that could imply to the end user that it's always a good thing to enable. The reality is that there are caveats that need to be evaluated before enabling.

That being said, would it be possible to do this using generics instead? That would probably be the best way to go about it if possible.

rj00a:
> I disagree with renaming it to fast_hash because that could imply to the end user that it's always a good thing to enable. The reality is that there are caveats that need to be evaluated before enabling.

That can be solved with some documentation. In particular, "HashDoS resistance may be weakened by enabling this feature".

My main concern is that if we ever wanted to change the hasher (`rustc-hash`?), renaming the `ahash` feature would be a breaking change. We can't just add another feature for the new hasher because it would be incompatible with the existing `ahash` feature. Most people probably just want things to go faster and don't really care about the specifics.

> That being said, would it be possible to do this using generics instead? That would probably be the best way to go about it if possible.

I can agree with that. We'd probably want to make `Compound` generic over `K = String`, `V = Value`, `S = std::collections::hash_map::RandomState`. And then make `Value` and `List` generic over `S`. However, this might cause type inference problems if the `compound!` macro were generic too.

It's worth pointing out that `serde_json` decided _not_ to make their `Map` type generic over the hasher and has no feature to change things. @Earthcomputer do you have a use case where you need high performance while preserving insertion order? I suspect that the `BTreeMap` is going to offer better performance in most realistic scenarios, even with a faster hasher for the `IndexMap`.

Earthcomputer:
To be honest, I didn't know that `BTreeMap` performed so well, I'm coming from Java where the only reason you would use a tree map is if you want order, so I assumed that was the purpose here.

rj00a:
Updated benchmarks from https://github.com/Moulberry/Graphite/blob/master/crates/graphite_binary/benches/nbt.rs

With BTreeMap:
```
valence_parse_bigtest   time:   [1.9596 µs 1.9632 µs 1.9678 µs]
```
With HashMap and default hasher:
```
valence_parse_bigtest   time:   [2.5210 µs 2.5225 µs 2.5241 µs]
```
With HashMap and ahash:
```
valence_parse_bigtest   time:   [2.3527 µs 2.3553 µs 2.3585 µs]
```
With IndexMap and default hasher:
```
valence_parse_bigtest   time:   [2.5948 µs 2.5969 µs 2.5992 µs]
```
With IndexMap and ahash:
```
valence_parse_bigtest   time:   [2.4413 µs 2.4432 µs 2.4454 µs]
```

---

Considering this info, adding the feature doesn't seem worth it to me. But I wouldn't mind revisiting if there are objections.

dyc3:
Yeah I agree. Let's hold off on this for now to avoid the complexity.

PR #479, opened 2023-08-18
Add Deref and DerefMut to all components (where appropriate)

### LINKED ISSUE #444: Add `Deref` and `DerefMut` to all components (where appropriate)

Accessing the data inside components is a pain. Writing `.0` or `.get()` everywhere is cumbersome and it's easy to forget.

Derive `Deref` and `DerefMut` on all non-ZSTs implementing `Component` by using the `derive_more` crate. This should be done in a way that does not violate the privacy of the inner data, so for example a component that only provides read-only access to the inner data should not derive `DerefMut`.

IoaNNUwU:
## Objective

- Add `Deref` and `DerefMut` to all components using `derive_more` crate
- Closes #444

## Solution

- Add `derive_more` crate to top level `Cargo.toml` and use cargo workspace dependency in necessary crates
- Derive `DerefMut` trait to components that provide mutable access to underlying data
- Derive `Deref` trait to components that provide immutable access to underlying data

## Compile times

I made some compilation time benchmarks on my machine (i don't know good way to do compile time benchmarks so i ran debug mode compilation couple of times before and after the change)

Clean `Valence` compilation
- Before: 2m 2s
- After: 2m 5s

Clean default `playground` example
- Before: 11s
- After: 11s

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -118,6 +118,7 @@ cesu8 = "1.1.0"
 cfb8 = "0.8.1"
 clap = { version = "4.0.30", features = ["derive"] }
 criterion = "0.5.1"
+derive_more = "0.99.17"
 directories = "5.0.0"
 eframe = { version = "0.22.0", default-features = false }
 egui = "0.22.0"
--- a/crates/valence_advancement/Cargo.toml
+++ b/crates/valence_advancement/Cargo.toml
@@ -13,4 +13,5 @@ valence_server.workspace = true
 bevy_app.workspace = true
 bevy_ecs.workspace = true
 bevy_hierarchy.workspace = true
+derive_more.workspace = true
 rustc-hash.workspace = true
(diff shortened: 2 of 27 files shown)


dyc3:
Good to see that it doesn't really seem to affect compile times. What CPU do you have?

IoaNNUwU:
My CPU is Ryzen 5500

dyc3 (approved):
Looks good. Thanks!

rj00a:
I would prefer to use the `1.0.0-beta.3` version of `derive_more` here.

dyc3:
I just took a look at the [`derive_more` changelog](https://github.com/JelteF/derive_more/blob/v1.0.0-beta.3/CHANGELOG.md). The only thing I saw that could affect us here is that the default for v1 is to have all features disabled, but I think that's a minor issue.

rj00a:
I would actually suggest using the macros from `bevy_derive` since we already transitively depend on it, but I recently started using `derive_more 1.0.0-beta.3` in my `better-packets` branch for `valence_protocol`.

rj00a:
Looks like this missed the generated components in `valence_entity`.

PR #482, opened 2023-08-20
`valence_nbt::List` convenience methods

Earthcomputer:
# Objective

- `valence_nbt::List` is kind of inconvenient to work with when you are not expecting a specific type of list. You often have to match over every type of list.

Just one example, I would like to write this code:
```rs
fn do_something_with_nbt_value(value: valence_nbt::Value) { ... }

for item in list {
   do_something_with_nbt_value(item);
}
```

# Solution

- Added several convenience methods to `valence_nbt::List` to hopefully reduce the `match` boilerplate in client code, including iterator methods.
- I had to introduce borrowing `ValueRef` and `ValueRefMut` types. These are only used for list access and modification. I cannot use `&Value` because the list does not store that type so there is nothing to reference, and worse, `&mut Value` would allow for non-heterogeneous lists.

### DIFF
--- a/crates/valence_nbt/src/lib.rs
+++ b/crates/valence_nbt/src/lib.rs
@@ -20,12 +20,14 @@
 use std::mem::ManuallyDrop;
 
 pub use compound::Compound;
+pub use list::List;
 pub use tag::Tag;
-pub use value::{List, Value};
+pub use value::Value;
 
 #[cfg(feature = "binary")]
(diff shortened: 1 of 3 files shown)


Earthcomputer:
Not sure why the CI is failing

rj00a (commented):
Looks good, thanks. Just a few minor things.

PR #485, opened 2023-08-21
Tweak game of life example

rj00a:
# Objective

Code in game of life example could be improved a bit

# Solution

do that

### DIFF
--- a/examples/game_of_life.rs
+++ b/examples/game_of_life.rs
@@ -64,7 +64,7 @@ fn setup(
     commands.spawn(layer);
 
     commands.insert_resource(LifeBoard {
-        paused: true,
+        playing: false,
         board: vec![false; BOARD_SIZE_X * BOARD_SIZE_Z].into(),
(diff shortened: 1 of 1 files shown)


examples/game_of_life.rs
```diff
@@ -210,18 +210,18 @@ fn update_board(
...
-                if board.paused {
-                    client.set_action_bar("Paused".italic().color(Color::RED));
-                } else {
-                    client.set_action_bar("Playing".italic().color(Color::GREEN));
-                }
+            if board.playing {
+                board.playing = false;
+                layer.set_action_bar("Paused".italic().color(Color::RED));
+            } else {
+                board.playing = true;
```
CodeDoctorDE, line 223: Why didn't you keep `board.paused = !board.paused`?

CodeDoctorDE: For me a toggle of playing is clearer when using the old code.
But otherwise I like the new entity layer system.

PR #493, opened 2023-08-24
Add warning to rustdocs when building from main branch

rj00a:
# Objective

When viewing the API docs from valence.rs, there is no obvious indication that they are for the `main` branch. This could lead to confusion in the future.

# Solution

Add a warning to the docs, but only when building from `website/build.sh`.

> **❗ NOTE:** This documentation is sourced from the `main` branch. If you're looking for the most recent stable release, go [here](https://docs.rs/valence/latest/valence/).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,8 @@
-#![doc = include_str!("../README.md")] // Points to the main project README.
+#![cfg_attr(
+    unstable_doc,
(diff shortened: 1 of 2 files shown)


dyc3 (approved):

rj00a:
Note to self: https://github.com/rust-lang/rust/pull/106561

Issue #359, opened 2023-06-08
Extractor output is non-determinisic.


dyc3:
## Valence Version

6338fc630073bda6fdcb09e6acbb72ba87ea4a7c

## What You Did

Run the extractor

## What Went Wrong

Running the extractor does not produce exactly the same output every time. The overall structure will be the same, but fields and items in arrays may be in different orders.

This causes very messy diffs when Minecraft updates, or when a new extractor is added.

rj00a:
The problem is caused by an incorrect assumption. The extractor assumes that the default value for a particular tracked data field is the same across entity classes that inherit the field, but this is incorrect. The output format of the extractor will have to be changed along with the entity code generator in `valence_entity`.

Rather than having one system per tracked data, there needs to be one system per entity archetype, where the tracked data is updated only for that specific archetype.

PR #496, opened 2023-08-25
Make player health 20.0 by default

jivvy:
# Objective

When testing making a minigame example, I noticed the player always spawned with 6 health, which is the default for living entities in Valence.

It would also cause a desync between the client and server, because the client implies that it has 20.0 health upon spawning.

### DIFF
--- a/crates/valence_server/src/client.rs
+++ b/crates/valence_server/src/client.rs
@@ -13,6 +13,7 @@ use bytes::{Bytes, BytesMut};
 use derive_more::{Deref, DerefMut, From, Into};
 use tracing::warn;
 use uuid::Uuid;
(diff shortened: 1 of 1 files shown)


dyc3 (approved):
Yeah I noticed this when working on the ctf example. LGTM.

dyc3:
The CI failures are unrelated.

rj00a (changes requested):
This is caused by https://github.com/valence-rs/valence/issues/359#issuecomment-1582405754. As a result, there are many other places where default values are incorrect. 

If we must have a temp fix now, we should edit the JSON file that has the default health value instead.

dyc3:
In that case, let's close this and open an issue for it instead.

PR #497, opened 2023-08-25
hotfix clippy 1.72 lints

dyc3:
This fixes the recent CI failures we've been getting.

### DIFF
--- a/crates/valence_text/src/tests.rs
+++ b/crates/valence_text/src/tests.rs
@@ -64,7 +64,7 @@ fn selector() {
     let deserialized = Text::from_str(&serialized).unwrap();
     assert_eq!(
         serialized,
-        r##"{"selector":"foo","separator":{"text":"bar","color":"red","bold":true}}"##
+        r#"{"selector":"foo","separator":{"text":"bar","color":"red","bold":true}}"#
     );
     assert_eq!(txt, deserialized);
 }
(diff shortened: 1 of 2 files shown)


AviiNL:
Clippy errors on it for a reason, no? This feels more like a duct tape patch than an actual fix, I'm fine with getting this merged in as a hotfix, but shouldnt the underlying problem not get fixed instead?

crates/valence_text/src/tests.rs
```diff
@@ -64,7 +64,7 @@ fn selector() {
     let deserialized = Text::from_str(&serialized).unwrap();
     assert_eq!(
         serialized,
-        r##"{"selector":"foo","separator":{"text":"bar","color":"red","bold":true}}"##
```
AviiNL, line 67: this looks unrelated, should this be in this pr?

dyc3: This was caught in a clippy lint when I ran it.

AviiNL: okay 👍 
LGTM then

rj00a:
> Clippy errors on it for a reason, no? This feels more like a duct tape patch than an actual fix, I'm fine with getting this merged in as a hotfix, but shouldnt the underlying problem not get fixed instead?

Snbt module is in need of a full rewrite so I'm fine with this temp fix.

PR #499, opened 2023-08-28
Add valence_nbt::List::retain

Earthcomputer:
# Objective

- It was annoying to remove elements from a `valence_nbt::List` by a predicate.

# Solution

- Add `valence_nbt::List::retain`, which the same API as `Vec::retain_mut` (similar to `HashMap::retain`, `Vec::retain_mut` only exists because they forgot to pass a mutable reference to `Vec::retain` and changing it would be breaking).

### DIFF
--- a/crates/valence_nbt/src/list.rs
+++ b/crates/valence_nbt/src/list.rs
@@ -374,7 +374,7 @@ impl List {
             panic!("removal index (is {index}) should be < len (is {len})");
         }
 
-        match self {
+        let removed = match self {
             List::End => assert_failed(index, 0),
             List::Byte(list) => Value::Byte(list.remove(index)),
             List::Short(list) => Value::Short(list.remove(index)),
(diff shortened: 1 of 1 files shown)


crates/valence_nbt/src/list.rs
```diff
@@ -388,6 +388,44 @@ impl List {
...
+    /// Returns only the elements specified by the predicate, passing a mutable
+    /// reference to it.
+    ///
+    /// In other words, removes all elements `e` such that `f(ValueMut(&mut e))`
+    /// returns `false`. This method operates in place, visiting each element
+    /// exactly once in the original order, and preserves the order of the
+    /// retained elements.
+    pub fn retain<F>(&mut self, mut f: F)
+    where
+        F: FnMut(ValueMut) -> bool,
```
dyc3, line 409: We shouldn't borrow the values in the list as mutable.

dyc3 (changes requested):

crates/valence_nbt/src/list.rs
Earthcomputer: Why not?

dyc3: It generally bad practice to make things mutable when they don't need to be, and I feel like we'd be encouraging some messy antipatterns. The way `retain` works is it uses a predicate which should make assertions about the value it's given in a read-only manor.

The [standard library does this](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain). It delegates the mutable version to `retain_mut`.

Earthcomputer: See the explanation in the OP

dyc3: I'll have to defer this to @rj00a

rj00a: I agree with @Earthcomputer on this one.

rj00a (approved):
LGTM

dyc3 (approved):
Sounds good then.

PR #500, opened 2023-08-28
Replace Option<ItemStack> with ItemStack

### LINKED ISSUE #490: Replace `Option<ItemStack>` with `ItemStack`

In Valence we are using `Option` to indicate the emptyness of an itemstack. However, vanilla does something semantically different.

```java
public boolean isEmpty() {
    return this == EMPTY || this.item == Items.AIR || this.count <= 0;
}
```

This has an impact in the protocol

```java
public PacketByteBuf writeItemStack(ItemStack stack) {
    if (stack.isEmpty()) {
        this.writeBoolean(false);
    } else {
        this.writeBoolean(true);
        Item item = stack.getItem();
        this.writeRegistryValue(Registries.ITEM, item);
        this.writeByte(stack.getCount());
        NbtCompound nbtCompound = null;
        if (item.isDamageable() || item.isNbtSynced()) {
            nbtCompound = stack.getNbt();
        }
        this.writeNbt(nbtCompound);
    }
    return this;
}
```

- Replace all occurrences of `Option<ItemStack>` with `ItemStack`.
- Add `is_empty` method to `ItemStack` with same semantics as vanilla `isEmpty`.
- Make the item count in `ItemStack` public and an `i8`.
- `impl {Encode, Decode} for ItemStack`.

IoaNNUwU:
## Objective

- Fixes #490 

## Solution
- Replace all occurrences of `Option<ItemStack>` with `ItemStack`.
- Add is_empty method to ItemStack with same semantics as vanilla isEmpty. 
- Add is_not_empty method to ItemStack.
- Make the item count in ItemStack public and an i8.
- impl {Encode, Decode} for ItemStack.


IoaNNUwU:
Hi, here is the progress so far, but i have a question about ItemStack.count:

Vanilla minecrafts tests if item stack is empty by checking if it is `AIR` or `count <= 0` but `ItemStack` in valence does a lot to ensure count is in `1..=127` (ItemStack::STACK_MIN, STACK_MAX). As far as i can understand:
- Exposing count as pub value violates this guarantee so i think it should be private.
- There is no need to check if `count <= 0` in `ItemStack.is_empty`. So only way to tell if item stack is empty is to check if it as `AIR` which is not exactly vanilla behaviour.
- If `ItemStack.count` is `i8` there is no need to use clamp because we can only check lowerbound.
- Does ItemStack::STACK_MAX need to be 127? I think sometimes it is better to clamp count  to `1..=64` or `1..=ItemKind.max_stack`? Maybe it is a good idea to add additional constructors to `ItemStack` or use wrapper struct in the future?

dyc3:
Yeah, the item stack count max needs to be 127 because the client can display that. We were enforcing `1..=127` because having a `Some<ItemStack>` with `count <= 0` makes no sense. Enforcing `1..=ItemKind.max_stack` is, IMO, vanilla behavior so it's out of scope for valence.

So I think it'll be fine to change the range to `0..=127`.

crates/valence_protocol/src/item.rs
```diff
@@ -63,38 +76,31 @@ impl Default for ItemStack {
...
-            return Ok(None);
-        }
+            return Ok(ItemStack::empty());
+        };
 
         let item = ItemKind::decode(r)?;
-        let count = u8::decode(r)?;
+        let count = i8::decode(r)?;
 
         ensure!(
```
rj00a, line 105: I'd remove this check and add one at the end:
```rust
// Normalize empty item stacks.
if stack.is_empty() {
    ItemStack::EMPTY
} else {
    stack
}
```

This would prevent some weird data leaking downstream, which I imagine players could use to exploit bugs. Senders aren't even supposed to be able to send empty item stacks that get past the `!present` check, so it seems fine to me that we're "losing" some information by doing this.

From what I can tell, vanilla _doesn't_ do this check, which feels like a mistake to me.

```java
public ItemStack readItemStack() {
    if (!this.readBoolean()) {
        return ItemStack.EMPTY;
    }
    Item item = this.readRegistryValue(Registries.ITEM);
    byte i = this.readByte();
    ItemStack itemStack = new ItemStack(item, (int)i);
    itemStack.setNbt(this.readNbt());
    return itemStack;
}
```

rj00a (commented):
(Sorry if these are things you already planned to fix)

rj00a:
> So I think it'll be fine to change the range to `0..=127`.

Did you mean to write `-128..=127` here since I suggested making the `count: i8` field public, or are you suggesting that we should enforce nonnegative counts?

dyc3:
Yeah, negative item counts don't really make sense. It's not really a big deal though, it would probably be better to be able to remove the clamping altogether.

crates/valence_protocol/src/item.rs
```diff
@@ -46,15 +51,23 @@ impl ItemStack {
...
     }
 
     /// Sets the number of items in this stack. Values are clamped to 1-127,
     /// which are the positive values accepted by clients.
-    pub fn set_count(&mut self, count: u8) {
+    pub fn set_count(&mut self, count: i8) {
         self.count = count.clamp(Self::STACK_MIN, Self::STACK_MAX);
     }
+
+    pub fn is_empty(&self) -> bool {
```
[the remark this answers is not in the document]
IoaNNUwU: as far as i can understand, `is_empty` cannot be const because it uses `==` which is not const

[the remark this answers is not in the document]
rj00a: I suppose you could use `matches!` to get around that limitation.

crates/valence_protocol/src/item.rs
```diff
@@ -46,15 +51,23 @@ impl ItemStack {
...
+        self.item == ItemKind::Air || self.count <= 0
+    }
+
+    pub fn is_not_empty(&self) -> bool {
+        !self.is_empty()
+    }
 }
 
 impl Default for ItemStack {
```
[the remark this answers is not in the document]
rj00a: Yeah, let's do that.

crates/valence_inventory/src/validate.rs
```diff
@@ -272,22 +275,41 @@ pub(super) fn validate_click_slot_packet(
...
+
+                match (old_slot, new_slot) {
+                    // TODO: make sure NBT is the same
+                    // Sometimes, the client will add nbt data to an item if it's missing, like
+                    // "Damage" to a sword
+                    (Some(old_slot), Some(new_slot)) => old_slot != new_slot,
+                    (_, None) => false,
+                    (None, Some(_)) => true,
+                }
             };
```
rj00a, line 299:
```suggestion
            let is_transmuting = match (!old_slot.is_empty(), !new_slot.is_empty()) {
                // TODO: make sure NBT is the same.
                // Sometimes, the client will add nbt data to an item if it's missing, like
                // "Damage" to a sword.
                (true, true) => old_slot.item != new_slot.item,
                (_, false) => false,
                (false, true) => true,
            };
```

crates/valence_inventory/src/validate.rs
```diff
@@ -336,20 +358,42 @@ fn calculate_net_item_delta(
...
+        } else {
+            Some(new_slot.count as i32)
+        };
+
+        net_item_delta += match (old_slot_count, new_slot_count) {
+            (Some(old), Some(new)) => new - old,
+            (Some(old), None) => -old,
+            (None, Some(new)) => new,
             (None, None) => 0,
         };
```
rj00a, line 379:
```suggestion
        net_item_delta += match (!old_slot.is_empty(), !new_slot.is_empty()) {
            (true, true) => new_slot.count as i32 - old_slot.count as i32,
            (true, false) => -(old_slot.count as i32),
            (false, true) => new_slot.count as i32,
            (false, false) => 0,
        };
```

crates/valence_inventory/src/validate.rs
```diff
@@ -336,20 +358,42 @@ fn calculate_net_item_delta(
...
+    } else {
+        Some(packet.carried_item.count as i32)
+    };
+
+    net_item_delta += match (cursor_item_count, carried_item_count) {
+        (Some(old), Some(new)) => new - old,
+        (Some(old), None) => -old,
+        (None, Some(new)) => new,
         (None, None) => 0,
     };
```
rj00a, line 398:
```suggestion
    net_item_delta += match (!cursor_item.is_empty(), !packet.carried_item.is_empty()) {
        (true, true) => cursor_item.count as i32 - packet.carried_item.count as i32,
        (true, false) => -(cursor_item.count as i32),
        (false, true) => packet.carried_item.count as i32,
        (false, false) => 0,
    };
```

examples/building.rs
```diff
@@ -138,7 +138,8 @@ fn place_blocks(
 
         // get the held item
         let slot_id = held.slot();
-        let Some(stack) = inventory.slot(slot_id) else {
+        let stack = inventory.slot(slot_id);
+        if inventory.slot(slot_id).is_empty() {
```
rj00a, line 142:
```suggestion
        if stack.is_empty() {
```

examples/building.rs
```diff
@@ -151,11 +152,11 @@ fn place_blocks(
         if *game_mode == GameMode::Survival {
             // check if the player has the item in their inventory and remove
             // it.
-            if stack.count() > 1 {
-                let count = stack.count();
+            if stack.count > 1 {
+                let count = stack.count;
                 inventory.set_slot_amount(slot_id, count - 1);
```
rj00a, line 157:
```suggestion
                inventory.set_slot_amount(slot_id, stack.count - 1);
```

IoaNNUwU: image

I think this is fine
```rust
if stack.count > 1 {
    let amount = stack.count - 1;
    inventory.set_slot_amount(slot_id, amount);
} else {
    inventory.set_slot(slot_id, ItemStack::EMPTY);
}
```

rj00a (approved):

PR #501, opened 2023-08-29
Add online mode detection in the Packet Inspector

### LINKED ISSUE #404: Make the packet inspector work when encryption is enabled.

Right now the packet inspector doesn't function correctly when encryption is enabled. It will occasionally display "unknown packet" and stop functioning.

I believe the inspector could intercept the server's request to enable encryption during the handshake process. It could encrypt and decrypt packets as they're sent to/from the server while sending and receiving unencrypted packets to/from the client. The server thinks it's talking to an encrypted client, while the client thinks it's talking to an unencrypted server.

This would save us a lot of headache when testing the examples, since users wouldn't need to modify the `ConnectionMode` to start using the inspector.

At the very least it should detect that encryption is enabled and display an error.

natanalt:
# Objective

- The packet inspector didn't respond properly when the server had encryption enabled
- Fixes #404 

# Solution

- The proxy now listens for authentication packets (LoginHelloS2c), kicking the player out with an appropriate message and shutting down the connection when that happens.
- The proxy code was also refactored a fair bit to allow easier two-way communication between the proxy tasks and the UI code, although so far it's mostly unused (I only wrote code for the CLI to print online mode errors whenever the server requests online mode, the GUI hasn't changed, although I may come back and add more info there in the future).

Also, for the sake of completeness, here's how the error looks like in-game:
Screenshot

### DIFF
--- a/tools/packet_inspector/src/lib.rs
+++ b/tools/packet_inspector/src/lib.rs
@@ -2,99 +2,203 @@ mod packet_io;
 mod packet_registry;
 
 use std::net::SocketAddr;
-use std::sync::{Arc, OnceLock};
+use std::sync::Arc;
+use std::time::Duration;
 
-pub use packet_registry::Packet;
-use tokio::net::TcpStream;
+use anyhow::bail;
+
+use bytes::{BufMut, BytesMut};
+
+use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::RwLock;
+use tokio::task::JoinHandle;
+
 use valence::protocol::decode::PacketFrame;
 use valence::protocol::packets::handshaking::handshake_c2s::HandshakeNextState;
 use valence::protocol::packets::handshaking::HandshakeC2s;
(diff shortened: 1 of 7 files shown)


tools/packet_inspector/src/lib.rs
```diff
@@ -2,99 +2,201 @@ mod packet_io;
...
+
+        Ok(Self {
+            main_task,
+            message_tx,
+            logs_rx,
+            packet_registry,
+        })
     }
 
+    /// Subscribes to the proxy's [`PacketRegistry`].
```
rj00a, line 88: There's a rustdoc warning here (that should have caused CI to fail)

rj00a (commented):
Looks good, thanks. Injecting a disconnect packet for the error message is clever.

Just got to fix that little doc issue.

tools/packet_inspector/src/lib.rs
natanalt: It seems I can never have a day without angering rustdoc somehow huh

Either way, I just pushed a fix

PR #505, opened 2023-08-30
Make Packet Inspector only dependent on valence_protocol

### LINKED ISSUE #502: Make the packet inspector only depend on `valence_protocol`.

The packet inspector depends on the `valence` crate, but it only needs the things from `valence_protocol`.

Remove the dependency on `valence` and depend on `valence_protocol` directly. Then, regenerate the file in `assets/depgraph.svg`.

## Additional context

There was a time when `valence_protocol` didn't exist, so the the current situation was necessary.

natanalt:
# Objective

- The packet inspector depends on the entire `valence` create, when realistically it only uses `valence_protocol` (and also `valence_text`, although that one is reexported by the protocol crate)
- Fixes #502 

# Solution

- Remove `valence` dependency, replace it with `valence_protocol` with `compression` feature enabled, fix up imports wherever it's necessary
- Fix up a single `#[cfg(feature)]` directive in the `valence_protocol` crate, which mistakenly checked for the `encryption` feature instead of `compression`
- Regenerate the dependency graph SVG


rj00a (approved):
Thanks!

PR #510, opened 2023-08-31
Place valence and bevy specific stuff in valence_anvil behind feature flags

### LINKED ISSUE #504: Change valence-anvil to be more general

Currently, there are two problems with `valence-anvil`:
1. It is specific to block state chunks, whereas nowadays Mojang uses the anvil format for entities and POIs too, and possibly other things in the future.
2. It is specific to Bevy and `valence-server`, including a Bevy plugin and requiring a biome registry respectively, which makes it harder to justify the use of this crate as a library outside of Valence, like you can with `valence_nbt`.

There needs to be a lower-level public API to simply get a `Compound` given some chunk coordinates, and to be able to store a new `Compound` to some new or existing chunk coordinates. The blockstate, entity and POI stuff can then be built on top of that, possibly in a separate crate or in the same crate with feature flags. The Bevy plugin could also be either a separate crate or in the same crate with a feature flag.

The `valence-anvil` crate could be left as it is, and I could make an alternative crate which is not specific to Valence. However, I think that not only could 3rd parties benefit to this restructuring, but Valence itself could too. It would help work towards some of the goals in the [0.2 roadmap](https://github.com/valence-rs/valence/issues/288), such as serialization support for anvil worlds, and would also be helpful for loading and saving entities if that is something you want.

## Additional context

I am working on a program to upgrade worlds (like the optimize worlds button in Minecraft), and need an anvil parser to upgrade chunks.

Earthcomputer:
# Objective

- Closes #504 

# Solution

- Place valence and bevy specific stuff in `valence_anvil` behind feature flags.
- Support for entity and POI chunks can easily be added to `DimensionFolder` in the future.
- Support for writing chunks is not included in this PR.


crates/valence_anvil/Cargo.toml
```diff
@@ -9,14 +9,19 @@ repository.workspace = true
...
 flate2.workspace = true
-flume.workspace = true
+flume = { workspace = true, optional = true }
 lru.workspace = true
-num-integer.workspace = true
+num-integer = { workspace = true, optional = true }
 thiserror.workspace = true
 tracing.workspace = true
-valence_server.workspace = true
+valence_nbt.workspace = true
```
rj00a, line 26:
```suggestion
valence_nbt = { workspace = true, features = ["binary"] }
```

crates/valence_anvil/Cargo.toml
```diff
@@ -9,14 +9,19 @@ repository.workspace = true
...
+bevy_ecs = { workspace = true, optional = true }
 byteorder.workspace = true
 flate2.workspace = true
-flume.workspace = true
+flume = { workspace = true, optional = true }
 lru.workspace = true
-num-integer.workspace = true
+num-integer = { workspace = true, optional = true }
 thiserror.workspace = true
 tracing.workspace = true
```
rj00a, line 25:
```suggestion
```

crates/valence_anvil/src/lib.rs
```diff
@@ -17,133 +17,75 @@
...
-use bevy_app::prelude::*;
-use bevy_ecs::prelude::*;
+#[cfg(feature = "bevy")]
+pub use bevy::*;
 use byteorder::{BigEndian, ReadBytesExt};
 use flate2::bufread::{GzDecoder, ZlibDecoder};
-use flume::{Receiver, Sender};
 use lru::LruCache;
+use thiserror::Error;
 use tracing::warn;
```
rj00a, line 31: :thinking: is this used anywhere? I'm surprised there's no warning. I guess the builtin `warn` attribute is getting the compiler confused.

rj00a (commented):
Looks good.

crates/valence_anvil/src/lib.rs
Earthcomputer: Line 44

rj00a: Sorry, I'm pointing at `use tracing::warn;` here.

Earthcomputer:
Looks like `check-depgraph` is failing to post a comment

rj00a:
Hmm yeah, I'll try fixing that separately.

PR #511, opened 2023-09-01
PlaceBlock

### LINKED ISSUE #503: Add block placement rules.

Valence is lacking placement rules, i.e. the rules that should be followed when a player places or breaks a block. This includes the orientation of logs and stairs, redstone dust, torches, etc.

The client will predict these behaviors so it's important that we have direct support for it.

1. Add a new method to `ChunkLayer` called `place_block`, similar to `set_block` but with placement rules.
2. Document the different between `set_block` and `place_block`.
3. Make use of it in the building example.

Putting this in a separate crate/plugin. But that doesn't seem necessary to me.

## Additional context

[Here's what Minestom does.](https://github.com/Minestom/Minestom/tree/master/src/main/java/net/minestom/server/instance/block/rule) Although I think its implementation is incomplete. We should check Vanilla source.

XBagon:
# Objective

Fixes #503

# Solution

Adds `PlaceBlockCommand` (used with `Command::add`), which places a block according to infos about itself, the entity placing the block and its environment. Tries to mimic client behaviour as close as possible.

This is the foundation and definitely hasn't all rules implemented. It's a proposal for the form and helper structs/functions to make implementing the rest of the rules simpler and more elegant. I tried to document my sources, open problems and other thoughts in the comments.

Maybe it makes sense to merge this is in a unfinished state, when most common interactions are handled and it generelly is preferable over `set_block`. At the moment it uses a lot of `unimplemented!` to catch missing features, those could be converted into silent ones.

### DIFF
--- a/crates/valence_server/src/placement.rs
+++ b/crates/valence_server/src/placement.rs
@@ -1,0 +1,895 @@
+use bevy_ecs::{
+    prelude::*,
+    system::{Command, SystemState},
+};
+
+use valence_entity::{hitbox::Hitbox, living::LivingEntity, Look};
+use valence_generated::block::{PropName, PropValue};
+use valence_math::{Aabb, DVec3};
+use valence_protocol::{BlockKind, BlockPos, BlockState, Direction};
+
+use crate::{interact_block::InteractBlockEvent, ChunkLayer};
+
+//from https://minecraft.fandom.com/wiki/Block_states
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub enum PlacementRule {
+    Basic,
+    //For blocks that don't necessarily have special props, but have special rules for what can be placed on top/side of it (put together from https://minecraft.fandom.com/wiki/Opacity/Placement)
+    //Overlap or equality with the set of not opaque blocks?
+    Special,
+    Anvils,
+    AmethystBudsAndAmethystCluster,
+    Bamboo,
+    Banners,
+    Barrel,
+    BasaltAndPolishedBasalt,
+    Beds,
+    Bedrock,
+    Beehive,
+    Beetroots,
+    Bell,
+    BigDripleaf,
+    BlastFurnace,
+    BlockOfBambooAndBlockOfStrippedBamboo,
+    BoneBlock,
+    Border,
+    BrewingStand,
+    BubbleColumn,
+    Buttons,
+    Cactus,
+    Cake,
+    Campfire,
+    Candles,
+    Carpets,
+    Carrots,
+    Cauldron,
+    CaveVines,
+    Chain,
+    ChemistryTable,
+    ChestAndTrappedChest,
+    EnderChest,
+    ChiseledBookshelf,
+    ChorusFlower,
+    ChorusPlant,
+    Cocoa,
+    CommandBlocks,
+    Composter,
+    Concrete,
+    ConcretePowder,
+    Conduit,
+    Coral,
+    CoralBlock,
+    CoralFan,
+    DaylightDetector,
+    //Missing from the overview:
+    DecoratedPot,
+    Deepslate,
+    //Bedrock Only: Dirt,
+    DispenserAndDropper,
+    Doors,
+    EndPortalFrame,
+    EndRod,
+    Farmland,
+    Fences,
+    FenceGates,
+    Fire,
+    Flowers,
+    FlowerPot,
+    Froglight,
+    FrostedIce,
+    Furnace,
+    //Bedrock Only: Glass,
+    GlassPanes,
+    GlazedTerracotta,
+    GlowLichen,
+    GrassBlockMyceliumAndPodzol,
+    Grindstone,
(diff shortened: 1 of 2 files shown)


crates/valence_server/src/placement.rs
```diff
@@ -0,0 +1,895 @@
...
+        let interact_block = layer.block(interact_pos).unwrap_or_default();
+
+        let facing_pos = interact_pos.get_in_direction(face);
+        let facing_block = layer.block(facing_pos).unwrap_or_default();
+
+        let mut block = block_kind.to_state();
+        let placement_rule = PlacementRule::from_kind(block_kind);
+
+        let is_facing_block_replaceable: bool = facing_block.state.is_replaceable();
+        let has_hitbox = block.collision_shapes().next().is_some();
```
XBagon, line 723: a

crates/valence_server/src/placement.rs
```diff
@@ -0,0 +1,895 @@
...
+use valence_entity::{hitbox::Hitbox, living::LivingEntity, Look};
+use valence_generated::block::{PropName, PropValue};
+use valence_math::{Aabb, DVec3};
+use valence_protocol::{BlockKind, BlockPos, BlockState, Direction};
+
+use crate::{interact_block::InteractBlockEvent, ChunkLayer};
+
+//from https://minecraft.fandom.com/wiki/Block_states
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub enum PlacementRule {
```
XBagon, line 15: a

crates/valence_server/src/placement.rs
```diff
@@ -0,0 +1,939 @@
...
+            Self::Left => Self::Right,
+            Self::Right => Self::Left,
+        }
+    }
+}
+
+impl PlacementRule {
+    //TODO: WIP
+    pub const fn from_kind(kind: BlockKind) -> Self {
+        match kind {
```
rj00a, line 262: Unfortunately this approach isn't really maintainable. I don't want to manually update this list every time a new Minecraft version is released.

I think we should investigate how Vanilla is mapping the blocks to their placement rules and extract that information if possible. If not, then perhaps we could use some heuristic based on properties, but this is not my desired solution.

XBagon: I don't think there's much room for alternatives, if we want to mimic the client behaviour. Minecraft seems to hardcode these rules for these sets of blocks as well. As you said these rules are very complicated and I don't see a good way around this. Having a good default, recognizing similarities and thus being able to deduplicate code are things I try to do and hopefully can limit the maintanence cost. Also adding new blocks to these rules shouldn't be too much work.

Only alternative I can think of is dropping this in the scope of Valence entirely and make a crate for vanilla behaviour, which is allowed to be behind Valence and the newest Minecraft update a bit.

dyc3: Yeah minecraft implements placement behavior for each block by implementing this function:

```java
public BlockState getPlacementState(ItemPlacementContext ctx)
```

For example, this is the impl in `EnderChestBlock`:

```java
public BlockState getPlacementState(ItemPlacementContext ctx) {
        FluidState fluidState = ctx.getWorld().getFluidState(ctx.getBlockPos());
        return (BlockState)((BlockState)this.getDefaultState().with(FACING, ctx.getHorizontalPlayerFacing().getOpposite())).with(WATERLOGGED, fluidState.getFluid() == Fluids.WATER);
    }
```

It might be feasible to have some codegen that reads the `with()` calls and derives the correct placement rule for the block. eg, if `with(FACING, ...)` is present, the block probably should face the player.

Bafbi: so extracting properties ? cause we already extract them, but yea I don't see placement rules that couldn't be derive from some properties.

crates/valence_server/src/placement.rs
```diff
@@ -0,0 +1,939 @@
...
+                },
+            look,
+            ignore_collisions,
+        } = self;
+
+        let mut system_state =
+            SystemState::<(Query<&mut ChunkLayer>, Query<&Hitbox, With<LivingEntity>>)>::new(world);
+        let (mut layers, entity_hitboxes) = system_state.get_mut(world);
+
+        let mut layer = layers.single_mut();
```
Bafbi, line 728: You can't do that, you either need to get the layer the client is on or ask for a layer to place the block

XBagon: I'm not sure what you mean, this is basically just copied from the `building` example.

dyc3: You can't call `layers.single_mut()` because it'll panic if there are more than one entity with a `ChunkLayer`. It's there in the example code for simplicity.

XBagon: ah yea I had thought for a moment, but though the example will have a reason.. and yea it has a reason.
Thanks for pointing it out :D

PR #516, opened 2023-09-01
Add ability to write anvil chunks

Earthcomputer:
# Objective

- Follow on from #510 by adding the ability to write and delete chunks in anvil files.

# Solution

- Add a `write` feature to `valence_anvil`, behind which the region file allocator is hidden (this is a bit more work to keep track of).
- Users can change the compression method and whether to write oversized chunks, in case they want to write anvil files compatible with older versions of Minecraft.
- `LruCache::try_get_or_insert_mut` is something I have [pull requested](https://github.com/jeromefroe/lru-rs/pull/178) to `lru`, we can switch to that once it's merged.

### DIFF
--- a/crates/valence_anvil/src/lib.rs
+++ b/crates/valence_anvil/src/lib.rs
@@ -25,6 +25,7 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 #[cfg(feature = "bevy_plugin")]
 pub use bevy::*;
+use bitfield_struct::bitfield;
 use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use flate2::bufread::{GzDecoder, ZlibDecoder};
 use flate2::write::{GzEncoder, ZlibEncoder};
@@ -86,11 +87,19 @@ impl Compression {
     }
 }
 
-#[cfg(feature = "write")]
 #[derive(Copy, Clone, Debug, Default)]
-struct WriteOptions {
-    compression: Compression,
-    skip_oversized_chunks: bool,
+#[non_exhaustive]
+pub struct WriteOptions {
+    /// Set the compression method used to write chunks. This can be useful to
+    /// change in order to write anvil files compatible with older Minecraft
+    /// versions.
+    pub compression: Compression,
+
+    /// Set whether to skip writing oversized chunks (>1MiB after compression).
+    /// Versions older than 1.15 (19w36a) cannot read oversized chunks, so this
+    /// may be useful for writing region files compatible with those
+    /// versions.
+    pub skip_oversized_chunks: bool,
 }
 
 #[derive(Debug)]
@@ -98,12 +107,12 @@ pub struct RegionFolder {
     /// Region files. An LRU cache is used to limit the number of open file
     /// handles.
     regions: LruCache<RegionPos, RegionEntry>,
-    /// Path to the "region" subdirectory in the world root.
+    /// Path to the directory containing the region files and chunk files.
     region_root: PathBuf,
     /// Scratch buffer for (de)compression.
     compression_buf: Vec<u8>,
-    #[cfg(feature = "write")]
-    write_options: WriteOptions,
+    /// Options to use for writing the chunk.
+    pub write_options: WriteOptions,
 }
 
 impl RegionFolder {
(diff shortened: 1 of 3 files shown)


crates/valence_anvil/src/lib.rs
```diff
@@ -64,18 +100,78 @@ pub struct RegionFolder {
...
+
+    fn get_region<'a>(
+        regions: &'a mut LruCache<RegionPos, RegionEntry>,
+        region_root: &Path,
+        region_x: i32,
+        region_z: i32,
+    ) -> Result<Option<&'a mut Region>, RegionError> {
+        // Need to double get the entry from the cache to make the borrow checker happy.
+        // Polonius will fix this eventually.
+        if regions.get_mut(&(region_x, region_z)).is_some() {
```
MalbaCato, line 145: am I hallucinating, or does this `if` achieve nothing?
(assuming the `None` arm get changes to a `()` value)

also, why `get_mut`?

Earthcomputer: It achieves making the borrow checker happy.

crates/valence_anvil/src/lib.rs
```diff
@@ -86,51 +182,275 @@ impl RegionFolder {
...
+    }
+
+    /// Sets the raw chunk at the given position, overwriting the old chunk if
+    /// it exists.
+    #[cfg(feature = "write")]
+    pub fn set_chunk(
+        &mut self,
+        pos_x: i32,
+        pos_z: i32,
+        chunk: Compound,
```
rj00a, line 222:
```suggestion
        chunk: &Compound,
```

crates/valence_anvil/src/lib.rs
```diff
@@ -86,51 +182,275 @@ impl RegionFolder {
...
+            let mut split = file_name.splitn(4, '.');
+            if split.next() != Some("r") {
+                return Ok(None);
+            }
+            let Some(Ok(x)) = split.next().map(|x| x.parse()) else {
+                return Ok(None);
+            };
+            let Some(Ok(z)) = split.next().map(|x| x.parse()) else {
+                return Ok(None);
+            };
```
rj00a, line 297:
```suggestion
            let Some(Ok(x)) = split.next().map(str::parse()) else {
                return Ok(None);
            };
            let Some(Ok(z)) = split.next().map(str::parse()) else {
                return Ok(None);
            };
```

crates/valence_anvil/src/lib.rs
```diff
@@ -86,51 +182,275 @@ impl RegionFolder {
...
+            let file = file?;
+
+            if !file.file_type()?.is_file() {
+                return Ok(None);
+            }
+
+            let file_name = match file.file_name().into_string() {
+                Ok(file_name) => file_name,
+                Err(_) => return Err(RegionError::OsStringConv),
+            };
```
rj00a, line 285: Can be written as a let-else

crates/valence_anvil/src/lib.rs
```diff
@@ -86,51 +182,275 @@ impl RegionFolder {
...
+
+    /// Deletes the chunk at the given chunk position, returning whether the
+    /// chunk existed before it was deleted.
+    ///
+    /// Note that this only marks the chunk as deleted so that it cannot be
+    /// retrieved, and can be overwritten by other chunks later. It does not
+    /// decrease the size of the region file.
+    ///
+    /// This method is simple enough to not require the `write` feature.
+    pub fn delete_chunk(&mut self, pos_x: i32, pos_z: i32) -> Result<bool, RegionError> {
```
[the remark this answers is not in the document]
Earthcomputer: I think it's more appropriate for an optimize method, especially since I'm not reducing the size of the file if there are any chunks left. Could leave that to another PR though.

crates/valence_anvil/src/lib.rs
```diff
@@ -185,32 +518,250 @@ impl RegionFolder {
...
+        (location_bytes & 0xff) as usize
+    }
+
+    fn is_external_stream_chunk(stream_version: u8) -> bool {
+        (stream_version & 0x80) != 0
+    }
+
+    fn get_external_chunk_version(stream_version: u8) -> u8 {
+        stream_version & !0x80
+    }
```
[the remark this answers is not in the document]
Earthcomputer: Since these are functions only used by `Region` I prefer them inside `Region`, but if you feel strongly about it I can move them outside.

rj00a (approved):

PR #518, opened 2023-09-04
Better position types

rj00a:
- Add `ChunkSectionPos` and `BiomePos` types.
- `From` conversions between position types.
- Packed block position and chunk section position types.
- Use `bitfield-struct` in more places.


crates/valence_server/src/layer/chunk/loaded.rs
```diff
@@ -442,26 +444,39 @@ impl Chunk for LoadedChunk {
...
+                                    ChunkDeltaUpdateEntry::new()
+                                        .with_off_x(x as u8)
+                                        .with_off_y(y as u8)
+                                        .with_off_z(z as u8)
+                                        .with_block_state(block.to_raw().into()),
+                                );
+                            }
                         }
                     }
                 }
```
rj00a, line 482: probably bugged but I'm going to remove it anyways.

Issue #104, opened 2022-10-07
Implement MOTION_BLOCKING heightmap.


rj00a:
If I'm not mistaken, rain uses this heightmap to determine when it should stop falling. Once weather is implemented, the heightmap can be properly tested.

- [x] #105

dyc3:
According to this: https://minecraft.fandom.com/wiki/Chunk_format

> MOTION_BLOCKING: The highest block that blocks motion or contains a fluid.

tim-kt:
Implemented by #523, can be closed.

PR #523, opened 2023-09-05
`MOTION_BLOCKING` heightmap

tim-kt:
# Objective
Aims to implement #104. This is a draft, I don't necessarily need a review yet.

# Solution
The server needs to generate a heightmap when chunks are generated, update them when blocks are changed, and send it in the `ChunkDataS2c` packet.

I think the implementation should be done in the following order:
- [x] Encode (constant) heightmap and send it in the `ChunkDataS2c` packet
- [x] Build heightmap in `write_init_packets`
- [x] Find out which blocks are "motion-blocking"
- [x] Implement a `blocks_motion` function
- [x] Change `weather.rs` to test the functionality`WORLD_SURFACE`
- [x] Find out why one needs to add 2 to the heightmap


dyc3:
I think it would be good to update the weather example to show this functionality.

rj00a (commented):
The way I think we should handle this is by keeping the heightmaps completely contained in the `write_init_packets` function. We can rebuild the heightmaps there from scratch whenever the `ChunkDataS2c` packet cache needs to be created. This would avoid all the complexity involved in updating the heightmap whenever a block is modified.

We should also figure out which blocks exactly are considered "motion blocking" by the game and extract that information. And also figure out what `WORLD_SURFACE` is used for.

tim-kt:
It seems like "motion-blocking" blocks are more or less solid and liquid blocks, waterlogged blocks included. One way to check this would be to examine the packets of the vanilla server in `minecraft:debug` and from this, generate a mapping from block state to motion-blocking or not. Has something like this been done before for Valence?

rj00a:
You'll want to take a look at the `extracted/` folder. That contains the fabric mod we are using to get information from the game.

So it seems like the main thing we're missing is an `is_solid` method for `BlockState`.

crates/valence_server/src/layer/chunk/loaded.rs
```diff
@@ -295,6 +293,72 @@ impl LoadedChunk {
...
 
+    /// Generates the `MOTION_BLOCKING` heightmap for this chunk, which stores
+    /// the height of the highest non motion-blocking block in each column.
+    ///
+    /// The lowest value of the heightmap is 0, which means that there are no
+    /// motion-blocking blocks in the column. In this case, rain will fall
+    /// through the void and there will be no rain particles.
+    ///
+    /// A value of 1 means that rain particles will appear at the lowest
+    /// possible height y=-64, but note that blocks cannot be placed at y=-65.
```
rj00a, line 304: Worth pointing out that -64 is not a hardcoded value. It's known as `min_y` in the "dimension type" and can be configured to go lower or higher.

crates/valence_server/src/layer/chunk/loaded.rs
```diff
@@ -295,6 +293,72 @@ impl LoadedChunk {
...
+    /// For example, the `WORLD_SURFACE` heightmap in an empty superflat world
+    /// is always 4. The first 36 long values will then be
+    ///
+    /// 0 000000100 000000100 000000100 000000100 000000100 000000100 000000100,
+    ///
+    /// and the last long will be
+    ///
+    /// 0 000000000 000000000 000000000 000000100 000000100 000000100 000000100.
+    fn encode_heightmap(&self, heightmap: Vec<Vec<u32>>) -> Value {
+        let mut encoded: Vec<i64> = vec![0; 37];
```
rj00a, line 347: Would be nice if the magic numbers here were computed from local constants to make their derivation more obvious. Specifically, `37`, `7`, `9`.

crates/valence_server/src/layer/chunk/loaded.rs
```diff
@@ -295,6 +293,72 @@ impl LoadedChunk {
...
+    /// `self.block_state(x, 0, z)` corresponds to the block at (x, -64, z)
+    /// ingame.
+    #[allow(clippy::needless_range_loop)]
+    fn motion_blocking(&self) -> Vec<Vec<u32>> {
+        let mut heightmap: Vec<Vec<u32>> = vec![vec![0; 16]; 16];
+
+        for z in 0..16 {
+            for x in 0..16 {
+                for y in (0..self.height()).rev() {
+                    if self.block_state(x as u32, y, z as u32).blocks_motion() {
```
rj00a, line 317: I believe this also needs to check if the block state is a fluid.

```java
MOTION_BLOCKING("MOTION_BLOCKING", Purpose.CLIENT, state -> state.blocksMovement() || !state.getFluidState().isEmpty()),
```

rj00a (changes requested):

crates/valence_server/src/layer/chunk/loaded.rs
tim-kt: Would it make sense to create a constant such as `CHUNK_LENGTH` for the value 16 or is that value ubiquitous enough?

rj00a: I think it's clear enough what `16` means, but I'm not opposed to it if the code were changed to use it everywhere.

crates/valence_server/src/layer/chunk/loaded.rs
```diff
@@ -292,6 +291,86 @@ impl LoadedChunk {
...
+    ///
+    /// 0 000000000 000000000 000000000 000000100 000000100 000000100 000000100.
+    fn encode_heightmap(heightmap: Vec<Vec<u32>>) -> Value {
+        const BITS_PER_ENTRY: u32 = 9;
+        const ENTRIES_PER_LONG: u32 = i64::BITS / BITS_PER_ENTRY;
+
+        // Unless `ENTRIES_PER_LONG` is a power of 2 and therefore evenly divides 16*16,
+        // we need to add one extra long to fit all values in the packet.
+        const LONGS_PER_PACKET: u32 =
+            16 * 16 / ENTRIES_PER_LONG + !ENTRIES_PER_LONG.is_power_of_two() as u32;
```
rj00a, line 357: It would be more precise to write this as
```suggestion
            16 * 16 / ENTRIES_PER_LONG + (ENTRIES_PER_LONG % (16 * 16) != 0) as u32;
```

What we really want is ceiling division, but unfortunately `div_ceil` is unstable.

rj00a (approved):

PR #525, opened 2023-09-08
Add HeldItem::set_slot

AviiNL:
# Objective

This PR adds functionality to allow the server to change the users selected hotbar slot using `HeldItem::set_slot(36..44)`

### DIFF
--- a/crates/valence_inventory/src/lib.rs
+++ b/crates/valence_inventory/src/lib.rs
@@ -35,6 +35,7 @@ pub use valence_server::protocol::packets::play::player_action_c2s::PlayerAction
 use valence_server::protocol::packets::play::{
     ClickSlotC2s, CloseHandledScreenC2s, CloseScreenS2c, CreativeInventoryActionC2s, InventoryS2c,
     OpenScreenS2c, PlayerActionC2s, ScreenHandlerSlotUpdateS2c, UpdateSelectedSlotC2s,
+    UpdateSelectedSlotS2c,
 };
 use valence_server::protocol::{VarInt, WritePacket};
 use valence_server::text::IntoText;
@@ -674,14 +675,14 @@ fn update_player_inventories(
     }
 }
 
+/// Handles the `HeldItem` component being changed on a client, which
+/// indicates that the server has changed the selected hotbar slot.
 fn update_player_selected_slot(mut clients: Query<(&mut Client, &mut HeldItem)>) {
     for (mut client, mut held_item) in &mut clients {
         if held_item.changed {
-            client.write_packet(
-                &valence_server::protocol::packets::play::UpdateSelectedSlotS2c {
-                    slot: (held_item.held_item_slot - PLAYER_INVENTORY_MAIN_SLOTS_COUNT) as u8,
-                },
-            );
+            client.write_packet(&UpdateSelectedSlotS2c {
+                slot: (held_item.held_item_slot - PLAYER_INVENTORY_MAIN_SLOTS_COUNT) as u8,
+            });
 
             held_item.changed = false;
         }
--- a/crates/valence_inventory/src/lib.rs
+++ b/crates/valence_inventory/src/lib.rs
@@ -399,8 +399,6 @@ impl ClientInventoryState {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Component, Deref)]
 pub struct HeldItem {
     held_item_slot: u16,
-    #[deref(ignore)]
-    changed: bool,
 }
 
 impl HeldItem {
@@ -418,7 +416,6 @@ impl HeldItem {
         );
 
         self.held_item_slot = slot;
-        self.changed = true;
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/valence_inventory/src/lib.rs
```diff
@@ -405,6 +409,17 @@ impl HeldItem {
     pub fn slot(&self) -> u16 {
         self.held_item_slot
     }
+
+    pub fn set_slot(&mut self, slot: u16) {
+        // temp
+        assert!(
+            (36..=44).contains(&slot),
```
Bafbi, line 416: The packet seems to be using 0..8 https://wiki.vg/Protocol#Set_Held_Item

Bafbi (changes requested):
The pattern you'r using made use of a second change detection variable, for the fact that you don't want to send the packet when it's the client that update it I suppose. 
But I think it would be better to use the events system to solve this problem (like what was use for [ablilitites](https://github.com/valence-rs/valence/blob/1ddde71963ca3b15e8d97d1e4383b61431a30bbc/crates/valence_server/src/abilities.rs) )
I think it's a better pattern because
- events (when #487 will be usable) will permit the dev to cancel the client behaviour.
- keep the consistencies to not really use getter and setter for our component.

crates/valence_inventory/src/lib.rs
AviiNL: I am aware that the packet uses 0..8, the reason why i'm not using it here is to keep it consistent with the slot() return value, which is 36..44, and changing that is out of scope of this PR imo and backwards-compat breaking. (see line 684 where it gets converted)

AviiNL:
There are no events involved here, this isn't code that responds to some client request, how would I fire off an event from the set_slot fn without any knowledge of anything other than a number of the selected slot? This is a server-side-only thing where the server can force a client to have a specific hotbar slot selected.

crates/valence_inventory/src/lib.rs
```diff
@@ -397,6 +399,8 @@ impl ClientInventoryState {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Component, Deref)]
 pub struct HeldItem {
     held_item_slot: u16,
+    #[deref(ignore)]
+    changed: bool,
```
icrayix, line 403: Can't we use bevy's built-in change detection instead?
https://docs.rs/bevy/0.9.1/bevy/ecs/query/struct.Changed.html

Bafbi: no because when the client change slot, we trigger one.

Bafbi: And that's where I wanted to go with my review, what I proposing is that the system, who handle the packet bypass the change detection, so the only way the user of the api know that something occurs would be the event send.
now you would'nt need a seconde variable.

crates/valence_inventory/src/lib.rs
Bafbi: yea right, weird, but not the scope of the pr

crates/valence_inventory/src/lib.rs
AviiNL: I dont really understand what needs to be bypassed, as i mentioned, this isnt triggered from something automatically, this function gives a developer the option to change a clients selected slot index. Imo, this is simple enough to use as-is, and when https://github.com/valence-rs/valence/pull/487 gets finalized, this can be look at again if required?

Bafbi: I'm talking about the system that handle the c2s held_item packet, the things I don't like here is the fact that you need a double change detection

AviiNL: i havent touched anything related to c2s in this pr, out of scope?

crates/valence_inventory/src/lib.rs
```diff
@@ -659,6 +675,20 @@ fn update_player_inventories(
     }
 }
 
+/// Handles the `HeldItem` component being changed on a client, which
+/// indicates that the server has changed the selected hotbar slot.
+fn update_player_selected_slot(mut clients: Query<(&mut Client, &mut HeldItem)>) {
```
dyc3, line 680: I would rather use bevy change detection to detect when `HeldItem` is changed, and bypass the change detection when the c2s packet is received.

dyc3 (changes requested):

crates/valence_inventory/src/lib.rs
Bafbi: no because, how the c2s was implemented was making sense when you didn't care that the change detection was trigger, but now you care for your purpose.

crates/valence_inventory/src/lib.rs
Bafbi: how am I so bad I explaining my thought when it was this easy.

crates/valence_inventory/src/lib.rs
```diff
@@ -1216,18 +1228,31 @@ pub struct UpdateSelectedSlotEvent {
     pub slot: u8,
 }
 
+/// Handles the `HeldItem` component being changed on a client, which
```
dyc3, line 1231:
```suggestion
/// Handles the `HeldItem` component being changed on a client entity, which
```

dyc3 (approved):
Looks good, just a nit

dyc3 (approved):

PR #529, opened 2023-09-20
Solve inability for items with improper NBT data to be moved out of inventories

Mac898:
Hi,
# Objective

I was playing around with the chest example.
After adding an item with NBT data, I noticed I was unable to remove it from the chest.
Sourced the error down to this check. All I have done is not check the NBT data for now.

Same as in last comment of #292 .

# Solution

Only item kind and count checked, nbt ignored.

Kind Regards,
Mac

### DIFF
--- a/crates/valence_inventory/src/validate.rs
+++ b/crates/valence_inventory/src/validate.rs
@@ -167,8 +167,13 @@ pub(super) fn validate_click_slot_packet(
                 if should_swap {
                     // assert that a swap occurs
                     ensure!(
-                        *old_slot == packet.carried_item
-                            && cursor_item.0 == packet.slot_changes[0].stack,
+                        // TODO: Patch for moving items with incorrect NBT Data.
+                        // The client has added NBT data to the item which does not exist
+                        // on the item in the inventory.
+                        old_slot.item == packet.carried_item.item &&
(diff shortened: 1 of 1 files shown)


crates/valence_inventory/src/validate.rs
```diff
@@ -167,7 +167,11 @@ pub(super) fn validate_click_slot_packet(
                 if should_swap {
                     // assert that a swap occurs
                     ensure!(
-                        *old_slot == packet.carried_item
+                        // TODO: Patch for moving items with incorrect NBT Data.
+                        // The client has added NBT data to the item which does not exist
+                        // on the item in the inventory.
```
dyc3, line 172: nit

```suggestion
                        // There are some cases where the client will add NBT data that 
                        // did not previously exist.
```

dyc3 (changes requested):
This needs a unit test.

Mac898:
I've never written a unit test in rust, but I'll give it a go and try to follow the examples alrady in validate.rs :)

Mac898:
I have verified the unit test fails on the old version, and passes after the changes.

Assuming all else is well, should be complete now.

dyc3 (approved):

PR #543, opened 2023-10-04
Fix encoding villager level in `TrackedData`

### LINKED ISSUE #453: Incorrect entity data for spawned entity
## Valence Version

8759a2163210403821c64a9e18f248a44ad71034

## What You Did

From discord:

```rust
commands.spawn(VillagerEntityBundle {
    location: Location(inst_ent),
    position: Position::new([0.5, 104.0, 0.5]),
    ..Default::default()
});

commands.spawn(VillagerEntityBundle {
    location: Location(inst_ent),
    position: Position::new([0.5, 104.0, 3.0]),
    villager_villager_data: VillagerData({
        valence::entity::VillagerData {
            kind: valence::entity::VillagerKind::Jungle,
            profession: VillagerProfession::Armorer,
            level: 5,
        }
    }),
    ..Default::default()
});
```

## What Went Wrong

One of those villagers is on fire(!)

## Additional Information

Something with `TrackedData` or one of the systems that initializes the tracked data is probably bugged.

There is a separate issue where tracked data is initialized with incorrect default values from other entities, but this looks like a different problem.

guac42:
Fix encoding villager level in `TrackedData`. The level should be encoded as a `VarInt` when it was being encoded as an `i32`.

# Objective

- Fixes #453 

# Solution

- The level should be encoded as a `VarInt` when it was being encoded as an `i32`.

### DIFF
--- a/crates/valence_entity/src/lib.rs
+++ b/crates/valence_entity/src/lib.rs
@@ -382,7 +382,7 @@ impl EntityAnimations {
 #[derive(Component, Default, Debug, Deref, DerefMut)]
 pub struct ObjectData(pub i32);
 
-#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Encode, Decode)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
 pub struct VillagerData {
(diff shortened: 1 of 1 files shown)


rj00a:
Nice find :)

rj00a (approved):

PR #546, opened 2023-10-06
Generify `valence_nbt` to allow other types of String

Earthcomputer:
# Objective

- Generify `valence_nbt` to allow other types of `String`, such as `JavaString`.
- A side effect of this is that NBT using `Cow<str>` is now also possible.

# Solution

- `valence_nbt::Compound`, `valence_nbt::List` and `valence_nbt::Value` all now have a generic `S` parameter for the string type, which defaults to `String`.
- Extra `From` implementations for `Cow` strings have been added.
- Support for `JavaString` in `valence_nbt` has been added behind the `java_string` feature flag.
   - Added an `AsBorrowed` trait to query a compound, to allow `Compound<JavaString>` to be queried by `&str`.
- Added `FromModifiedUtf8` and `ToModifiedUtf8` traits to support decoding and encoding custom strings, respectively.
- `valence_anvil` has been adapted to allow custom string types in compounds in anvil files.
- The following breaking changes have been introduced:
   - `decode::from_binary` and `valence_anvil::RegionFolder::get_chunk` now have a generic parameter for the string. This can usually be inferred but not necessarily. Existing code that breaks should be changed to use `from_binary::<String>(...)` or `get_chunk::<String>(...)` respectively.

### DIFF
--- a/crates/valence_nbt/src/compound.rs
+++ b/crates/valence_nbt/src/compound.rs
@@ -26,7 +26,7 @@ impl<S: fmt::Debug> fmt::Debug for Compound<S> {
 
 impl<S> PartialEq for Compound<S>
 where
-    S: Eq + Ord + Hash,
+    S: Ord + Hash,
 {
     fn eq(&self, other: &Self) -> bool {
         self.map == other.map
@@ -36,7 +36,7 @@ where
 #[cfg(feature = "serde")]
 impl<Str> serde::Serialize for Compound<Str>
 where
-    Str: Eq + Ord + Hash + serde::Serialize,
+    Str: Ord + Hash + serde::Serialize,
 {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
@@ -49,7 +49,7 @@ where
 #[cfg(feature = "serde")]
 impl<'de, S> serde::Deserialize<'de> for Compound<S>
 where
-    S: Eq + Ord + Hash + serde::Deserialize<'de>,
+    S: Ord + Hash + serde::Deserialize<'de>,
 {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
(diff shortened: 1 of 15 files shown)


Earthcomputer:
Oh there are miri tests, I didn't mean to push those, they're meant to only run on `java_string`.

dyc3:
Would this allow for things like `Arc<str>` as well?

Earthcomputer:
In theory yes, although I haven't added any special support for that

crates/valence_nbt/src/value.rs
```diff
@@ -134,63 +135,82 @@ macro_rules! impl_value {
...
     }
 }
 
 impl_value!(Value,,(*),);
 impl_value!(ValueRef, 'a, (**), &'a);
 impl_value!(ValueMut, 'a, (**), &'a mut);
 
-impl Value {
+impl<S> Value<S> {
     /// Converts a reference to a value to a [ValueRef].
```
rj00a, line 212:
```suggestion
    /// Converts a reference to a value to a [`ValueRef`].
```

Earthcomputer: Looks like there's a lot of instances of this. Mind if I fix them all in bulk in a separate PR?

rj00a: Yeah that's fine

rj00a (approved):
I'm cool to merge this, but I'm starting to think the complexity involved with supporting both `String` and `JavaString` at the same time isn't really worth it. How would you feel if we reverted this in the future and simply replaced `String` with `JavaString`?

...although, that would probably close the door to bedrock edition NBT support, but idk what character encoding is used there.

crates/valence_nbt/src/compound.rs
```diff
@@ -232,50 +279,115 @@ impl Compound {
...
 }
 
-impl Extend<(String, Value)> for Compound {
+/// Trait that can be used as a key to query a compound. Basically something
+/// that can be converted to a type `B` such that `S: Borrow<B>`.
+pub trait AsBorrowed<S> {
+    type Borrowed: ?Sized;
+
+    fn as_borrowed(&self) -> &Self::Borrowed;
+}
```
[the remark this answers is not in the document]
Earthcomputer: Not only that but the maps don't support that trait either.

I can improve this a little bit once either negative trait impls or trait specialization is stabilized, then I at least wouldn't have to list every string type we expect to encounter. But until then I can't think of anything better.

rj00a:
Alright, probably not going to revert anything. Thanks for working on this.

PR #551, opened 2023-10-09
Added potion effects

### LINKED ISSUE #401: Potion Effects

There is no way to give entities potion effects.

A way to give potion effects to an entity and track them.

Sending the packets manually, but that didn't seem to work for whatever reason.

## Additional context

The `amplifier` field in `EntityStatusEffectS2c` is of type `u8` but it should be a signed byte.
[Packet info](https://wiki.vg/Protocol#Entity_Effect)

SelfMadeSystem:
# Objective

- Potion effects aren't supported by valence yet.
- You can send a packet to tell the client about a potion effect, but the server still has no idea what they are
<details>
<summary>Example code</summary>
```rs
pub fn add_potion_effect(mut clients: Query<&mut Client>, mut events: EventReader<SneakEvent>) {
    for event in events.iter() {
            if let Ok(mut client) = clients.get_mut(event.client) {
                client.write_packet(&EntityStatusEffectS2c {
                    entity_id: VarInt(0),
                    effect_id: VarInt(22),
                    amplifier: 0,
                    duration: VarInt(600),
                    flags: entity_status_effect_s2c::Flags::new()
                        .with_show_particles(true)
                        .with_show_icon(true),
                    factor_codec: None,
                });
            }
        }
    }
}```
</details>
- Closes #401
- Also, when the potion effect expires, we need to tell the client that their potion effect is no longer. Right now, with sending a packet, the effect doesn't get removed when it goes down to 00:00.

# Solution

I want to add the necessary components and stuff to facilitate potion effects.

> Note: I'm still somewhat new to rust and very new to bevy, so please lmk if I can improve anything or if I should do anything differently. Thanks!

### CODE REVIEW

File: extractor/src/main/java/rs/valence/extractor/extractors/Effects.java [hunk 1]
```diff
@@ -0,0 +1,55 @@
...
+public class Effects implements Main.Extractor {
+    public Effects() {
+    }
+
+    @Override
+    public String fileName() {
+        return "effects.json";
+    }
+
+    public static String toPascalCase(String str) {
```
dyc3, line 18: nit: doesn't need to be public

We should probably just put this in another class. I imagine a `Utils` or `ValenceUtils` class would be a good place for this.

crates/valence_generated/build/effect.rs
```diff
@@ -0,0 +1,218 @@
...
+
+            /// Gets the identifier of this effect.
+            pub const fn to_ident(self) -> Ident<&'static str> {
+                match self {
+                    #effect_to_ident_arms
+                }
+            }
+
+            /// Gets the translation key of this effect.
+            pub const fn translation_key(self) -> &'static str {
```
dyc3, line 187: `self` consumes the object `self` when this function is called. It should be `&self`. Same for `category`, `color`, etc.

dyc3 (commented):
Regarding your proposed API: This is something that should be a component, or set of components, that would go on the same entity as a player or any other entity. Minecraft entities other than players can have effects.

extractor/src/main/java/rs/valence/extractor/extractors/Effects.java
SelfMadeSystem: Alright. I moved it to a `ValenceUtils` class

crates/valence_generated/build/effect.rs
SelfMadeSystem: GitHub Copilot momento :skull:

JackCrumpLeys:
> > Regarding your proposed API: This is something that should be a component, or set of components, that would go on the same entity as a player or any other entity. Minecraft entities other than players can have effects. In other words, this shouldn't go on the `Client` component.
> 
> Thanks for the feedback! Seems like a good idea. I'm still not fully familiar with the ECS and I still need some time to adapt to it. I'll see how other stuff is implemented, and I'll see what I can do.

I would put this potion component on every Living Entity (https://valence.rs/rustdoc/valence/entity/living/struct.LivingEntity.html)

ethanwater (approved):
i recommend checking out https://github.com/WangTingZheng/mcp940/tree/master/bin/minecraft/net/minecraft/potion to see how minecraft 1.12 handled the potion system across entities ;)

crates/valence_entity/src/active_status_effects.rs
```diff
@@ -0,0 +1,148 @@
...
+    pub fn new_effects_mut(&mut self) -> &mut HashMap<StatusEffect, ActiveStatusEffect> {
+        &mut self.new_effects
+    }
+}
+
+/// Represents an active status effect.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+pub struct ActiveStatusEffect {
+    effect: StatusEffect,
+    amplifier: i8,
```
JackCrumpLeys, line 40: Why is this an i8? This field is a byte and so cannot be negative. We can also only go up to 127 not 255 with an i8.
```suggestion
    amplifier: u8,
```

crates/valence_entity/src/active_status_effects.rs
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    /// Returns the [`StatusEffect`] of the [`ActiveStatusEffect`].
+    pub fn status_effect(&self) -> StatusEffect {
+        self.effect
+    }
+
+    /// Returns the amplifier of the [`ActiveStatusEffect`].
+    pub fn amplifier(&self) -> i8 {
```
JackCrumpLeys, line 120: See previous comment
```suggestion
    pub fn amplifier(&self) -> u8 {
```

JackCrumpLeys, line 120: See previous comment
```suggestion
    pub fn amplifier(&self) -> u8 {
```

crates/valence_entity/src/active_status_effects.rs
```diff
@@ -0,0 +1,148 @@
...
+    pub fn ambient(&self) -> bool {
+        self.ambient
+    }
+
+    /// Returns whether the [`ActiveStatusEffect`] shows particles.
+    pub fn show_particles(&self) -> bool {
+        self.show_particles
+    }
+
+    /// Returns whether the [`ActiveStatusEffect`] shows an icon.
```
JackCrumpLeys, line 139: NIT: For all these comments I prefer the format of: Returns [true] if the [thing] has [something]. (like how you did it with `expired`)

crates/valence_protocol/src/packets/play/entity_status_effect_s2c.rs
```diff
@@ -7,7 +7,7 @@ use crate::{Decode, Encode, Packet, VarInt};
 pub struct EntityStatusEffectS2c {
     pub entity_id: VarInt,
     pub effect_id: VarInt, // TODO: effect ID registry.
-    pub amplifier: u8,
+    pub amplifier: i8,
```
JackCrumpLeys, line 10: See previous concerns.

crates/valence_server/src/status_effect.rs
```diff
@@ -0,0 +1,145 @@
...
+             * - poison
+             * - wither
+             */
+        }
+    }
+}
+
+fn create_packet(effect: &ActiveStatusEffect) -> EntityStatusEffectS2c {
+    EntityStatusEffectS2c {
+        entity_id: VarInt(0),
```
JackCrumpLeys, line 53: Why are you not taking entity id as a parameter? Does zero mean self? if I understand correctly every client in the same entity layer should receive the status effects of all entities in that layer. I understand if this is WIP disregard me.

crates/valence_entity/src/active_status_effects.rs
```diff
@@ -0,0 +1,148 @@
+use std::collections::HashMap;
+
+use bevy_ecs::prelude::*;
+
+use crate::status_effects::StatusEffect;
+
+/// [`Component`] that stores the [`ActiveStatusEffect`]s of an [`Entity`].
+#[derive(Component, Default, Debug)]
+pub struct ActiveStatusEffects {
+    active_effects: HashMap<StatusEffect, ActiveStatusEffect>,
```
JackCrumpLeys, line 10: I don't see why this is a hash-map as you don't randomly access based on `StatusEffect`. You only iterate it (something witch a hash-map is slow at). I would recommend a `Vec<(StatusEffect, ActiveStatusEffect)>` or maybe a `BTreeMap`.

crates/valence_entity/src/active_status_effects.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    pub fn new_effects(&self) -> &HashMap<StatusEffect, ActiveStatusEffect> {
+        &self.new_effects
+    }
+
+    pub fn new_effects_mut(&mut self) -> &mut HashMap<StatusEffect, ActiveStatusEffect> {
+        &mut self.new_effects
+    }
+}
```
JackCrumpLeys, line 34: This is what I would recommend using a Vec:
```suggestion
impl ActiveStatusEffects {
    pub fn add(&mut self, effect: ActiveStatusEffect) {
        if self.new_effects.iter().chain(self.active_effects.iter()).any(|(e, _)| e == &effect.status_effect()) {
            return;
        } // In a vector less then 100 elements, this is fast enough to justify not using a hashmap
    }

    pub fn iter(&self) -> impl Iterator<Item = &ActiveStatusEffect> {
        self.active_effects.iter().map(|(_, effect)| effect)
    }

    pub fn iter_mut(&mut self) -> impl Iterator<Item = &mut ActiveStatusEffect> {
        self.active_effects.iter_mut().map(|(_, effect)| effect)
    }

    pub fn new_effects(&self) -> impl Iterator<Item = &ActiveStatusEffect> {
        self.new_effects.iter().map(|(_, effect)| effect)
    }

    pub fn move_new_to_active(&mut self) -> impl Iterator<Item = &mut ActiveStatusEffect> {
        self.active_effects.extend( self.new_effects.drain(..));
        
        self.active_effects.iter_mut().map(|(_, effect)| effect)
    }
}
```

crates/valence_server/src/status_effect.rs
```diff
@@ -0,0 +1,145 @@
...
+             * - haste
+             * - mining fatigue
+             * - strength
+             * - weakness
+             * - luck
+             * - unluck
+             */
+        }
+
+        let old_map = std::mem::take(active_status_effects.new_effects_mut());
```
JackCrumpLeys, line 99: This is inefficient as hash-maps are costly to create and making a new one every tick is not ideal. look at my suggestion using a Vec.

JackCrumpLeys (commented):
I had a look at your code. it looks pretty good. 
You will want to get all Effects in a entity layer and send it to all clients viewing that layer.

crates/valence_entity/src/active_status_effects.rs
JackCrumpLeys: oops made a mistake with the last one, use this:
```rs
pub fn move_new_to_active(&mut self) -> impl Iterator<Item = &mut ActiveStatusEffect> {
        let old_len = self.active_effects.len();
        self.active_effects.extend( self.new_effects.drain(..));

        self.active_effects[old_len..].iter_mut().map(|(_, effect)| effect)
    }
    ```

File: crates/valence_entity/src/active_status_effects.rs [hunk 11]
```diff
@@ -0,0 +1,148 @@
...
+
+/// Represents an active status effect.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+pub struct ActiveStatusEffect {
+    effect: StatusEffect,
+    amplifier: i8,
+    /// The total duration of the status effect in ticks.
+    duration: i32,
```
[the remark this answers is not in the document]
SelfMadeSystem: My reasoning is that Minecraft uses -1 as infinity, but I guess we're not Java.

crates/valence_entity/src/active_status_effects.rs
SelfMadeSystem: Well Java doesn't have unsigned bytes. Afaict, the packet also doesn't use an unsigned byte. If you really believe it should be unsigned, I'll change it back

crates/valence_server/src/status_effect.rs
SelfMadeSystem: In other parts of the code across Valence, the ID of the player is 0. The server always seems to tell the client that it's 0. That's just what I'm following. If I make it be the entity id, then the client doesn't seem to think that it has any effect and thus nothing happens.

crates/valence_entity/src/active_status_effects.rs
SelfMadeSystem: I'm pretty certain I want some sort of a map just because I **don't** want any duplicates.

JackCrumpLeys: A vec would suit it as we can only have like 30 different effects and checking is a vec contains less then 50 is basicly costless. Infact I would remove the useless data and just store some kind of numerical id to check againsed . a Vec is the fastest data structure to iterate.

dyc3: Will the client display duplicate effects? eg could you have 2 speed effects

SelfMadeSystem: no that's impossible. the client will just replace the old effect

dyc3: It's more important that we don't have inconsistencies with client state. Stick with a map.

SelfMadeSystem: @dyc3 Could you look at the new code? I can't see how it can have inconsistencies with the client state.

crates/valence_entity/src/active_status_effects.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// ## Note
+    ///
+    /// It actually adds the [`ActiveStatusEffect`] to the new effects. The
+    /// [`ActiveStatusEffect`] will be added to the active effects in the next
+    /// tick (more specifically, during the upcoming [`EventLoopPostUpdate`]).
+    /// If the [`ActiveStatusEffect`] is already in the new effects, it will be
+    /// replaced.
+    ///
+    /// [`EventLoopPostUpdate`]: https://docs.rs/valence_server/0.2.0-alpha.1+mc.1.20.1/valence_server/event_loop/struct.EventLoopPostUpdate.html
+    pub fn add(&mut self, effect: ActiveStatusEffect) {
```
JackCrumpLeys, line 24: why not check in `active_effects` and `new_effects` here?

crates/valence_entity/src/active_status_effects.rs
```diff
@@ -0,0 +1,222 @@
...
+    ///
+    /// If you want to add, remove or modify [`ActiveStatusEffect`]s, use the
+    /// [`add`](ActiveStatusEffects::add) and
+    /// [`remove`](ActiveStatusEffects::remove) methods instead of directly
+    /// modifying the [`ActiveStatusEffect`]s.
+    pub fn active_effects_mut(&mut self) -> &mut Vec<ActiveStatusEffect> {
+        &mut self.active_effects
+    }
+
+    fn remove_new_from_active(&mut self) {
```
JackCrumpLeys, line 73: This function is not needed if you take my other suggestion.

crates/valence_entity/src/active_status_effects.rs
SelfMadeSystem: wdym?

Issue #557, opened 2023-10-12
Absorption


SelfMadeSystem:
## Describe the problem related to your feature request.

Absorption for the **player** is already somewhat implemented (see [`AbsorptionAmount`](https://valence.rs/rustdoc/valence/entity/player/struct.AbsorptionAmount.html)), but:
1. we need a way to damage the player w/ absorption (see #556 for damaging)
2. ALL living entities should have absorption, not just the player

## What solution would you like?

A full and proper implementation of absorption. Damaging the absorption should be properly dealt with by #556

## What alternative(s) have you considered?

Just continuing using [`AbsorptionAmount`](https://valence.rs/rustdoc/valence/entity/player/struct.AbsorptionAmount.html), but that only works for players.

## Additional context

Required for properly damaging entities with absorption (see #556) and for potions (see #558)

See [`net.minecraft.entity.LivingEntity#absorptionAmount`](https://maven.fabricmc.net/docs/yarn-1.20.2+build.4/net/minecraft/entity/LivingEntity.html#absorptionAmount)

Issue #556, opened 2023-10-12
Damaging entities + death


SelfMadeSystem:
## Describe the problem related to your feature request.

We need a way to damage & heal entities. We can currently just set their health via [`valence_entity::living::Health`](https://valence.rs/rustdoc/valence/entity/living/struct.Health.html), but there are two problems with this:

If decrementing health:

1. This doesn't have a damage source (e.g. magic, wither, pvp ig idk, etc.)
2. The entity's health just drops, without any effect (sound effect, camera effect, etc.)
3. There's no death either. If the entity dies, they just stay there and if the client gets a death screen, they're soft locked because they can't respawn. 
4. Absorption?

If incrementing health:

1. We have no way of knowing the entity's max health yet. See #555 for `EntityAttributes`
2. Absorption?

## What solution would you like?

1. A way to damage entities
2. Damage sources
3. A way to heal entities

## What alternative(s) have you considered?

Just setting `Health`, but it has drawbacks as previously mentioned.

## Additional context

Required to implement instant health, instant damage, regeneration, poison and wither in #551

See [`net.minecraft.entity.damage`](https://maven.fabricmc.net/docs/yarn-1.20.2+build.4/net/minecraft/entity/damage/package-summary.html)

Issue #555, opened 2023-10-11
Add EntityAttributes


SelfMadeSystem:
## Describe the problem related to your feature request.

Certain stuff requires `EntityAttribute`s. For example, movement speed, max health, armor, etc. all require `EntityAttribute`s (+ the corresponding packet) to work.

## What solution would you like?

Implement `EntityAttribute` and `EntityAttributes`.

## What alternative(s) have you considered?

I don't think any is possible.

## Additional context

Required to implement speed, slowness, haste, etc. for potions (#558).

See [`net.minecraft.entity.attribute.EntityAttributes`](https://maven.fabricmc.net/docs/yarn-1.20.2+build.4/net/minecraft/entity/attribute/EntityAttributes.html) and https://minecraft.wiki/w/Attribute

PR #558, opened 2023-10-12
Potions / Status Effects

### LINKED ISSUE #401: Potion Effects

SelfMadeSystem:
# Objective

- Potion effects aren't supported by valence yet.
- You can send a packet to tell the client about a potion effect, but the server still has no idea what they are
<details>
<summary>Example code</summary>

```rs
pub fn add_potion_effect(mut clients: Query<&mut Client>, mut events: EventReader<SneakEvent>) {
    for event in events.iter() {
            if let Ok(mut client) = clients.get_mut(event.client) {
                client.write_packet(&EntityStatusEffectS2c {
                    entity_id: VarInt(0),
                    effect_id: VarInt(22),
                    amplifier: 0,
                    duration: VarInt(600),
                    flags: entity_status_effect_s2c::Flags::new()
                        .with_show_particles(true)
                        .with_show_icon(true),
                    factor_codec: None,
                });
            }
        }
    }
}
```
</details>

- Closes #401
- Also, when the potion effect expires, we need to tell the client that their potion effect is no longer. Right now, with sending a packet, the effect doesn't get removed when it goes down to 00:00.

# Solution

I want to add the necessary components and stuff to facilitate potion effects.

> Note: I'm still somewhat new to rust and very new to bevy, so please lmk if I can improve anything or if I should do anything differently. Thanks!

# To do:

- [x] Extractor
- [x] `ActiveStatusEffects` component to handle the actual status effect applied to the mc entity
- [x] Add `ActiveStatusEffects` component to all entities.
- [x] Make a plugin to handle potion effects
  - [x] Decrease tick count
  - [x] Remove effect (& tell client) when tick count is 0
  - [x] Make the effects do stuff
    - [x] Particles
    - [x] Tell client
- [x] Add tests
- [x] Add example
  - [x] Add examples on how to implement potion effects (speed, instant health, etc.)
- [x] Extract stuff to make it easier to implement potions. See #593 

# Playground

<details>
<summary>Current playground</summary>

```rs
use valence::entity::active_status_effects::{ActiveStatusEffect, ActiveStatusEffects};
use valence::entity::status_effects::StatusEffect;
use valence::log::LogPlugin;
use valence::prelude::*;

#[allow(unused_imports)]
use crate::extras::*;

const SPAWN_Y: i32 = 64;

        ..Default::default()
    })
    .add_systems(Startup, setup)
    .add_systems(EventLoopUpdate, add_potion_effect)
    .run();
}

fn setup(
    mut commands: Commands,
    server: Res<Server>,
    biomes: Res<BiomeRegistry>,
) {
    let mut layer = LayerBundle::new(ident!("overworld"), &dimensions, &biomes, &server);

    for z in -5..5 {
        for x in -5..5 {
            layer.chunk.insert_chunk([x, z], UnloadedChunk::new());
        }
    }

    for z in -25..25 {
        for x in -25..25 {
            layer
                .chunk
                .set_block([x, SPAWN_Y, z], BlockState::GRASS_BLOCK);
        }
    }

    commands.spawn(layer);
}

fn init_clients(
    mut clients: Query<
        (
            &mut EntityLayerId,
            &mut VisibleChunkLayer,
            &mut VisibleEntityLayers,
            &mut Position,
            &mut GameMode,
        ),
        Added<Client>,
    >,
) {
    for (
        mut layer_id,
        mut visible_chunk_layer,
        mut visible_entity_layers,
        mut pos,
        mut game_mode,
    ) in &mut clients
    {
        let layer = layers.single();

        layer_id.0 = layer;
        *game_mode = GameMode::Survival;
    }
}

pub fn add_potion_effect(
    mut clients: Query<&mut ActiveStatusEffects>,
    mut events: EventReader<SneakEvent>,
) {
    for event in events.iter() {
            if let Ok(mut status) = clients.get_mut(event.client) {
                status.add(
                    ActiveStatusEffect::from_effect(StatusEffect::Wither)
                        .with_amplifier(2)
                        .with_duration(200),
                );
            }
        }
    }
}
```
</details>


dyc3:
This is getting rather large. Were you planning to handle #555 and #556 in separate PRs? That would make stuff easier to review.

SelfMadeSystem:
> This is getting rather large. Were you planning to handle #555 and #556 in separate PRs? That would make stuff easier to review.

yes, #555 and #556 (+#557) are very much out of the range of this PR. I'll probably also have to wait for those to be implemented before I can finish potions off.

SelfMadeSystem:
> Also, splitting the new extractor and supporting rust code into a new PR would help with size here.

oh okay I could try that

qualterz:
Is this feature finished? I want it to be merged into the main branch.

SelfMadeSystem:
> Is this feature finished? I want it to be merged into the main branch.

Yeah. Please review it, though, before merging it, just to make sure. It's been a while since I touched it so I don't know if it's outdated or anything

crates/valence_server/src/status_effect.rs
```diff
@@ -0,0 +1,152 @@
...
+    for mut active_status_effects in query.iter_mut() {
+        active_status_effects.increment_active_ticks();
+    }
+}
+
+fn create_packet(effect: &ActiveStatusEffect) -> EntityStatusEffectS2c {
+    EntityStatusEffectS2c {
+        // everywhere else in the codebase, this is the player's entity id
+        // will probably need to change this later
+        entity_id: VarInt(0),
```
dyc3, line 41: This needs to be addressed

crates/valence_server/src/status_effect.rs
```diff
@@ -0,0 +1,152 @@
...
+            .iter()
+            .any(|effect| effect.ambient());
+    }
+
+    if let Some(ref mut swirl_color) = swirl_color {
+        swirl_color.0 = get_color(active_status_effects);
+    }
+}
+
+/// Ctrl+C Ctrl+V from net.minecraft.potion.PotionUtil#getColor
```
dyc3, line 115: nit

```suggestion
/// Used to set the color of the swirls in the potion effect.
///
/// Equivalent to from net.minecraft.potion.PotionUtil#getColor
```

dyc3 (commented):
Partial review, haven't tried it in game yet. There's a merge conflict that also need to be addressed.

crates/valence_server/src/status_effect.rs
SelfMadeSystem: Out of the scope of this pr tho, kuz we'd have to change it everywhere

qualterz: I see that entity id of zero means client id as mentioned here: https://github.com/valence-rs/valence/blob/14cc9bdeb2b82e71e52882f65285c07e20e8bf6a/crates/valence_server/src/spawn.rs#L113

So maybe it's fine to leave zero here.

qualterz:
I've just tested on my machine, effect statuses seems to work fine.
I suppose that only several effects are client-sided and others should be implemented manually.
The only question is how these effects would be implemented by users, for example slowness effect.
My suggestion is to add a snippet of code in the example where one of server-sided effects is implemented.

examples/potions.rs
```diff
@@ -0,0 +1,104 @@
...
+        client.send_chat_message("Note: Most potion effects are not implemented yet.".into_text());
+    }
+}
+
+pub fn add_potion_effect(
+    mut clients: Query<&mut ActiveStatusEffects>,
+    mut events: EventReader<SneakEvent>,
+) {
+    let mut rng = rand::thread_rng();
+    for event in events.iter() {
```
qualterz, line 93: When running this example compiler outputs deprecated warning message of `iter()`.

```suggestion
    for event in events.read() {
```

But I'm not sure if it's a good way to fix the deprecation warning.

crates/valence_server/src/status_effect.rs
dyc3: Ah ok, then lets tweak the comments so that's more clear.

SelfMadeSystem:
> My suggestion is to add a snippet of code in the example where one of server-sided effects is implemented.

I'm trying to do that right now (admittedly, I should've done this earlier), but I now realise that I also need to slightly change how EntityAttributes work in valence (I made them too btw) to have feature parity with vanilla minecraft. I made this decision:
```rs
/// Right now, uses a random UUID for the modifier UUIDs.
///
/// Minecraft actually uses constant UUIDs. Uncertain if
/// I should do the same.
```
I'll make a new pr to change the decision to actually use UUIDs instead of arbitrary names because in hindsight, that's a pretty stupid decision lmao.

SelfMadeSystem:
kk I made #591

SelfMadeSystem:
ok I did a few things.

crates/valence_server/src/status_effect.rs
SelfMadeSystem: Changed it :+1:

examples/potions.rs
SelfMadeSystem: That's what it tells me to do. I'm just doing it

examples/potions.rs
```diff
@@ -0,0 +1,428 @@
...
+
+const SPEED_UUID: Uuid = parse_uuid_const(b"91AEAA56-376B-4498-935B-2F7F68070635");
+const SLOW_UUID: Uuid = parse_uuid_const(b"7107DE5E-7CE8-4030-940E-514C1F160890");
+const HASTE_UUID: Uuid = parse_uuid_const(b"AF8B6E3F-3328-4C0A-AA36-5BA2BB9DBEF3");
+const MINING_FATIGUE_UUID: Uuid = parse_uuid_const(b"55FCED67-E92A-486E-9800-B47F202C4386");
+const STRENGTH_UUID: Uuid = parse_uuid_const(b"648D7064-6A60-4F59-8ABE-C2C23A6DD7A9");
+const WEAKNESS_UUID: Uuid = parse_uuid_const(b"22653B89-116E-49DC-9B6B-9971489B5BE5");
+const HEALTH_BOOST_UUID: Uuid = parse_uuid_const(b"5D6F0BA2-1186-46AC-B896-C61C5CEE99CC");
+const LUCK_UUID: Uuid = parse_uuid_const(b"03C3C89D-7037-4B42-869F-B146BCB64D2E");
+const UNLUCK_UUID: Uuid = parse_uuid_const(b"CC5AF142-2BD2-4215-B636-2605AED11727");
```
qualterz, line 171: Is there a way to extract these constants from Minecraft and generate them in source code?

SelfMadeSystem: Now that I think about it, I could actually (+ a bit more that might make it easier to implement). I don't have the time to do that now, but I'll do that when I can

SelfMadeSystem: I'll make another PR for that I think (lmk if you think I shouldn't)

qualterz (approved):
Looks good enough.

qualterz (commented):
I think there's no need to implement effects with the same logic multiple times, here's glowing effect already implemented. It just reduces brevity of the code.

crates/valence_entity/src/active_status_effects.rs
```diff
@@ -0,0 +1,522 @@
...
+        }
+    }
+
+    /// **For internal use only.**
+    ///
+    /// Applies all the changes.
+    ///
+    /// Returns a [`IndexMap`] of [`StatusEffect`]s that were updated or removed
+    /// and their previous values.
+    pub fn apply_changes(&mut self) -> IndexMap<StatusEffect, Option<ActiveStatusEffect>> {
```
dyc3, line 251: Since this is intended to be an internal api:
```suggestion
    #[doc(hidden)]
    pub fn apply_changes(&mut self) -> IndexMap<StatusEffect, Option<ActiveStatusEffect>> {
```

dyc3 (commented):
Looks good, just a minor thing.

dyc3 (approved):

PR #559, opened 2023-10-12
Potions Extractor

SelfMadeSystem:
# Objective

We need a way to acquire (and thus extract) all potion/status effects + their attributes

Also, #558 was getting really large.

# Solution

I made an extractor.

Also, as per [the request](https://github.com/valence-rs/valence/pull/558#issuecomment-1760283675) of @dyc3, I split #558's extractor code into this pr.

### DIFF
--- a/crates/valence_generated/build/status_effects.rs
+++ b/crates/valence_generated/build/status_effects.rs
@@ -1,0 +1,227 @@
+use heck::ToPascalCase;
+use proc_macro2::TokenStream;
+use quote::quote;
+use serde::Deserialize;
+use valence_build_utils::{ident, rerun_if_changed};
+
+#[derive(Deserialize, Debug)]
+pub enum StatusEffectCategory {
+    Beneficial,
+    Harmful,
+    Neutral,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct StatusEffect {
+    id: u16,
+    name: String,
+    translation_key: String,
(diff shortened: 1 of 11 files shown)


crates/valence_generated/build/status_effects.rs
```diff
@@ -0,0 +1,227 @@
...
+            quote! {
+                Self::#name => #id,
+            }
+        })
+        .collect::<TokenStream>();
+
+    let effect_from_ident_arms = effects
+        .iter()
+        .map(|effect| {
+            // TODO: put the full resource identifier in the extracted JSON.
```
dyc3, line 59: Does this TODO need to be resolved in this PR?

dyc3 (commented):
Looks good.

crates/valence_generated/build/status_effects.rs
SelfMadeSystem: I copied the TODO from `sound.rs`

SelfMadeSystem: I would say no

dyc3: Ok, then lets remove that comment.

SelfMadeSystem: > Ok, then lets remove that comment.

alright

dyc3 (approved):

PR #561, opened 2023-10-13
Set brand

### LINKED ISSUE #560: Setting the server name (On F3 screen)
## Reproduction
I entered any playground and pressed F3. On the third line, it says "null" for the server brand/name.

This can however be solved quite easily:
<details>

The server name can be set i.E. to "Valence". 
```rust
fn init_clients(
    mut clients: Query<
        (
            &mut EntityLayerId,
            &mut VisibleChunkLayer,
            &mut VisibleEntityLayers,
            &mut Position,
            &mut GameMode,
            &mut Client,
        ),
        Added<Client>,
    >,
) {
    for (
        mut layer_id,
        mut visible_chunk_layer,
        mut visible_entity_layers,
        mut pos,
        mut game_mode,
        mut client
    ) in &mut clients
    {
        let layer = layers.single();

        layer_id.0 = layer;
        *game_mode = GameMode::Creative;

        // The payload is a &[u8] whose first byte is the length of the string 
        // followed by the string itself.
        let payload = [7, b'V', b'a', b'l', b'e', b'n', b'c', b'e'];

        client.send_custom_payload(ident!("minecraft:brand"), &payload);
    }
}
```

The `send_custom_payload` function with channel `"minecraft:brand"` can set the server name to any payload.

</details>

## Now what

The question is whether this should be set at all/ be turned into a plugin/ etc.

I'd be glad to turn it into a tiny plugin.

SeseMueller:
# Objective
Allows the setting of the brand/server name, visible on the F3 screen. Fixes #560 

# Solution
As recommended, uses an extension trait.

### DIFF
--- a/crates/valence_server/src/brand.rs
+++ b/crates/valence_server/src/brand.rs
@@ -1,0 +1,23 @@
+use valence_protocol::{WritePacket, ident, Bounded};
+use valence_protocol::packets::play::CustomPayloadS2c;
+
+pub trait SetBrand {
+    /// Sets the brand of the server.
+    ///
+    /// The Brand is displayed to the client in the F3 screen
+    /// and is often used to display the server name.
+    fn set_brand(&mut self, brand: &str);
+}
+
+impl<T: WritePacket> SetBrand for T {
+    fn set_brand(&mut self, brand: &str) {
+        // The data is a &[u8], where the first byte is the length of the string.
+        // It is also UTF-8 encoded, so any valid &str can be used.
+        let vec_data = [brand.len() as u8].iter().chain(brand.as_bytes().iter()).copied().collect::<Vec<_>>();
+        let data = vec_data.as_slice();
+        self.write_packet(&CustomPayloadS2c {
+            channel: ident!("minecraft:brand").into(),
+            data: Bounded(data.into()),
+        });
+    }
+}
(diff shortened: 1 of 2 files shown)


crates/valence_server/src/brand.rs
```diff
@@ -0,0 +1,28 @@
...
+        let vec_data = [brand.len() as u8]
+            .iter()
+            .chain(brand.as_bytes().iter())
+            .copied()
+            .collect::<Vec<_>>();
+        let data = vec_data.as_slice();
+        self.write_packet(&CustomPayloadS2c {
+            channel: ident!("minecraft:brand").into(),
+            data: Bounded(data.into()),
+        });
```
rj00a, line 26: The length is actually a `VarInt` and not a byte. I suggest doing this:

```suggestion
        let mut buf = vec![];

        let _ = VarInt(brand.len() as _).encode(&mut buf);
        buf.extend_from_slice(brand.as_bytes());

        self.write_packet(&CustomPayloadS2c {
            channel: ident!("minecraft:brand").into(),
            data: Bounded(buf.as_slice().into()),
        });
```

rj00a (changes requested):

SeseMueller:
Changed the implementation to the suggestion. In testing, it works exactly as it should, including all formatting.

rj00a (approved):

PR #562, opened 2023-10-13
Fixed Hex Viewer writing white text on white background when in light mode

SeseMueller:
# Objective
The Hex Viewer would write the raw data in white, on white background, when in light mode. This fixes that.

# Solution
The Color of the text is now not hardcoded anymore, but inferred.

### DIFF
--- a/tools/packet_inspector/src/app/hex_viewer.rs
+++ b/tools/packet_inspector/src/app/hex_viewer.rs
@@ -44,8 +44,13 @@ impl View for HexView {
                     }
 
                     ui.label(format!("{:08X}", count));
+                    let text_color = if ui.style().visuals.dark_mode {
+                        egui::Color32::from_gray(255)
+                    } else {
+                        egui::Color32::from_gray(0)
+                    };
                     for b in buf.iter().take(bytes_read) {
-                        ui.label(format!("{:02X}", b));
+                        ui.colored_label(text_color, format!("{:02X}", b));
                     }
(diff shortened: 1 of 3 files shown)


tools/packet_inspector/src/app/hex_viewer.rs
```diff
@@ -47,7 +45,7 @@ impl View for HexView {
 
                     ui.label(format!("{:08X}", count));
                     for b in buf.iter().take(bytes_read) {
-                        ui.colored_label(Color32::from_rgb(255, 255, 255), format!("{:02X}", b));
+                        ui.label(format!("{:02X}", b));
```
AviiNL, line 48:
```suggestion
                        if ui.style_mut().visuals.dark_mode {
                            ui.colored_label(
                                Color32::from_rgb(255, 255, 255),
                                format!("{:02X}", b),
                            );
                        } else {
                            ui.colored_label(Color32::from_rgb(0, 0, 0), format!("{:02X}", b));
                        }
```

the point of the white is so it pops out compared to the ascii representation, something like this will allow it to still do so

AviiNL (changes requested):

SeseMueller:
The latest commit makes the raw bytes in the hex viewer black or white depending on the background, changes the color of the selected packet in the packet list and adds a hint text to the filter bar to make it more obvious

AviiNL:
LGTM

rj00a (approved):

PR #563, opened 2023-10-13
Absorption

SelfMadeSystem:
# Problem

See #557

Will be necessary to implement #556 and #558

# Objective

Implement absorption for ALL living entities.

# Solution

Make a special case for LivingEntity for Absorption since it's untracked.

# Playground

<details>
<summary>Playground</summary>

```rs
use valence::entity::living::Absorption;
use valence::log::LogPlugin;
use valence::prelude::*;

#[allow(unused_imports)]
use crate::extras::*;

const SPAWN_Y: i32 = 64;

        ..Default::default()
    })
    .add_systems(Startup, setup)
    .add_systems(EventLoopUpdate, toggle_absorption_on_sneak)
    .run();
}

fn setup(
    mut commands: Commands,
    server: Res<Server>,
    biomes: Res<BiomeRegistry>,
) {
    let mut layer = LayerBundle::new(ident!("overworld"), &dimensions, &biomes, &server);

    for z in -5..5 {
        for x in -5..5 {
            layer.chunk.insert_chunk([x, z], UnloadedChunk::new());
        }
    }

    for z in -25..25 {
        for x in -25..25 {
            layer
                .chunk
                .set_block([x, SPAWN_Y, z], BlockState::GRASS_BLOCK);
        }
    }

    commands.spawn(layer);
}

fn init_clients(
    mut clients: Query<
        (
            &mut EntityLayerId,
            &mut VisibleChunkLayer,
            &mut VisibleEntityLayers,
            &mut Position,
            &mut GameMode,
        ),
        Added<Client>,
    >,
) {
    for (
        mut layer_id,
        mut visible_chunk_layer,
        mut visible_entity_layers,
        mut pos,
        mut game_mode,
    ) in &mut clients
    {
        let layer = layers.single();

        layer_id.0 = layer;
        *game_mode = GameMode::Survival;
    }
}

pub fn toggle_absorption_on_sneak(
    mut clients: Query<&mut Absorption>,
    mut events: EventReader<SneakEvent>,
) {
    for event in events.iter() {
            if let Ok(mut absorption) = clients.get_mut(event.client) {
                absorption.0 = if absorption.0 == 0.0 { 4.0 } else { 0.0 };
            }
        }
    }
}
```
</details>

### DIFF
--- a/crates/valence_entity/build.rs
+++ b/crates/valence_entity/build.rs
@@ -532,7 +532,7 @@ fn build() -> anyhow::Result<TokenStream> {
         fn update_living_and_player_absorption(
             mut query: Query<(&living::Absorption, &mut player::AbsorptionAmount), Changed<living::Absorption>>
         ) {
-            for (living_absorption, mut player_absorption) in &mut query {
+            for (living_absorption, mut player_absorption) in query.iter_mut() {
                 player_absorption.0 = living_absorption.0;
             }
         }
--- a/crates/valence_entity/build.rs
+++ b/crates/valence_entity/build.rs
@@ -358,6 +358,16 @@ fn build() -> anyhow::Result<TokenStream> {
                         bundle_init_fields.extend([quote! {
                             #snake_entity_name_ident: Default::default(),
                         }]);
+
+                        if entity_name == "LivingEntity" {
+                            bundle_fields.extend([quote! {
+                                pub living_absorption: super::living::Absorption,
+                            }]);
+
+                            bundle_init_fields.extend([quote! {
+                                living_absorption: Default::default(),
+                            }]);
+                        }
                     }
                     MarkerOrField::Field { entity_name, field } => {
                         let snake_field_name = field.name.to_snake_case();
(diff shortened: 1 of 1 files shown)


crates/valence_entity/build.rs
```diff
@@ -506,6 +524,22 @@ fn build() -> anyhow::Result<TokenStream> {
         }]);
     }
 
+    systems.extend([quote! {
+        #[allow(clippy::needless_borrow)]
+        #[allow(clippy::suspicious_else_formatting)]
```
dyc3, line 529: do we need to ignore these lints?

crates/valence_entity/build.rs
```diff
@@ -506,6 +524,22 @@ fn build() -> anyhow::Result<TokenStream> {
...
+        /// Special case for `living::Absorption`.
+        /// Updates the `AbsorptionAmount` component of the player entity.
+        fn update_living_and_player_absorption(
+            mut query: Query<(&living::Absorption, &mut player::AbsorptionAmount), Changed<living::Absorption>>
```
dyc3, line 533: What's the difference between `living::Absorption` and `player::AbsorptionAmount`? Feels like the same thing.

crates/valence_entity/build.rs
```diff
@@ -506,6 +524,22 @@ fn build() -> anyhow::Result<TokenStream> {
...
+        ) {
+            for (living_absorption, mut player_absorption) in &mut query {
```
dyc3, line 535: nit:
```suggestion
            for (living_absorption, mut player_absorption) in query.iter_mut() {
```

dyc3 (changes requested):

crates/valence_entity/build.rs
SelfMadeSystem: `player::AbsorptionAmount` is a **tracked** field (so updated via the `EntityTrackerUpdateS2c` packet), but `living::Absorption` is NOT tracked. `player::AbsorptionAmount` is only tracked to update the client on their own absorption level. Furthermore, `player::AbsorptionAmount` is **only** for the client, not for any other entity.

dyc3 (approved):

PR #566, opened 2023-10-16
Default to IPv6 bind address

belohnung:
# Default to IPv6 bind address

- By default applications should listen to the ipv6 wildcard address as this also includes IPv4 addresses and enables IPv6 support

# Solution

- The default bind ip address was set to the UNSPECIFIED constant of IPv6Adress

### DIFF
--- a/crates/valence_network/src/lib.rs
+++ b/crates/valence_network/src/lib.rs
@@ -23,7 +23,7 @@ mod legacy_ping;
 mod packet_io;
 
 use std::borrow::Cow;
-use std::net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4};
+use std::net::{IpAddr, Ipv6Addr, SocketAddr, SocketAddrV4};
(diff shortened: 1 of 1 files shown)


dyc3:
IIRC, this port binding behavior is only applicable to Linux. Does it do the same thing on Windows and Mac?

dyc3:
It appears that on windows, this PR will make it so that it only binds to ipv6, and not both.

Source: https://learn.microsoft.com/en-us/windows/win32/winsock/dual-stack-sockets#creating-a-dual-stack-socket

belohnung:
My bad. Could we set those sock opts so that the behavior matches?

dyc3:
You tell me, this is your PR.

belohnung:
It really sucks that the behavior is not uniform. With a quick search i also did not find a way to set a sockopt on the TcpListener.. Perhaps we could default to IPv6 on Linux only for now?

PR #568, opened 2023-10-18
Added food and saturation

### LINKED ISSUE #565: Player hunger

Player hunger isn't implemented yet (food & saturation).

> Note: implementing when the player loses hunger isn't needed for this issue, only that they can.

Implement the hunger manager & tell client about hunger updates.

Only implement the numbers ig idk

## Additional context

See [`net.minecraft.entity.player.HungerManager`](https://maven.fabricmc.net/docs/yarn-1.20.2+build.4/net/minecraft/entity/player/HungerManager.html) & [`HealthUpdateS2CPacket`](https://maven.fabricmc.net/docs/yarn-1.20.2+build.4/net/minecraft/network/packet/s2c/play/HealthUpdateS2CPacket.html)

SelfMadeSystem:
# Objective

- Food and saturation currently don't exist for the player in valence
- Closes #565 

# Solution

Adds food and saturation to the `PlayerEntity` and tells the client about any changes. The only problem right now is that the default value for `Health` is the same for ALL entities, so the server tells the client that their health is at 14 when it should be at 20.

## Note

Uncertain if best approach. Please lmk if another approach would be better

### DIFF
--- a/crates/valence_entity/build.rs
+++ b/crates/valence_entity/build.rs
@@ -408,6 +408,7 @@ fn build() -> anyhow::Result<TokenStream> {
                                 }]);
                             }
                             "PlayerEntity" => {
+                                //
                                 bundle_fields.extend([quote! {
                                     pub player_food: super::player::Food,
                                     pub player_saturation: super::player::Saturation,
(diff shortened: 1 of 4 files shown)


dyc3 (commented):
Looks good, but I would like to have a test or two to make sure this packet isn't being sent a tick late.

This is where we are putting tests right now: https://github.com/valence-rs/valence/tree/main/src/tests

dyc3 (changes requested):

dyc3 (approved):

PR #570, opened 2023-10-19
Set the default health of entities

### LINKED ISSUE #569: Set default health based on max health attribute

Currently, all entities' default health is the same. However, now that we have their default max health attribute, we can now set the default health value for each entity.

This problem isn't very apparent, but it becomes apparent when we include food and saturation via #565 / #568

Default the max health of entities

I don't think one is feasible.

## Additional context

Since #568 (HealthUpdateS2c) tells the client about their food level, saturation, AND health, we need to have good defaults.

SelfMadeSystem:
# Objective

- Set the default health of all entities now that we have the max health attribute
- Closes #569 

# Solution

- Use the max health attribute to set the default health of entities


SelfMadeSystem:
this appears to just be a workaround for the root issue of #359

PR #571, opened 2023-10-19
Add default fields for all entities in extractor

### LINKED ISSUE #359: Extractor output is non-determinisic.

SelfMadeSystem:
# Objective

- Extractor output is non-deterministic
- Leads to inconsistencies between defaults in valence and defaults in Minecraft
- Related to #359

# Solution

- Get the defaults of every entity and use them
- Remove defaults for individual fields
- Sort everything by id


SelfMadeSystem:
defo not gr8 approach. gonna need to think up a better one

SelfMadeSystem:
perhaps not as crazy as I initially thought

rj00a (approved):

crates/valence_entity/build.rs
```diff
@@ -421,23 +464,98 @@ fn build() -> anyhow::Result<TokenStream> {
...
+                            pub #field_name_ident: super::#stripped_snake_entity_name_1_ident::#pascal_field_name_ident,
                         }]);
 
                         bundle_init_fields.extend([quote! {
-                            #field_name_ident: Default::default(),
+                            #field_name_ident: super::#stripped_snake_entity_name_1_ident::#pascal_field_name_ident(#value_expr),
+                        }]);
+
+                        let system_name_ident = ident(format!(
+                            "update_{stripped_snake_entity_name}_{snake_field_name}_{stripped_snake_entity_name}",
```
rj00a, line 530: Does `stripped_snake_entity_name` need to be in the generated name twice?

crates/valence_entity/build.rs
```diff
@@ -421,23 +464,98 @@ fn build() -> anyhow::Result<TokenStream> {
...
+                    } => {
                         let snake_field_name = field.name.to_snake_case();
                         let pascal_field_name = field.name.to_pascal_case();
                         let pascal_field_name_ident = ident(&pascal_field_name);
-                        let stripped_entity_name = strip_entity_suffix(entity_name);
-                        let stripped_snake_entity_name = stripped_entity_name.to_snake_case();
-                        let stripped_snake_entity_name_ident = ident(&stripped_snake_entity_name);
+                        let stripped_entity_name = strip_entity_suffix(entity_name_2);
+                        let stripped_snake_entity_name_1 = stripped_entity_name.to_snake_case();
+                        let stripped_snake_entity_name_1_ident =
```
rj00a, line 476: Do we need these variables with the `_1` and `_2` suffixes? I would prefer something more descriptive.

crates/valence_entity/build.rs
```diff
@@ -421,23 +464,98 @@ fn build() -> anyhow::Result<TokenStream> {
...
+
+                        let data_index = field.index;
+                        let data_type = field.typ.type_id();
+                        let encodable_expr = field.typ.encodable_expr(quote!(value.0));
+
+                        systems.extend([quote! {
+                            #[allow(clippy::needless_borrow)]
+                            #[allow(clippy::suspicious_else_formatting)]
+                            fn #system_name_ident(
+                                mut query: Query<(&#component_path, &mut tracked_data::TrackedData), (Changed<#component_path>, With<#stripped_snake_entity_name_1_ident::#pascal_field_name_ident>)>,
```
rj00a, line 545: `With` in the query is redundant.

rj00a (commented):
Nice work!

I suspect bevy is not going to be happy about the huge number of systems being scheduled now. Each system has some constant overhead, especially with the multithreaded executor. You could try merging the systems together, but this is one problem I'm trying to address with my new ECS library.

PR #582, opened 2023-11-16
Fix command argument parse error checking

guac42:
# Objective

This pull request aims to fix an error when parsing commands with arguments.  This error can be recreated by running the command example and executing the command "/gamemode creative @a" or "/gamemode creative {username}" etc. in game.

# Solution

The order of subtraction between the pre-parse argument string and post-parse argument string was changed to hopefully accomplish the intended purpose. This fixes the error when running the command in the example.

Note: I'm not an expert on how command parsing works, maybe @JackCrumpLeys can verify the solution?

### DIFF
--- a/crates/valence_command/src/manager.rs
+++ b/crates/valence_command/src/manager.rs
@@ -350,8 +350,9 @@ fn parse_command_args(
                 let pre_input = input.clone().into_inner();
                 let valid = parser(&mut input);
                 if valid {
+                    // If input.len() > pre_input.len() the parser replaced the input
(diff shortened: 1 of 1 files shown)


JackCrumpLeys:
How did I miss this LOL!!! Thanks!!!

It looks like this was added in one of the last commits before merging to replace an unsafe version using manual memory management for the string.

Good catch, I recommend it be merged ASAP by @rj00a or @dyc3.

rj00a (approved):

PR #583, opened 2023-11-19
update bevy to 0.12.0

JackCrumpLeys:
# Objective

Bevy should be updated to 0.12.0.

# Solution

Update bevy.


rj00a (approved):
Looks good once the lints are fixed

PR #587, opened 2023-12-03
Process client packets before run schedulers

ramon-bernardo:
# Objective
- Process client packets before run schedulers, removing duplicate code.


dyc3 (commented):
This looks OK to me, but I would prefer a second opinion considering the context.

rj00a:
Closing as per #596

PR #591, opened 2023-12-26
Refactor attribute modifiers to use UUID

SelfMadeSystem:
# Objective

I made a bad decision to use names instead of UUID, like how Minecraft does it. This means that implementing certain features won't behave like they do in Vanilla (e.g. potion effects like speed, and running. they use specific UUIDs)

# Solution

- Use UUIDs


dyc3 (approved):
Looks good

PR #594, opened 2024-01-23
Update extractor to mc1.20.4

Bafbi:
# Objective

Update the protocol to version 765 for 1.20.4 and 1.20.3

# Solution

- [x] Update and run extractor
- [ ] Implement new packets


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,10 @@
+{
```
dyc3, line 1: You wanna move this into a separate PR so we can get  it in the repo a little quicker?

Bafbi: Yes, but how would I do that ?

dyc3: 1. copy this file
2. make a new branch based off main
3. commit and make pr

Bafbi: Ok yea that was I was thinking

PR #598, opened 2024-01-29
fix indexmap deprication warnings

AviiNL:
# Objective

IndexMap remove methods are deprecated in favor of [swap|shift]_remove

# Solution

match serde_json's implementation of swap and shift remove

### DIFF
--- a/crates/valence_nbt/src/compound.rs
+++ b/crates/valence_nbt/src/compound.rs
@@ -137,13 +137,39 @@ where
         self.map.insert(k.into(), v.into())
     }
 
+    #[inline]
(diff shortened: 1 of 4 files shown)


crates/valence_nbt/src/compound.rs
```diff
@@ -137,21 +137,71 @@ where
         self.map.insert(k.into(), v.into())
     }
 
+    #[inline]
```
rj00a, line 140: Let's not add `#[inline]` for consistency with the other methods. (I believe a recent Rust version made this unnecessary anyway)

Issue #323, opened 2023-04-23
Poor performance in `bench_players` example.


rj00a:
## Valence Version

abf90649017ef4037f7cb75c4d6c9679042ffc40

## What You Did

The `bench_players` example is indicating poor performance. This is running in release mode without any players connected:

```
Tick=10, MSPT=8.1462ms, Clients=0
Tick=20, MSPT=7.8771ms, Clients=0
Tick=30, MSPT=7.8221ms, Clients=0
Tick=40, MSPT=7.7786ms, Clients=0
Tick=50, MSPT=7.8106ms, Clients=0
Tick=60, MSPT=8.5027ms, Clients=0
Tick=70, MSPT=8.0414ms, Clients=0
Tick=80, MSPT=8.5672ms, Clients=0
Tick=90, MSPT=7.6703ms, Clients=0
```

But the `idle_update` benchmark is under similar conditions, and seems to suggest that a tick should take ~300 _microseconds_.

```
idle_update             time:   [378.47 µs 380.01 µs 381.72 µs]
                        change: [-5.5286% -1.2899% +3.5451%] (p = 0.63 > 0.05)
```

## What Went Wrong

The `bench_players` example should have a MSPT similar to the `idle_update` benchmark, but it does not.

## Additional Information

`cargo flamegraph` does not seem to indicate anything. Possibly a bug in `valence_network` or timing issue?

rj00a:
Updating to bevy 0.11-dev in #392 improved the situation significantly, but there's still an unacceptably high MSPT here.

```
Tick=10, MSPT=2.5595ms, Clients=0
Tick=20, MSPT=2.6124ms, Clients=0
Tick=30, MSPT=2.4096ms, Clients=0
Tick=40, MSPT=2.7876ms, Clients=0
Tick=50, MSPT=2.5686ms, Clients=0
Tick=60, MSPT=2.3812ms, Clients=0
Tick=70, MSPT=2.3343ms, Clients=0
Tick=80, MSPT=2.7529ms, Clients=0
Tick=90, MSPT=2.5710ms, Clients=0
Tick=100, MSPT=1.7957ms, Clients=0
Tick=110, MSPT=2.4565ms, Clients=0
Tick=120, MSPT=0.8715ms, Clients=0
```

However, for some strange reason, if I have singleplayer Minecraft open on my machine at the same time I'm running bench_players, the performance actually _improves_.

```
Tick=10, MSPT=0.5308ms, Clients=0
[...]

PR #599, opened 2024-02-03
Rewrite Project

### LINKED ISSUE #596: Rewrite the project with `evenio`
For the past few months I have been working on a little ECS library called [`evenio`](https://github.com/rj00a/evenio). This is in response to some limitations I have encountered using `bevy_ecs` in valence. In essence, I believe that "run once per frame" systems as the primary abstraction is fundamentally inadequate. This one fact has been the source of many design issues in valence. Here are some of the problems and how `evenio` could fix them:

- In bevy_ecs, there aren't any reliable ways to enforce that entities have certain combinations of components because there's always a delay between when the violation occurs and when it can be acted on. As a result, we don't even attempt to detect these potential mistakes in valence. `evenio` lets us write handlers that run the moment before a component is added or removed, making these violations feasible to detect and act on.

- Whenever an entity is despawned we almost always need to do some cleanup work. Bevy lets us detect that an entity is removed from the world, but doesn't give access to the entity's components because the entity is already gone. We work around this in valence by using a `Despawned` component to signal removal, then letting systems in `PostUpdate` run before it's removed. This works, but it's very error prone because there's nothing stopping someone from calling `.despawn()` instead of adding the marker. `evenio` fixes this problem by signaling entity despawns with an event, allowing code to run immediately _before_ the entity is removed.

- A lot of valence's APIs are based on a pattern where we let the user make some changes to the world in `Update` and then defer handling the changes until `PostUpdate`. Compared to the simple procedural approach, This is a lot more cumbersome. Case in point, I tried to improve layers and clean things up in #533 but the resulting code was so complicated and error prone that I gave up. In `evenio`, we can just use events to signal state changes instead of using change detection or diffing.

- While reading packets from clients, we split the stream of packets into separate event buffers based on packet type. The problem is, this loses the order between different packet types. To get around this, [I made an entirely separate  schedule](https://github.com/valence-rs/valence/blob/51238264b9e97def609e014e30f7228b9e1c8bb7/crates/valence_server/src/event_loop.rs) which runs in a loop to ensure that packet handler code is executed in the correct order. This is ugly and slow. #587 tried to simplify this a bit without being aware of these details. We could have easily fixed this by using system callbacks, but this wouldn't really integrate with bevy's schedules. In `evenio`, event handlers are run in the order events are produced (regardless of type) so the problem goes away.

- Every system added in Bevy has a cost. Systems run every frame even if they have no work to do. This is especially bad with the multithreaded executor due to the overhead involved. #571 tried to fix the entity extractor and entity tracked data, but the solution would have caused an explosion in the number of systems. With `evenio`, systems don't have a cost unless the event they're listening for is actually sent.

- In #323 I noticed poor performance while the server isn't doing anything. Likely related to the previous point.

- Bevy's change detection feature is not flexible enough. Sometimes it works fine, but in other cases [we end up using `.bypass_change_detection()`](https://github.com/valence-rs/valence/blob/51238264b9e97def609e014e30f7228b9e1c8bb7/crates/valence_server/src/abilities.rs#L106-L175) which results in footguns for end users.

- `evenio` doesn't have system order ambiguities, so #542 goes away.

So, here's my plan. Let's burn it all to the ground and start over with `evenio`. We can leave `valence_protocol` and everything below it mostly intact, but everything else needs to go. Although `evenio` needs quite a lot of work before I would call it production ready, it should be usable enough to get started. Bevy has recently adopted some features like one-shot systems to address the above problems, but I don't really think they get at the root of the issue.

During this rewrite I would like to do a few other things:
- Update to the latest MC version (assuming #594 doesn't do it first).
- Use completion-based IO for better perf (#489).
- Aggressively cut out unused dependencies. Compile times are too long.
- CI needs to be cleaned up.
- Actually handle timesteps correctly (#327).

rj00a:
Rewrite the project from the ground up.  Closes #596

# Steps
- [x] Update data extractor to the latest version and fix known issues.
- [x] Fix issues in `java_string`.
- [x] Rework interface of `valence_nbt` to support redesign of `valence_protocol`.
- [x] Remove `valence_generated`.
  - Computing everything statically doesn't work in a world where more and more registries are networked.
  - Compile times were too long. We can generate stuff from the JSON at runtime if we need to.
- [ ] Redesign `valence_protocol` and update to 1.20.4.
  - Reorganize packets into a single module. (Minecraft shares some packets between protocol stages now, so this accommodates for that). 
  - Introduce `McRead` and `McWrite` traits similar to `PacketByteBuf` in vanilla.
  - Remove bespoke derive macro for "encodable" and "decodable" types and write all the packet impls out manually.
  - Remove dependence on `valence_generated`, `bevy_ecs`, `valence_math`.
  - Consider merging `valence_text` and `valence_ident` into `valence_protocol` since we don't have `valence_generated` anymore.
- [ ] Update the packet inspector and get it to a usable state.
   - Create "packet registry" in `valence_protocol` so the inspector can access all packets programmatically.
- [ ] Rewrite main server crates.
  - Replace `bevy_ecs` with `evenio`. Use event driven design described in #596.
  - Use `monoio` + `ureq` on Linux and MacOS, `tokio` everywhere else.
  - Handle switching between configuration phase and play phase.
  - Get basic chunk stuff working at least.
  - Combine the miscellaneous crates together like weather, player_list, boss_bar, registry, etc.
- [ ] Pick a linear algebra crate (`glam`, `vek`, `ultraviolet`, something else?)
- [ ] Port some of the examples over.
- [ ] Update `valence_anvil` (or leave that for follow up?)
- [ ] Update CI scripts.


andrewgazelka:
Given that you are making a bunch of changes and performance is important, have you considered adding benching to CI? I feel like I have seen Rust projects that have good performance regression detection in CI, but I might need to do a bit of searching. 

This would be nice because performance would be repeatable unlike

> if I have singleplayer Minecraft open on my machine at the same time I'm running bench_players, the performance actually improves.

https://github.com/valence-rs/valence/issues/323

rj00a:
@andrewgazelka I'm open to the idea, but it seems like a lot of work to do it correctly. For starters we would need dedicated CI runners and a decent suite of benches. Definitely something to think about after the rewrite is finished.

A-z-z-y:
The `criterion` crate sounds like it may be useful here, unless ya'll have your own favorite (micro)benchmarking tools.

andrewgazelka:
> @andrewgazelka I'm open to the idea, but it seems like a lot of work to do it correctly. For starters we would need dedicated CI runners and a decent suite of benches. Definitely something to think about after the rewrite is finished.

That makes sense. Why would you need dedicated CI runners, though? Do you think performance would not be predictable using the default CI runners, which run in a VM on the same machine as other runners?

A-z-z-y:
If consistent environments are available (e.g. on your own machine), it may be worth it to use a microbenchmarking crate like criterion to check for performance regressions.

splurf:
For the lin alg crate, I'd say either [glam](https://crates.io/crates/glam) or [nalgebra](https://crates.io/crates/nalgebra). There's a [benchmark](https://github.com/bitshifter/mathbench-rs#scalar-benchmarks) that compares a bunch of popular lin alg crates but it's over 3 years old.

rj00a:
Closing this since I'm not happy with the direction this was going and it's too much to do in one PR anyway.

PR #608, opened 2024-02-21
added switching offhand with mainhand via PlayerAction packet

cooltexture1:
i added the code to make switching the items in main and offhand work with the PlayerAction from it.
I tested it and it worked.

### DIFF
--- a/crates/valence_inventory/src/lib.rs
+++ b/crates/valence_inventory/src/lib.rs
@@ -1151,7 +1151,12 @@ fn handle_player_actions(
                     }
                 }
                 PlayerAction::SwapItemWithOffhand => {
-                    // TODO
+                    if let Ok((mut inv, _, held)) = clients.get_mut(packet.client) {
+                        let stack1 = inv.replace_slot(held.slot(), ItemStack::EMPTY);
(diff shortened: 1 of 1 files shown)


crates/valence_inventory/src/lib.rs
```diff
@@ -1151,7 +1151,12 @@ fn handle_player_actions(
                     }
                 }
                 PlayerAction::SwapItemWithOffhand => {
-                    // TODO
+                    if let Ok((mut inv, _, held)) = clients.get_mut(packet.client) {
+                        let stack1 = inv.replace_slot(held.slot(), ItemStack::EMPTY);
+                        let stack2 = inv.replace_slot(PlayerInventory::SLOT_OFFHAND, stack1);
+
+                        inv.set_slot(held.slot(), stack2);
```
dyc3, line 1158: you should use `swap_slot` instead

dyc3 (changes requested):

crates/valence_inventory/src/lib.rs
cooltexture1: ah i didnt realize it could be done so easily. i changed it use `swap_slot`.

dyc3 (approved):

PR #635, opened 2024-07-24
Extractor 1.21

Bafbi:
# Objective

Having the extractor upgrade for 1.21

# Solution

The extractor is runnable but with some difference in the generated files :
- The codex is no longer generated until I find how
- same for paintings and enchants but they were not used
- the packets file structure have changed


dyc3:
It looks like the "Build Extractor" CI job needs to be updated

JackCrumpLeys:
@dyc3 @Bafbi this should be closed in light of #641 being merged

PR #654, opened 2024-10-07
update nix flake environment

lukashermansson:
# Objective

I noticed that when trying to diagnose an issue that we don't compile with the current nix lock, (rust being to old) 
```
error: package `tokio-macros v2.4.0` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.65.0-nightly
```

# Solution
ran `nix flake update`, and now it brings in a recent enough rust nightly for everything to compile

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "flake-compat": {
       "flake": false,
       "locked": {
-        "lastModified": 1650374568,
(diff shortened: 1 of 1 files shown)


dyc3 (approved):

lukashermansson:
This has a test failure, dont merge this just yet, will need to get an implementation that does not send a packet to the client when it was the client that updated it! thanks for the quick review!

Issue #307, opened 2023-03-24
Inventory Menus


dyc3:
## Describe the problem related to your feature request.

It's pretty common to (ab)use inventory windows to provide menu-like user interfaces.

## What solution would you like?

We need to cover the following use cases:
- [ ] A simple 1 page menu that has some buttons, eg. a menu that allows players to select minigame to play
- [ ] A menu with an initial page, and you can navigate to sub pages, eg. pick a minigame, then pick a team to join
    - [ ] Unique window types for each page
- [ ] A paginated list of items, eg. viewing a long list of search results
- [ ] A paginated list of items that you can interact with like a regular inventory, eg. a really really big chest that has pagination to display all the items

We should have a new component `InventoryMenu`, that represents a menu. 

This could go on the client's entity that has the menu open, similar to how `OpenInventory` works. This would allow users to render pages with customized information easier.

Alternatively, `InventoryMenu` could go on its own entity, similar to how `Inventory` works. This would allow menus to be cached and reused easier. This approach could end up looking something like this:

```mermaid
classDiagram
    direction TB
    class InventoryMenu {
        <<Component>>
        page: Vec~Page~

        + new()
        + with_pages(Vec~Page~ pages)
        - render(state: &InventoryMenuState)
    }

    class Page {
        buttons: HashMap "HashMap of slot to Button"
        kind: InventoryKind

        +new(kind: InventoryKind)
        +with_buttons(buttons)
        -render(): Vec~Option~ItemStack~~ 
    }

    class Button {
        item: ItemKind
        name: Option~Text~
        lore: Option~Vec~Text~~

        new()
        -render(): ItemStack
    }

    InventoryMenu "1" --> "*" Page
    Page "1" --> "*" Button

    class InventoryMenuState {
        <<Component>>
        page: usize
        menu: Entity
        inventory: Entity
    }

    InventoryMenuState .. InventoryMenu
    InventoryMenuState .. MenuInventory

    class MenuInventory {
        <<Component>>
        Inventory
    }
    note for MenuInventory "would be struct MenuInventory(Inventory), \nrepresents the Inventory that is used to render\nthe page"

    class Inventory {
        <<Component>>
    }

    MenuInventory --* Inventory
```

`MenuInventory` is a wrapper component around `Inventory` so that queries for `Inventory` don't conflict with menus.

Unresolved questions:
- [ ] Tree structure for pages?
[...]

PR #655, opened 2024-10-08
Readonly inventory

### LINKED ISSUE #427: read-only inventories

Creating a Item menu I noticed that there was no obvious way to disable parts of the inventory system to do with adding and removing items.

A flag for inventory to label them as read only to the internal systems to make it so that the menu cannot be modified by anything other than the server. This could be a component or a field in `Inventory`.

Re implement all the systems manually for my server, this would be non-ideal.

maxomatic458:
# Objective
- ability to make inventories read only (player should be able to click things in the inventory and still emit a click event, this can be useful for creating inventory menus. 
- closes  #427
- related #307 #355 
# Solution

- adds a public ``readonly: bool`` field to the ``Inventory`` component, that will make any interactions with this item impossible (includes: moving, shift moving, hotbar moving, dropping) if a player inventory is readonly, then the player will also not be able to drop items (even when not in the inventory), so the drop event will not be emitted (this could be changed if requested)
- when implementing this i discovered a bug where a player is not able to put a item from a open inventory in the offhand (by hitting F) that will cause a desync. On the client the item will be in the offhand, but if you try to interact with that it dissapears. (unrelated to this PR and will not be fixed in this PR)

### DIFF
--- a/crates/valence_inventory/src/lib.rs
+++ b/crates/valence_inventory/src/lib.rs
@@ -71,6 +71,11 @@ pub struct Inventory {
     /// Contains a set bit for each modified slot in `slots`.
     #[doc(hidden)]
     pub changed: u64,
+    /// Makes a inventory read-only. This will prevent adding
+    /// or removing items. If this is a player inventory
+    /// This will also make it impossible to drop items while not 
+    /// in the inventory (e.g. by pressing Q)
+    pub readonly: bool,
 }
 
 impl Inventory {
@@ -85,6 +90,7 @@ impl Inventory {
             kind,
             slots: vec![ItemStack::EMPTY; kind.slot_count()].into(),
             changed: 0,
+            readonly: false,
         }
     }
 
@@ -862,7 +868,7 @@ fn handle_click_slot(
 
         if pkt.slot_idx < 0 && pkt.mode == ClickMode::Click {
             // The client is dropping the cursor item by clicking outside the window.
-
+            
             let stack = std::mem::take(&mut cursor_item.0);
 
             if !stack.is_empty() {
@@ -907,6 +913,17 @@ fn handle_click_slot(
                 if (0_i16..target_inventory.slot_count() as i16).contains(&pkt.slot_idx) {
                     // The player is dropping an item from another inventory.
 
+                    if target_inventory.readonly {
+                        // resync target inventory
+                        client.write_packet(&InventoryS2c {
+                            window_id: inv_state.window_id,
+                            state_id: VarInt(inv_state.state_id.0),
+                            slots: Cow::Borrowed(target_inventory.slot_slice()),
+                            carried_item: Cow::Borrowed(&cursor_item.0),
+                        });
+                        continue;
+                    }
+
                     let stack = target_inventory.slot(pkt.slot_idx as u16);
 
                     if !stack.is_empty() {
@@ -930,6 +947,18 @@ fn handle_click_slot(
                     }
                 } else {
                     // The player is dropping an item from their inventory.
+
+                    if client_inv.readonly {
+                        // resync the client inventory
+                        client.write_packet(&InventoryS2c {
+                            window_id: 0,
+                            state_id: VarInt(inv_state.state_id.0),
+                            slots: Cow::Borrowed(client_inv.slot_slice()),
+                            carried_item: Cow::Borrowed(&cursor_item.0),
+                        });
+                        continue;
+                    }
+
                     let slot_id =
                         convert_to_player_slot_id(target_inventory.kind, pkt.slot_idx as u16);
 
@@ -957,6 +986,17 @@ fn handle_click_slot(
             } else {
                 // The player has no inventory open and is dropping an item from their
                 // inventory.
+
+                if client_inv.readonly {
+                    // resync the client inventory
+                    client.write_packet(&InventoryS2c {
+                        window_id: 0,
+                        state_id: VarInt(inv_state.state_id.0),
+                        slots: Cow::Borrowed(client_inv.slot_slice()),
+                        carried_item: Cow::Borrowed(&cursor_item.0),
+                    });
+                    continue;
+                }
 
                 let stack = client_inv.slot(pkt.slot_idx as u16);
 
@@ -994,7 +1034,8 @@ fn handle_click_slot(
             }
 
             if let Some(mut open_inventory) = open_inventory {
-                // The player is interacting with an inventory that is open.
+                // The player is interacting with an inventory that is
+                // open or has an inventory open while interacting with their own inventory.
 
                 let Ok(mut target_inventory) = inventories.get_mut(open_inventory.entity) else {
                     // The inventory does not exist, ignore.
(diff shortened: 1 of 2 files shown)


lukashermansson:
Hey nice work, just chiming in with a couple of thoughts, 

I read what RJ wrote in the draft for inventory menus in the draft pr #355 it is possible to implement readonly behavior in user space. So I think exploring implementing this as a separate crate we maintain is worthwhile.

The alternative approach is to not bloat up the inventory crate with the readonly stuff, it should be implementable as its own separate crate with the approach to listen to the `ClickSlotEvent`, `DropItemStackEvent` and `CreativeInventoryActionEvent` and modify the inventory back to its previous state based on the information presented in the event. (setting the modified slots back to how they where, and clearing the cursor_item)

With this approach it should be possible *i think* to have another create `valance_readonly_inventory` that could be optionally added along with its own "read only" marker component (to know when these actions should be reverted by the readonly inventory plugin)

One footgun with the described approach (the separate crate one) is that *for example* listing to the `DropItemStackEvent` and spawning an item entity https://minecraft.fandom.com/wiki/Item_(entity) without also properly filtering out the `Readonly` inventories would be easy to forget. If its in the same crate I guess its simpler to not emit the events unless "real stuff happend", so I guess that should be considered before picking one approach over the other. (This might not be the first place such a problem exists?)

I don't really have any strong argument for one approach or the other myself currently. I think its worthwile to bring us this approach too :D

crates/valence_inventory/src/lib.rs
```diff
@@ -71,6 +71,11 @@ pub struct Inventory {
     /// Contains a set bit for each modified slot in `slots`.
     #[doc(hidden)]
     pub changed: u64,
+    /// Makes a inventory read-only. This will prevent adding
```
dyc3, line 74:
```suggestion
    /// Makes an inventory read-only for clients. This will prevent adding
```

dyc3 (changes requested):
I'm not entirely opposed to having a `readonly` flag on `Inventory`. Should make inventory menus easier.

There needs to be some tests though.

maxomatic458:
im currently making some tests

and i made this test here:

```rs
#[test]
fn test_prevent_hotbar_item_click_container_readonly_inventory() {
    let ScenarioSingleClient {
        mut app,
        client,
        mut helper,
        ..
    } = ScenarioSingleClient::new();

    // Process a tick to get past the "on join" logic.
    app.update();
    helper.clear_received();

    // player inventory is not read-only
    let mut player_inventory = app
        .world_mut()
        .get_mut::<Inventory>(client)
        .expect("could not find inventory for client");

    // 36 is the first hotbar slot
    player_inventory.set_slot(36, ItemStack::new(ItemKind::Diamond, 1, None));

    let open_inv_ent = set_up_open_inventory(&mut app, client);

    let mut open_inventory = app
        .world_mut()
        .get_mut::<Inventory>(open_inv_ent)
        .expect("could not find inventory for client");

    // Open inventory is read-only
    // open_inventory.readonly = true;
    open_inventory.set_slot(0, ItemStack::new(ItemKind::IronIngot, 10, None));

    let inv_state = app.world_mut().get::<ClientInventoryState>(client).unwrap();
    let state_id = inv_state.state_id();
    let window_id = inv_state.window_id();

    // The player hovers over the iron ingots in the open inventory, and tries
    // to move them to their own (via pressing 1), which should swap the iron
    // for the diamonds. However the opened inventory is read-only, so nothing 
    // should happen.
    helper.send(&ClickSlotC2s {
        window_id,
        state_id: VarInt(state_id.0),
        slot_idx: 0,
        button: 0, // hotbar slot starting at 0
        mode: ClickMode::Hotbar,
        slot_changes: vec![
            // First SlotChange is the item is the slot in the player's hotbar.
            // target slot.
            SlotChange {
                idx: 0,
                stack: ItemStack::new(ItemKind::Diamond, 1, None),
            },
            SlotChange {
                // 54 is the players hotbar slot 1, when the 9x3 inventory is opnened.
                idx: 54,
                stack: ItemStack::new(ItemKind::IronIngot, 10, None),
            },
            // The second one is the slot in the open inventory, after the ClickSlot action
            // source slot.
        ]
        .into(),
        carried_item: ItemStack::EMPTY,
    });

    app.update();

    let sent_packets = helper.collect_received();

    // 1 resync for each inventory
    sent_packets.assert_count::<InventoryS2c>(2);

    // Make assertions
    let player_inventory = app
        .world_mut()
        .get::<Inventory>(client)
        .expect("could not find client");

    // Opened inventory is read-only, the items are not swapped.
    assert_eq!(player_inventory.slot(36), &ItemStack::new(ItemKind::Diamond, 1, None));

    let open_inventory = app
        .world_mut()
        .get::<Inventory>(open_inv_ent)
        .expect("could not find inventory");

    // Opened inventory is read-only, the items are not swapped.
    assert_eq!(open_inventory.slot(0), &ItemStack::new(ItemKind::IronIngot, 10, None));
}
```

this test here will pass if the items were not swapped (as they usually should)

for some reason, this always passes (even when the inventory is not readonly, and this also happens on the latest master branch)

The event being sent is the exact same that is also sent when i test it with a game client and replicate this (and that works fine, no desync)

Any ideas why the inventory slots are not updating in the inventory component?

maxomatic458:
ive added all the tests ive made so far, as i said in the message above some of them do not work (because of the strange behavior where the inventory content does not update). Ive commented them out for now.
Let me know if you have any idea why those are not working as expected.

dyc3:
IIRC, We have some basic anti cheat validation on `slot_changes` that ignores changes that would result in item duplication. The items need to exist in the inventories before they can be moved. I can't give you a code link at the moment, but I would look in that direction.

lukashermansson:
After a lot of trial and error i finally figured this out 

This test below is an implementation that can properly swap two items (NON-read only) should be easy enough to make it pass in the read only case.

there was an update missing, so the items did not really "exist" yet at the time the client packet got sent.
```
#[test]
fn test_hotbar_item_swap_container() {
    let ScenarioSingleClient {
        mut app,
        client,
        mut helper,
        ..
    } = ScenarioSingleClient::new();

    // Process a tick to get past the "on join" logic.
    app.update();

    let mut player_inventory = app
        .world_mut()
        .get_mut::<Inventory>(client)
        .expect("could not find inventory for client");

    // 36 is the first hotbar slot
    player_inventory.set_slot(36, ItemStack::new(ItemKind::Diamond, 1, None));

    let open_inv_ent = set_up_open_inventory(&mut app, client);

    let mut open_inventory = app
        .world_mut()
        .get_mut::<Inventory>(open_inv_ent)
        .expect("could not find inventory for client");

    open_inventory.set_slot(0, ItemStack::new(ItemKind::IronIngot, 10, None));

    // This update makes sure we have the items in the inventory by the time the
    // client wants to update these
    app.update();
    helper.clear_received();
    let inv_state = app.world_mut().get::<ClientInventoryState>(client).unwrap();
    let state_id = inv_state.state_id();
    let window_id = inv_state.window_id();

    // The player hovers over the iron ingots in the open inventory, and tries
    // to move them to their own (via pressing 1), which should swap the iron
    // for the diamonds.
    helper.send(&ClickSlotC2s {
        window_id,
        state_id: VarInt(state_id.0),
        slot_idx: 0,
        button: 0, // hotbar slot starting at 0
        mode: ClickMode::Hotbar,
        slot_changes: vec![
            // First SlotChange is the item is the slot in the player's hotbar.
            // target slot.
            SlotChange {
                idx: 0,
                stack: ItemStack::new(ItemKind::Diamond, 1, None),
            },
            SlotChange {
                // 54 is the players hotbar slot 1, when the 9x3 inventory is opnened.
                idx: 54,
                stack: ItemStack::new(ItemKind::IronIngot, 10, None),
            },
            // The second one is the slot in the open inventory, after the ClickSlot action
            // source slot.
        ]
        .into(),
        carried_item: ItemStack::EMPTY,
    });

    app.update();

    let sent_packets = helper.collect_received();

    // No resyncs beacuse the client was in sync and sent us the updates
    sent_packets.assert_count::<InventoryS2c>(0);

    // Make assertions
    let player_inventory = app
        .world_mut()
        .get::<Inventory>(client)
        .expect("could not find client");

    // Swapped items sucessfully
    assert_eq!(
        player_inventory.slot(36),
        &ItemStack::new(ItemKind::IronIngot, 10, None)
    );

    let open_inventory = app
        .world_mut()
        .get::<Inventory>(open_inv_ent)
        .expect("could not find inventory");

    assert_eq!(
        open_inventory.slot(0),
        &ItemStack::new(ItemKind::Diamond, 1, None)
    );
}
```

what finnaly helped me figure everything out was setting
`RUST_LOG=debug` as an env variable and running 
cargo test tests::inventory::test_hotbar_item_swap_container

the output gave a lot away, the resyncs where also a bit suspicious, as when i looked at what happens with a client and server (with the packet inspector) in this instance no re-syncs where observed

maxomatic458:
should be ready for review now!

dyc3 (approved):

PR #659, opened 2024-10-11
add ``item_menu`` example

maxomatic458:
Adds a example for using an inventory with the ``readonly`` flag as a item menu (using a ``ItemMenu`` component, that can be attached to a player)

related #307

video clip:

https://github.com/user-attachments/assets/441e165a-139a-43ec-9970-3055c7f7f79a

### DIFF
--- a/examples/item_menu.rs
+++ b/examples/item_menu.rs
@@ -134,7 +134,7 @@ fn on_make_selection(
             1.0,
             1.0,
         );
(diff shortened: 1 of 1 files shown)


dyc3 (approved):
Great!

Could you add a brief module doc comment to the example explaining the approach to users? 1-2 sentences is fine.

dyc3 (approved):

PR #660, opened 2024-10-11
implement (de)serialize for ``ItemStack`` and ``ItemKind``

maxomatic458:
# Objective
- Make it possible to serialize/deserialize ItemStacks without having to implement custom logic
# Solution
- implement ``serde::Serialize`` and ``serde::Deserialize``, ItemKinds will be serialized to their snake case name: e.g `white_wool`

### DIFF
--- a/crates/valence_generated/build/item.rs
+++ b/crates/valence_generated/build/item.rs
@@ -173,6 +173,9 @@ pub(crate) fn build() -> anyhow::Result<TokenStream> {
         .collect::<TokenStream>();
 
     Ok(quote! {
+        use serde::{Deserialize, Deserializer, Serialize, Serializer};
+        use serde::de::{self, Unexpected};
+
         /// Represents an item from the game
         #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
         #[repr(u16)]
@@ -191,6 +194,27 @@ pub(crate) fn build() -> anyhow::Result<TokenStream> {
             pub always_edible: bool,
             pub meat: bool,
             pub snack: bool,
+        }
+
+        impl Serialize for ItemKind {
+            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+            where
+                S: Serializer,
+            {
+                serializer.serialize_str(self.to_str())
+            }
+        }
+
+        impl<'de> Deserialize<'de> for ItemKind {
+            fn deserialize<D>(deserializer: D) -> Result<ItemKind, D::Error>
+            where
+                D: Deserializer<'de>,
+            {
+                let s: &str = Deserialize::deserialize(deserializer)?;
+                ItemKind::from_str(s).ok_or_else(|| {
+                    de::Error::invalid_value(Unexpected::Str(s), &"the snake case item name, like \"white_wool\"")
+                })
+            }
         }
 
         impl ItemKind {
--- a/crates/valence_generated/build/item.rs
+++ b/crates/valence_generated/build/item.rs
@@ -173,9 +173,6 @@ pub(crate) fn build() -> anyhow::Result<TokenStream> {
         .collect::<TokenStream>();
 
     Ok(quote! {
-        use serde::{Deserialize, Deserializer, Serialize, Serializer};
-        use serde::de::{self, Unexpected};
-
         /// Represents an item from the game
         #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
         #[repr(u16)]
@@ -196,23 +193,23 @@ pub(crate) fn build() -> anyhow::Result<TokenStream> {
             pub snack: bool,
         }
 
-        impl Serialize for ItemKind {
+        impl serde::Serialize for ItemKind {
             fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
             where
-                S: Serializer,
+                S: serde::Serializer,
             {
                 serializer.serialize_str(self.to_str())
             }
         }
 
-        impl<'de> Deserialize<'de> for ItemKind {
+        impl<'de> serde::Deserialize<'de> for ItemKind {
             fn deserialize<D>(deserializer: D) -> Result<ItemKind, D::Error>
             where
-                D: Deserializer<'de>,
+                D: serde::Deserializer<'de>,
             {
-                let s: &str = Deserialize::deserialize(deserializer)?;
+                let s: &str = serde::Deserialize::deserialize(deserializer)?;
                 ItemKind::from_str(s).ok_or_else(|| {
-                    de::Error::invalid_value(Unexpected::Str(s), &"the snake case item name, like \"white_wool\"")
+                    serde::de::Error::invalid_value(serde::de::Unexpected::Str(s), &"the snake case item name, like \"white_wool\"")
                 })
             }
         }
--- a/crates/valence_protocol/src/item.rs
+++ b/crates/valence_protocol/src/item.rs
@@ -1,12 +1,13 @@
 use std::io::Write;
 
+use serde::{Deserialize, Serialize};
 pub use valence_generated::item::ItemKind;
 use valence_nbt::Compound;
 
 use crate::{Decode, Encode};
 
 /// A stack of items in an inventory.
-#[derive(Clone, PartialEq, Debug, Default)]
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug, Default)]
 pub struct ItemStack {
     pub item: ItemKind,
     pub count: i8,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,7 @@ divan.workspace = true
 flume.workspace = true
 noise.workspace = true     # For the terrain example.
 tracing.workspace = true
+serde_json.workspace = true # For serialization tests.
 
 [dev-dependencies.reqwest]
 workspace = true
--- a/crates/valence_generated/Cargo.toml
+++ b/crates/valence_generated/Cargo.toml
@@ -16,6 +16,7 @@ workspace = true
 valence_math.workspace = true
 valence_ident.workspace = true
 uuid.workspace = true
+serde = { workspace = true, features = ["derive"] }
 
 [build-dependencies]
 anyhow.workspace = true
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -7,5 +7,6 @@ mod layer;
 mod player_list;
 mod potions;
 mod scoreboard;
+mod serialization;
 mod weather;
 mod world_border;
--- a/src/tests/serialization.rs
+++ b/src/tests/serialization.rs
@@ -1,0 +1,35 @@
+use crate::ItemKind;
+
+#[test]
+fn test_serialize_item_kind() {
+    let item = ItemKind::WhiteWool;
+    let serialized = serde_json::to_string(&item).unwrap();
+    assert_eq!(serialized, "\"white_wool\"");
+
+    let item = ItemKind::GoldenSword;
+    let serialized = serde_json::to_string(&item).unwrap();
+    assert_eq!(serialized, "\"golden_sword\"");
+
+    let item = ItemKind::NetheriteAxe;
+    let serialized = serde_json::to_string(&item).unwrap();
+    assert_eq!(serialized, "\"netherite_axe\"");
+
+    let item = ItemKind::WaxedWeatheredCutCopperStairs;
+    let serialized = serde_json::to_string(&item).unwrap();
+    assert_eq!(serialized, "\"waxed_weathered_cut_copper_stairs\"");
+}
+
+#[test]
+fn test_deserialize_item_kind() {
+    let id = "\"diamond_shovel\"";
+    let deserialized: ItemKind = serde_json::from_str(id).unwrap();
+    assert_eq!(deserialized, ItemKind::DiamondShovel);
+
+    let id = "\"minecart\"";
+    let deserialized: ItemKind = serde_json::from_str(id).unwrap();
+    assert_eq!(deserialized, ItemKind::Minecart);
+
+    let id = "\"vindicator_spawn_egg\"";
+    let deserialized: ItemKind = serde_json::from_str(id).unwrap();
+    assert_eq!(deserialized, ItemKind::VindicatorSpawnEgg);
+}


lukashermansson:
I thought about doing this PR myself also, however I felt like it was a bit dangerous to expose this feature in the "core" of the framwork.

If people start relying on this serialization to work across versions we are limited in what kind of changes we can do to the generation of `ItemKind`, and the internal structure of what a `ItemStack` is without causing breakage.

My personal stance on this is that we should keep serialization out of the scope for the framework(core), as the added overhead of serialized items will result in one of two things (or both):
1. We are scared to break previously serialized structs of this type (making any adjustments to the way we store these internally harder and may limit what kind of changes we can do)
2. We do not care about breaking previously serialized structs between versions, possibly silently breaking between versions. Would be error prone for our users (the changes in the code would not break *silently* between versions). Additionally, measures taken to try to not break silently here might tie into option 1, some changes might negatively impact what type of changes we can do.

I think we would be leaning a bit more towards approach 2 here right? from the perspective of what commitment we make for the serialization and de serialization to work over time. 

I'm not really sure what would be the ideal solution for this (possibly a valance_serialize module that could deal with all (or part of the complexity)? (at the very least tag a serialized "thing" with a version so we can reject de-serializing and make sure we never fail silently)) until then i think it might be wise to leave any kind of serialization and de-serialization to the user.

I would be happy to hear all of your opinions on it, but I personally would be a bit scared of us going around and making the structs serializable without some kind of abstraction layer. I think it could be a false sense of security to the users of the framework that this serialization can be safely used, and if we want to commit to that assumption users might make, we are limited in the ways we can change these structs.

crates/valence_generated/build/item.rs
```diff
@@ -193,6 +196,27 @@ pub(crate) fn build() -> anyhow::Result<TokenStream> {
             pub snack: bool,
         }
 
+        impl Serialize for ItemKind {
```
dyc3, line 199: The imports from serde should use the fully qualified name, not imported with `use` (macro hygiene)

crates/valence_generated/build/item.rs
```diff
@@ -193,6 +196,27 @@ pub(crate) fn build() -> anyhow::Result<TokenStream> {
...
+            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+            where
+                S: Serializer,
+            {
+                serializer.serialize_str(self.to_str())
```
dyc3, line 204: What if we serialized this into the Minecraft identifier for the item instead? Would that make more sense?

crates/valence_protocol/src/item.rs
```diff
@@ -1,12 +1,13 @@
...
 
+use serde::{Deserialize, Serialize};
 pub use valence_generated::item::ItemKind;
 use valence_nbt::Compound;
 
 use crate::{Decode, Encode};
 
 /// A stack of items in an inventory.
-#[derive(Clone, PartialEq, Debug, Default)]
+#[derive(Serialize, Deserialize, Clone, PartialEq, Debug, Default)]
```
dyc3, line 10: Should we put these behind a `serde` feature?

dyc3 (changes requested):
Definitely needs some tests for the ItemKind impls.

crates/valence_generated/build/item.rs
maxomatic458: like `minecraft:white_wool`?

crates/valence_protocol/src/item.rs
maxomatic458: I think serializing Itemstacks isnt that rare of a case.
Im currently creating a minigame server and i had to save items to a json file (for a ingame shop menu).

Im fine with a either.

maxomatic458:
@lukashermansson 
> My personal stance on this is that we should keep serialization out of the scope for the framework(core), as the added overhead of serialized items will result in one of two things (or both):

we could put that behind a feature flag

> We do not care about breaking previously serialized structs between versions, possibly silently breaking between versions. 

we could either not care, or we could also implement a custom deserializer that would be backwards compatible, but i think its pretty unlikely that the ItemStack / ItemKind format will change any time soon.

crates/valence_generated/build/item.rs
dyc3: Yeah the string ident. Minecraft doesn't have public facing numerical IDs anymore

crates/valence_protocol/src/item.rs
dyc3: Ok, then let's forego the serde feature for now.

crates/valence_generated/build/item.rs
lukashermansson: From what I can tell from the latest efforts to update the string ids are not stable across versions either, updates to Minecraft versions could break de-serilization.

These changes should not be silent (unless a rename (from A to B) and a new item gets introduced with the name A. Where the produced result would yeild another item than what was intended when serializing.

maxomatic458: i assume that if minecraft were to change their IDs, then the enum variants generated by the build script would also change.
If thats the case id say thats out of scope for valence.

dyc3: Yes. Valence always tries to target the latest stable version of Minecraft.

lukashermansson: Yep the enum members would change between versions, the main thing i'm worried about it that it may look like its a well supported mechanism while it in reality would be error prone. I personally would be happy to put this kind of serialization and de-serialization in its own little package that can document the caveats (or work around the caveats eventually?) (maybe as a fisrt implementation it can reject deserilaizing if some kind of version tag does not match)? 

I think people might use the serialization features without properly considering the caveats.

dyc3: Can you find an example of Minecraft item names changing between versions? I'm not convinced that this is something we can't work around.

Minecraft changing their items names frequently would also mean that they break user's data packs frequently. For any name changes that do occur, we could add exceptions to the deserialize impl to auto convert it to the new name.

lukashermansson: I could only find such changes listed here in the extracted json:
https://github.com/valence-rs/valence/pull/594/
here is one example:
Skärmbild från 2024-10-13 10-41-17

Yep we could put a translation layer somewhere (if we think its in scope for valance to support deserializing old serailized files) for that translation layer to work i think we would require tagging the output with some kind of version identifier (to be able to know that grass -> short_grass when moving from 1.20.1 -> 1.20.4, without it we would not be able to correctly translate names if they ever become reused )

I have no idea if minecraft does these things internally for the datapack interface, i guess they might? 

I think all of this should probably be in scope for valance but i would really just love it if its a separate module, that can wrap these structs, and also tag with a serialized version alongside the serialized data. (and maybe initially it does not have to have built in translation mechanism).

maxomatic458: if serialization fails it would throw a pretty clear error, so one google search would fix it.
Not sure if it would be that useful to add full backwards compatibility for valence for cases like that.

But it could certainly by done by just checking for the outdated ids and then for the current ones.

lukashermansson:
> @lukashermansson
> 
> > My personal stance on this is that we should keep serialization out of the scope for the framework(core), as the added overhead of serialized items will result in one of two things (or both):
> 
> we could put that behind a feature flag
> 
> > We do not care about breaking previously serialized structs between versions, possibly silently breaking between versions.
> 
> we could either not care, or we could also implement a custom deserializer that would be backwards compatible, but i think its pretty unlikely that the ItemStack / ItemKind format will change any time soon.
@maxomatic458 

You are right in that we could add a custom deserializer that is backwards compatible, but does such a concern really belong in `valance_protocol` or `valance_genereated`? Imo it does not.

I also don't see why `ItemStack` might not change in the near future, one limitation currently its hardcoded to only be able to understand vanilla item. As far as I am aware the vanilla client registers its items in a "registry" this allows for mods to register their items in said registry. I could see a lot of benefits that valance replicates this approach of an item_registry (this might mean the enum for ItemKind might just be an abstraction for easily naming the vanilla items in an `item_registry` and `ItemStack` might not refer to an enum of vanilla items at all) (similar to the [biome registry](https://github.com/valence-rs/valence/blob/main/crates/valence_registry/src/biome.rs) although currently the item_registry system that Minecraft has does not need to be synced by the server while the biome registry does, (atleast as of the current Minecraft version))

I could see a future where `ItemStack` looks something like this in the code
```rust
#[derive(Clone, PartialEq, Debug, Default)]
pub struct ItemStack {
    pub item: ItemRegistyItemRef, // <- where this is just an identifier to an item registry that has all the vanilla items and all of the user defined items that might be present in mods.
    pub count: i8,
    pub nbt: Option<Compound>,
}
```
^ im not saying we can't support a backwards compatible de-serialize here either by referencing the registry, but it doubt it would be trivial, and definitely not something that should live in this module imo. 

While everybody might not consider supporting communicating inventory updates and similar stuff containing itemstacks with items that are added via some mod to be in scope of valance, I think a change such as this could have the ability to limit what kind of changes are easy to do (while also being backwards compatible with a serialization feature, and I think most people expect a serialized format to be stable). For a change as the above if serialization to still work with the old format would have to take a dependency on wherever the registry lives just for the sake of being able to support deserializing, while it might not if serialization is not a consern of the `valance_protocol` crate.

I also think all my concerns goes away if we put all of this serialization stuff in its own module, that can deal with the eventual complexities that might arrive if the core framework evolves to for example support the above (and possibly other cases too)

crates/valence_generated/build/item.rs
lukashermansson: I think one core limitation of this is that we might not be able to unambiguously translate names unless we also know what version they where serialized on unless there is some commitment to never reuse old identifiers that mojang has.

Unless we also are able to capture when an item was serialized we cant provide this feature to users. I know its an edge case, not sure it will ever really happen, but we are going to be limited to not be able to support it unless we also consider tagging the serialized output so we can properly account for these changes. 

Supporting backwards compatibility in any way is going to impact what improvements we can do, I don't think any of those concerns should be apart of the core library modules, having those conserns in thier own crates, that live in the repo would imo be the best approach (as long as they are automatically maintained for the case of a mapping layer) or just refuse to deseialize an old version because we know what kind of work it would take to gracefully implement such compatibility. 

The ids that are renamed and thus no longer exist are "fine" imo. They break visibly to the user at point of deserilaizing, the oddities come in when people can deserilaize old files that potentially could contain ids that have over a long period of time been reused, and we would have no way to prevent such a footgun from happening (either by mapping it to the new version, or by refusing to deserialize it). If someone for example built an inventory saving system built on top of these trait implementations it could arbitrarily silently change user "data". 

I don't personaly have an opinion if valance should be concerned with mapping or not, it certainly could. I just don't think valance should expose an api to make it seem like we have some kind of good support for this (as would be expected by our users if they see an implementation for serialize and deserialize). Instead it should either be clear that we support serializing and deserilaizing for a given version identifier, or that with a version identifier we can properly map items to work across version boundaries.

I can also imagine the dangers to be documented and explicitly opted into the dangerous version, something like:
```rust

/// To serialize and deseraize item stacks, its important to know at what version we serialized the struct
struct SerializableItemStack {
   item: ItemStack,
   version: u32
}

impl SerializableItemStack {
   fn get_item() -> Result<ItemStack, SomeError> {
     // check version and return item, or in-case we are ambitious perform mapping of item kind and return items anyway even if versions dont match (if it can be done automatically) 
     // the api here should document what kind of guarantees we make.
   }
   /// Caution does not check serialized version, can transform serialized the intent of the serilized item across versions
   /// beweare as properties such as item type can be different at point of serilaization and deserialization 
   fn get_item_unsafe() -> ItemStack {
      // just return the item
   }
}
```

maxomatic458: I think this might be the best approach if we want to ensure backwards compatibility.

Im im still not 100% conviced valence should even ensure backwards compatibility.

lukashermansson: My personal stance on it is that I also doubt we need backwards compatibility, so i think we should make that explicit (and guarantee we are not sometimes backwards compatible, and other times we may silently "transmute" items) (and properly document that we only can serialize and de serialize within a version) 

I'm sorry if I'm a pain here, I just think its worthwhile to consider an api that will not break user expectations, especially since all of the fields here are pub, and if you need a crude serialize and de serialize (if you know and accept the risks of serializing without tagging a version) for your use case you can implement those traits on your own newtype wrapper.

I ultimately like the idea of serializing and i think it should also be apart of the library, I think we should implement it in a way that clarifies what guarantees we make, and also in its own module (nice for compilation speeds but also more crucially that we can support a move to data driven items when/if minecraft implements those (such a move would mean the protocol module cant reference this hardcoded set of items as it currently does))

maxomatic458: yes i think then we probably want to create a `valence_serde` or `valence_serialize` crate where we can create wrapper structs for commonly serialized structs, like Items, Inventory-content, etc.

Ill try to implement that if I have time for that.

maxomatic458:
ill close that and open up a new PR where the serialization will be implemented in its own crate (using wrapper structs that can be serialized as @lukashermansson suggested)

crates/valence_generated/build/item.rs
maxomatic458: thoughts on something like this? would live in its own `valence_serde` crate and be behind a `serde` feature flag

```rs

/// A Wrapper around [`ItemKind`] that provides serialization and deserialization support.
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
pub struct SerItemKind(pub ItemKind);

impl Serialize for SerItemKind {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        self.0.to_str().serialize(serializer)
    }
}

impl<'de> Deserialize<'de> for SerItemKind {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let s: &str = serde::Deserialize::deserialize(deserializer)?;
        let item_kind = ItemKind::from_str(s).ok_or_else(|| {
            serde::de::Error::invalid_value(serde::de::Unexpected::Str(s), &"the snake case item name, like \"white_wool\"")
        });

        item_kind.map(SerItemKind)
    }
}

impl From<ItemKind> for SerItemKind {
    fn from(kind: ItemKind) -> Self {
        SerItemKind(kind)
    }
}

impl From<SerItemKind> for ItemKind {
    fn from(kind: SerItemKind) -> Self {
        kind.0
    }
}
```

lukashermansson: I  think a wrapper like this should store a private field about what version it was serialized on so we can prevent silent misuse of these across versions. (By giving an error when de-serilization happens in case the version is not the same as the constant defined as "current")

Other than that I think its along the lines of what i expect. 

It also needs some documentation about what compability we provide (or in this case lack of compability), doc comments on the wrapper is probably a good place to place such a note.

Also the use cases I primarily imagined dealt mostly with item stacks, but i assume that might also be in scope of what you want to provide in the upcoming pr, there you might also want to support lists of items with a single version tag for space efficiency.

lukashermansson: it could also defeer the "failing" to the call to into() (on the SerItemKind) (making it a try_into() and failing when version does not match) 

optionaly it could expose a way to get at the itemkind, as long as its really clear that its dangerous, maybe something like `unsafe fn get_itemkind_unchecked() -> ItemKind`, should ofc have clear docs on why its dangerous and what should be considered 
But it really depends on if we think people are going to use that, it could be left out if we think it would not bring any value

maxomatic458: Its pretty much a tradeoff between ease of use and safety.

Either we can have those safety checks, but it wouldnt be as easy to use (because of the wrapper types)
or we dont have safety checks and directly implement serialization on the raw types (behind a feature flag).

or we could also implement both maybe, as a `serde` and `serde_unsafe` feature?

where one would perform those version checks and the other one wouldnt.

I do really think being able to serialize and deserialize the raw structs would be more convenient.

PR #254, opened 2023-02-20
Add entity equipment support

mymatsubara:
## Description

Main additions:
- Add a  Bevy `Component` called `Equipments` which tracks entity's current equipments and if they have been modified.
- Add a Bevy system called `update_equipment` which sends [`SetEquipment`](https://wiki.vg/Protocol#Set_Equipment) packets to all clients in view, when an entity with a `Equipments` component is modified.
- Modify [McEntity::write_init_packets] to also send `SetEquipment` packets.
- Add `random_equipment.rs`

These change only have cosmetic effect for now, but the `Equipments` component can be also used when dealing with game logic when equipments data is required.

## Test Plan

Steps:
1. `cargo test -p valence --tests` (there is a unit test in `equipment.rs`)
2. Run the `random_equipment` example with: `cargo run --example random_equipment`
3. Join the server on `localhost`
4. An armor stand is spawned near the spawn point
5. Every `McEntity` will be assigned a random set of equipments every second

### Video of the `random_equipment` example
https://user-images.githubusercontent.com/49251043/220204501-9cfa5e61-d486-4ecc-93d8-34e965c63dfd.mp4

**PROBLEM**: As we can see from the video, our client only display equipment changes from other players, but not from our own player entity. This happens because the equipments of our player entity are tracked client-side, thus the `SetEquipment` packet does have effect when the `entity_id` is equal to the your player entity id. There might me another packet type which can be used for such case, but I'm not sure don't know which one 😞 

#### Related

Issue related: #223

PR #663, opened 2024-10-13
Add Equipment support

### LINKED ISSUE #662: Entity Equipment

Valence does not sync player equipment (held items, armor, etc)

a Equipment plugin which exposes a Equipment component.

this could also be implemented in the inventory crate, although in my opinion it would be better to implement equipment as a seperate Plugin.

## Additional context
related #254

I believe an Equipment Component could be implemented like this

```rs
#[derive(Debug, Default, Clone, Component)]
pub struct Equipment {
    main_hand: ItemStack,
    off_hand: ItemStack,
    boots: ItemStack,
    leggings: ItemStack,
    chestplate: ItemStack,
    helmet: ItemStack,

    /// Contains a set bit for each modified slot in `slots`.
    #[doc(hidden)]
    pub changed: u8,
}

impl Equipment {
    pub const SLOT_COUNT: u8 = 5;
    
    pub const MAIN_HAND_SLOT: u8 = 0;
    pub const OFF_HAND_SLOT: u8 = 1;
    pub const BOOTS_SLOT: u8 = 2;
    pub const LEGGINGS_SLOT: u8 = 3;
    pub const CHESTPLATE_SLOT: u8 = 4;
    pub const HELMET_SLOT: u8 = 5;
    
    pub fn main_hand(&self) -> &ItemStack {
        self.main_hand;
    }
    
    pub fn set_main_hand(&self) {
         ...
     }
     ...
}
```

that would make it pretty easy to use.

I also think this should be as seperate as possible from ``valence_inventory`` (so people using this crate would have to implement custom logic to update equipment based on inventory changes, but that should not be to hard)

we would have to use Inventory to get the HeldItem

thats the packet https://wiki.vg/index.php?title=Protocol&oldid=18375#Set_Equipment

Edit:
since we would have to use setters/getters (so we garuantee that updated the changed field)
we might as well store the actual items in a list instead of the fields.

Edit 2:
My idea was that valence would only provide this component, and an updater system.
So you would have to implement custom logic if you want to lets say set your equipment when you put on armor (as i do think there are use cases where you dont want to bind it to the inventory)

Any thoughts on that?

maxomatic458:
# Objective

- make valence work with entity equipment (armor, main/off hand items)
# Solution

adds the crate `valence_equipment` that exposes the Equipment Plugin.
every `LivingEntity` will have a `Equipment` Component.

The Equipment plugin will NOT be compatible with the inventory by default (thats intended, as that makes it possible to use the equipment feature without the inventory feature + I do think there are probably use cases where you want to make players appear as having armor, although they dont have it in their inventory)

This PR would add Events when entities are (un)loaded by a player (used for sending equipment once the player loads an entity). I do believe this might also be a useful feature outside of the equipment feature.

used #254 as reference + stole example idea

fixes #662 

Opening as a draft for now (might refactor the updating/event emitting system).
Let me know what you think of the Entity Load/Unload events.


maxomatic458:
if the `equipment` feature is enabled (which it is by default), every `LivingEntity` will have the `Equipment` Component.

maxomatic458:
Should be ready now (not sure why the CI is failing, the EquipmentPlugin is used in the DefaultPlugins)

As said above, this touches the client code and implements Custom events there, kind of a hack, but that seems to be the easiest way to do it without having to reimplement the "view" logic.

dyc3 (commented):
I haven't reviewed this in depth yet, but I like how this is decoupled from inventories. Good tests too.

I think it would be nice to provide a utility system to keep equipment in sync with the entity's inventory. I foresee that being common enough that we should just provide that.

maxomatic458:
> I think it would be nice to provide a utility system to keep equipment in sync with the entity's inventory. I foresee that being common enough that we should just provide that.

I agree with that for player inventories (as players can change the inventory themselves).

But im not sure that would be good for all `LivingEntities`, that would open up 2 ways of dealing with equipment (either through the `Equipment`, or through the `Inventory` component.
And i believe it would be better to just allow equipment handling through the `Equipment` component.

I can't think of any case where you would have to handle anything equipment related through the inventory

maxomatic458:
Ive added the `EquipmentInventorySync` component that will have to be attached to entities, once attached the inventory & equipment components will be synced (equipment changes will be prioritized if both change in the same tick).

crates/valence_equipment/src/lib.rs
dyc3, line 1: This should include readme.md as the module level docs. I'm pretty sure that's how we have it set up for other packages.

crates/valence_equipment/README.md
dyc3, line 1: Feel free to elaborate on these docs with code examples, explanations of equipment slots, links to other docs, etc. Consider what would be useful as an end user.

crates/valence_equipment/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+    mut commands: Commands,
+    mut entities: Query<Entity, (Added<LivingEntity>, Without<Equipment>)>,
+) {
+    for entity in &mut entities {
+        commands.entity(entity).insert(Equipment::default());
+    }
+}
+
+#[cfg(feature = "inventory")]
+mod inventory_sync {
```
dyc3, line 252: Move this module to a new file

crates/valence_equipment/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+fn on_entity_init(
+    mut commands: Commands,
+    mut entities: Query<Entity, (Added<LivingEntity>, Without<Equipment>)>,
+) {
+    for entity in &mut entities {
+        commands.entity(entity).insert(Equipment::default());
+    }
+}
+
+#[cfg(feature = "inventory")]
```
dyc3, line 251: I don't think it's necessary to hide this behind a feature.

crates/valence_equipment/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+                inventory_sync::on_attach_inventory_sync,
+            ),
+        )
+        .add_systems(
+            Update,
+            (
+                update_equipment.before(FlushPacketsSet),
+                on_entity_load.before(FlushPacketsSet),
+                #[cfg(feature = "inventory")]
+                inventory_sync::equipment_inventory_sync,
```
dyc3, line 31: We don't need to put this behind a feature, the system will not run if nothing matches the query.

crates/valence_equipment/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+        app.add_systems(
+            PreUpdate,
+            (
+                on_entity_init,
+                #[cfg(feature = "inventory")]
+                inventory_sync::on_attach_inventory_sync,
+            ),
+        )
+        .add_systems(
+            Update,
```
dyc3, line 26: IIRC, we've reserved the Update stage to be *just* for end user systems. This should probably be PostUpdate.

dyc3 (changes requested):
Looks good, just minor things and docs.

crates/valence_equipment/src/lib.rs
maxomatic458: i had to move inventory_sync::equipment_inventory_sync to ``PreUpdate``, because the tests were being flaky.

dyc3:
Go ahead and hit resolve on the comments you've addressed in new commits so I know you got them.

maxomatic458:
should be ready now!

maxomatic458:
i just noticed that the `HeldItem` is not marked as changed. (which means the InventorySync will not work for Helditems)
https://github.com/valence-rs/valence/blob/f526a175bb7fee13d59c6987d2bf4e9b532866e5/crates/valence_inventory/src/lib.rs#L1411


is there any reason for not updating it?
@AviiNL (i saw you made that change)

maxomatic458:
yea looks like removing that causes some funky behavior.

Ill try to change that so we dont rely on the HeldItem, but listen for the `UpdateSelectedSlotEvent` instead

AviiNL:
> i just noticed that the `HeldItem` is not marked as changed. (which means the InventorySync will not work for Helditems)
> 
> https://github.com/valence-rs/valence/blob/f526a175bb7fee13d59c6987d2bf4e9b532866e5/crates/valence_inventory/src/lib.rs#L1411
> 
> is there any reason for not updating it? @AviiNL (i saw you made that change)

I don't even remember what I had for breakfast this morning, let alone something I've done over a year ago lol, there probably was a reason for it since it was discussed and the implementation changed in PR #525 there might be more info in discord but I really can't remember.
This is also from pre 1.21, so it's very possible that 1.21 changes made this requirement (whatever it was) obsolete.

lukashermansson:
> This is also from pre 1.21, so it's very possible that 1.21 changes made this requirement (whatever it was) obsolete.

We are still on 1.20, the work to get us to 1.21 is still ongoing, so as far as I'm aware any requirement that was present then should still be a requirement as of right now (in `main` at least)

maxomatic458:
> yea looks like removing that causes some funky behavior.
> Ill try to change that so we dont rely on the HeldItem, but listen for the UpdateSelectedSlotEvent instead

I added a seperate system specifically for handling client HeldItem changes now

crates/valence_equipment/src/inventory_sync.rs
```diff
@@ -0,0 +1,89 @@
+use valence_inventory::player_inventory::PlayerInventory;
+use valence_inventory::{HeldItem, Inventory, UpdateSelectedSlotEvent};
+use valence_server::entity::player::PlayerEntity;
+
+use super::*;
+#[derive(Debug, Default, Clone, Component)]
+pub struct EquipmentInventorySync;
```
dyc3, line 7: This is user facing and needs a doc comment.

dyc3 (approved):
Looks good, just a minor doc thing.

Issue #396, opened 2023-06-22
Flaky inventory test


rj00a:
## Valence Version

f1e8b59587a9496d4c2c96c5a65599c982239ace

## What You Did

Run tests locally

## What Went Wrong

The `dragging_items` test occasionally fails.

## Additional Information

This bug probably existed prior to bevy 0.11, but is only now manifesting due to changes to bevy_ecs scheduling.

The problem is likely caused by command application not being properly taken into account, since that was the issue in 6f88017ee17a6db4ea41e5ec22172b4e100c62ed. `apply_deferred` (previously known as `apply_system_buffers`) is allowed to be scheduled between systems by bevy, which may break assumptions.

PR #671, opened 2024-10-25
resync readonly inventory after block interaction

maxomatic458:
fixes a desync where a player with readonly inventory perfoms a block interaction (such as placing a block).


maxomatic458:
^ flaky test from #396

PR #680, opened 2025-03-07
Broadcast equipment interaction

Cheezer1656:
# Objective

To broadcast players' item interactions to other players as part of `EquipmentPlugin`

# Solution

When a player has the `EquipmentInteractionBroadcast` component:
- Set the "use_item" LivingFlag to true whenever an `InteractItemEvent` is received
- Set the flag false when a `PlayerActionC2s` packet with a `PlayerAction::ReleaseUseItem` action is received

### DIFF
--- a/crates/valence_equipment/src/interaction_broadcast.rs
+++ b/crates/valence_equipment/src/interaction_broadcast.rs
@@ -1,8 +1,11 @@
-use valence_inventory::PlayerAction;
+use std::i8;
+
+use valence_inventory::{HeldItem, Inventory, PlayerAction};
 use valence_server::entity::living::LivingFlags;
 use valence_server::event_loop::PacketEvent;
 use valence_server::interact_item::InteractItemEvent;
 use valence_server::protocol::packets::play::PlayerActionC2s;
+use valence_server::ItemKind;
 
 use super::*;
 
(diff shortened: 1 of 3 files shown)


Cheezer1656:
I'm not sure why the test is failing. All of the tests pass on my machine for some reason

dyc3:
FYI, CI is using rust 1.85.0. I'd prefer if the clippy stuff is resolved in a different PR since it looks unrelated to the changes here.

dyc3 (approved):
This looks good aside from the CI failures.

Cheezer1656:
Oops I forgot about the Rust version and I was using `1.87.0-nightly` before. I ran it with `1.85.0` and it still passed though (I'm on Arch not Ubuntu so there might be a difference there), so I'm not sure what's going on. Meanwhile I'll see if I can fix the clippy problems and open another PR for that.

crates/valence_equipment/src/interaction_broadcast.rs
```diff
@@ -0,0 +1,42 @@
...
+pub(crate) fn start_interaction(
+    mut clients: Query<&mut LivingFlags, (With<Client>, With<EquipmentInteractionBroadcast>)>,
+    mut events: EventReader<InteractItemEvent>,
+) {
+    for event in events.read() {
+        if let Ok(mut flags) = clients.get_mut(event.client) {
+            flags.set_using_item(true);
+        }
+    }
+}
```
maxomatic458, line 25: Im pretty sure that this will play the bow draw animation even if it does not play for the client (in the case the client has no arrows in the inventory). Probably also the same for the crossbow.

at least i implemented a check like this:
https://github.com/maxomatic458/bedwa-rs/blob/28808f46845a2fd8c656d9beaeef8f20c0a033d2/src/base/bow.rs#L116-L123

Cheezer1656: You're right I should add checks.

By the way, do you know how to make clients not be able to draw their bows when they don't have arrows? In valence it seems like you're always able to draw your bow, even if you're in survival without arrows.

Cheezer1656: Oh nevermind that was a side effect of not having the checks (the flag played for the client too)

maxomatic458: maybe another edge case would be items with a cooldown. like ender pearls.
although im not sure if the mc client would even send the InteractItemEvent in the first place

and since valence doesnt have any API right now to interact with item cooldowns (as far as i know)
this might be a thing for another PR.

Cheezer1656: Yeah probably best for another PR. If valence kept track of item usage/cooldowns we might also be able to keep crossbows drawn (rn if you switch to another hotbar slot and then back to the drawn crossbow, it won't appear drawn to other players)

dyc3:
Make sure to rebase this PR so it has the clippy fixes

dyc3:
There's some clippy failures in this PRs code

Cheezer1656:
Yup I saw. I'll fix it right now

PR #681, opened 2025-03-08
Fix clippy issues

Cheezer1656:
# Objective

Fix the clippy errors in the CI

# Solution

- Run `cargo +stable clippy --workspace --no-deps --all-features --all-targets --fix -- -D warnings` multiple times
- Add `#[allow(clippy::large_stack_arrays)]` https://github.com/valence-rs/valence/blob/8f3f84d557dacddd7faddb2ad724185ecee2e482/crates/valence_generated/build/block.rs#L695
- Replace every triple slash comment in `valence_generated` and `valence_entity` with `#[doc]`
- Replace eliding `'a` lifetimes in `stresser.rs` with a single `'_`

### DIFF
--- a/crates/java_string/src/iter.rs
+++ b/crates/java_string/src/iter.rs
@@ -140,7 +140,7 @@ pub struct EscapeDebug<'a> {
 }
 delegate!(Iterator for EscapeDebug<'a> => char);
 delegate!(FusedIterator for EscapeDebug<'a>);
-impl<'a> Display for EscapeDebug<'a> {
+impl Display for EscapeDebug<'_> {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         self.clone().try_for_each(|c| f.write_char(c))
     }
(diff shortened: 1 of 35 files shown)


maxomatic458:
@Cheezer1656 
add `#[allow(clippy::large_stack_arrays)]`
above the const

https://github.com/valence-rs/valence/blob/8f3f84d557dacddd7faddb2ad724185ecee2e482/crates/valence_generated/build/block.rs#L695

Cheezer1656:
I think it's just the needless string literals now. Should I convert every triple slash comment into `#[doc]` to avoid Rust generating the raw strings?

crates/java_string/src/iter.rs
```diff
@@ -206,7 +206,7 @@ impl<'a> Iterator for Chars<'a> {
         // `(len + 3)` can't overflow, because we know that the `slice::Iter`
```
JackCrumpLeys, line 206: This comment needs to be removed as it dosen't make sense in this context

crates/valence_anvil/src/lib.rs
```diff
@@ -570,7 +570,7 @@ impl Region {
 
         // additional 5 bytes for exact chunk size + compression type, then add
         // SECTOR_SIZE - 1 for rounding up
```
JackCrumpLeys, line 572: part of this comment doesn't make sense

JackCrumpLeys (commented):
2 minor issues with inconsistent comments. otherwise LGTM. @dyc3

JackCrumpLeys (approved):
LGTM :+1:

dyc3 (approved):
It would probably be a good idea to pin our rust version so clippy doesn't unexpectedly block PRs again. Not gonna block this PR for that though. Lgtm

PR #682, opened 2025-03-10
TODO Check in CI

Cheezer1656:
# Objective

Get people to make issues for every TODO so that they are well defined, accessible, and in a central location.

# Solution

Add a CI check for TODOs that do not reference a Github issue.


scripts/check_todos.py
```diff
@@ -0,0 +1,31 @@
...
+        result = re.search(r"((//|/\*|///).*TODO|todo!)(?!.*\(#\d+\))", line)
+        if result:
+            results.append((i, line))
+    return results
+
+if __name__ == "__main__":
+    paths = list(pathlib.Path(".").rglob("*"))
+
+    clean = True
+    for path in paths:
```
JackCrumpLeys, line 17: I know you are just starting but this will include code in target so be careful. We cannot copntroll TODO usage in code outside valence. I would stick to `examples/*`, `src/*`, `crates/*`, `extractor/src/*` and any stand alone files in `.` (like cargo.toml).

JackCrumpLeys (commented):
just had some concerns about what code we are scanning

scripts/check_todos.py
Cheezer1656: Good point. Do you think it would be better to have a whitelist or blacklist for scanning?

JackCrumpLeys: probably both (whitelist text files and skip checking things like json, while blacklisting certain dirs) idk tho if its worth going outside of just .rs files to cover some on the config files. for detecting comments most languages either have // or # so that should cover alot. the java is a little more complex as with the /* syntax

Cheezer1656: In commit 5c8a13df7858acc9a39bc489ae2bce36403c4052 I made it so that the script only searches these directories: `.github assets benches crates examples extractor/src scripts src tools website`. Is this good enough?

PR #688, opened 2025-05-18
Pin Rust version in CI

Cheezer1656:
# Objective

Specify the Rust toolchain version in the CI workflows so that unrelated issues don't pop up in PRs.

# Solution

Replace all instances of `toolchain: "nightly"` with `toolchain: "1.89.0"`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Setup Rust toolchain and cache
         uses: actions-rust-lang/setup-rust-toolchain@v1.9.0
         with:
-          toolchain: "nightly"
+          toolchain: "1.89.0"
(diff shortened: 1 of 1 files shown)


dyc3 (commented):
Hmm, these must have been on `nightly` for a reason.

Regardless, the latest stable rust is `1.87`, not `1.89`.

Cheezer1656:
Turns out `cargo-udeps` requires `nightly` to run. Is it possible to pin a nightly version?


> Regardless, the latest stable rust is `1.87`, not `1.89`.

I put `1.89.0` because I thought you could put the latest nightly version. That's my bad.

dyc3:
I think keeping udeps on nightly is fine. It's mostly clippy that's problematic

dyc3 (approved):

PR #713, opened 2026-06-13
Fix tests error log by enabling LogPlugin only once

### LINKED ISSUE #712: Global logger and tracing subscribers are set multiple times in tests
Hello. I'm new to valence!

HEAD: 2b705351b4b99557b1a53360f73cd485c05eeb64

I just ran `cargo test` and saw a lot of same errors coming up:

```
ERROR bevy_log: Could not set global logger and tracing subscriber as they are already set. Consider disabling LogPlugin.
```

This is because `LogPlugin` is initialized multiple times when running multiple tests together.
No error would pop up if only running a single test at a time.

How should we address this? Should we just disable `LogPlugin` in `valence::testing::ScenarioSingleClient::new`?

Bowen951209:
# Objective
Fixes #712.
When running tests, there are many:

```
ERROR bevy_log: Could not set global logger and tracing subscriber as they are already set. Consider disabling LogPlugin.
```

Because tests usually run in the same context, but
each test tries to set logger and tracing subscriber,
which should be only set once in a context.

# Solution

This commit takes the approach of only setting up
[LogPlugin] once in the whole test run, rather than
removing [LogPlugin] entirely for all test apps.

Although there are no logs in Valence's existing tests
currently, keeping the logger available is useful for
debugging single tests in the future.

Also note by calling `testing::add_plugins`,
`example_test_server_tick_increment()` and
`idle_update` now disables [NetworkPlugin],
which should be okay in these tests.

### DIFF
--- a/benches/idle.rs
+++ b/benches/idle.rs
@@ -1,5 +1,5 @@
 use divan::Bencher;
-use valence::prelude::*;
+use valence::{prelude::*, testing};
 
 /// Benches the performance of a single server tick while nothing much is
 /// happening.
@@ -7,7 +7,7 @@ use valence::prelude::*;
 pub(crate) fn idle_update(bencher: Bencher) {
     let mut app = App::new();
 
-    app.add_plugins(DefaultPlugins);
+    testing::add_plugins(&mut app);
     app.add_systems(Startup, setup);
 
     // Run startup schedule.
--- a/benches/many_players.rs
+++ b/benches/many_players.rs
@@ -8,11 +8,10 @@ use valence::keepalive::KeepaliveSettings;
 use valence::layer::chunk::UnloadedChunk;
 use valence::layer::LayerBundle;
 use valence::math::DVec3;
-use valence::network::NetworkPlugin;
 use valence::protocol::packets::play::{FullC2s, HandSwingC2s};
 use valence::registry::{BiomeRegistry, DimensionTypeRegistry};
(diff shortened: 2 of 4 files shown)


dyc3:
CI is broken

Bowen951209:
> CI is broken

Yeah. But it doesn't look like it's due to this change.
Should I fix it?

dyc3:
Unfortunately I dont have permissions to merge anything if CI isn't passing. So either this PR or a new PR to fix it is fine.

Bowen951209:
> Unfortunately I dont have permissions to merge anything if CI isn't passing. So either this PR or a new PR to fix it is fine.

I have opened a CI fix PR in #714.

dyc3:
Merged, go ahead and rebase :)

Bowen951209:
I'll deal with the test failure tomorrow.

dyc3:
Could just be a flaky test

Bowen951209:
Hmm... Testing directly on my environment passes, but running on CI environment with `act` locally fails on `tests::inventory::dragging_items`, too.

Also, I noticed that benches have the same issue on `Could not set global logger and tracing subscriber`.

Bowen951209:
> Could just be a flaky test

Job succeed on  another local `act` run. 
It should be indeed a flaky test.

dyc3 (approved):

